PR #13, opened 2015-08-12
Added sorting notes list

pmarki:
I sorted out sorting notes list alphabetical or by date.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5,6 +5,7 @@
 #include <QSplitter>
 #include <QVBoxLayout>
 #include <QListView>
+#include <QSystemTrayIcon>
 
 
 int main(int argc, char *argv[])
@@ -13,6 +14,6 @@ int main(int argc, char *argv[])
 
     MainWindow w;
     w.show();
-
+   // a.setQuitOnLastWindowClosed(false);
     return a.exec();
 }
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -77,6 +77,7 @@ MainWindow::MainWindow(QWidget *parent) :
 
     this->updateService = new UpdateService(this);
     this->updateService->checkForUpdates();
+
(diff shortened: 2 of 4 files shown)


pbek:
Thanks a lot for your contribution! Although there is an issue with it.
When you edit a note or a note gets changed externally the list will be sorted by date again (without showing it in your new menu entry).

pbek:
And it is possible to un-check the sort order in menu. :)

pmarki:
I think I have fixed the issues. I can implement system tray if you 
wish. That would be useful for me too :)

pbek:
Thank you for your changes! A lot better now! :)

I still have found some issues:
- When in "alphabetical" mode and other (note) file gets added externally and this file changes the first note will be opened regardless what note was edited currently.
- Could be be so kind and keep the spaces in method calls, like for example: 
```C++
QCoreApplication::setOrganizationName( "PBE" );
```
versus
```C++
QActionGroup *sorting = new QActionGroup(this);
```
- Please don't work in the master branch or I will have to add the pull request as patch to the development branch.
- You could store the sorting as QSettings, so the user gets the same sorting after he/she restarts the app.

regarding the Systray: Sure, sounds nice, if it will work with Linux/Mac/Windows and can be turned off for those who want to have a normal "running application".

Thanks again for your great work!

pbek:
@pmarki I merged your commits now by hand, tested on all platforms and fixed some problems. Thank you again for your great work! Please be sure to work in the develop branch and file pull requests from there in the future! ;)

The master branch is just for releases and triggers the creation of release builds and deployment of binaries for all platforms!

pmarki:
I'm glad I could help and sorry for causing problems with branches. I couldn't test in on other platforms than linux.

pbek:
Maintaining software on more than one platform is hard work. ;) You did great!

PR #51, opened 2016-02-03
Make all string strings translatable and add french translation

Dalan94:
Here is my Internationalization work. The only thing (I think) which is not translated is the about text because you use an HTML file and I don't know how to translate it.


pbek:
We seem to have some merge conflicts here. Is your fork uptodate?

pbek:
And you also modified the German translation, was that intentional? :grin: 
I can't wait to test your changes.

Dalan94:
I don't know why there is a merge conflict but I will try to update my repo.
The changes in the German translation are the new strings created when I use `lupdate` but I can remove it if you want.

pbek:
Yes, please discard those German modifications. And the English translation files were erase too meanwhile.

Dalan94:
Ok I update the repo and use your version of the german translation.
But there is again merge conflict, I don't know why and where. Maybe you can fix them during the merge.

src/dialogs/linkdialog.ui
```diff
@@ -11,7 +11,7 @@
    </rect>
   </property>
   <property name="windowTitle">
-   <string>Search note</string>
```
pbek: Why did you have to change the window title?

pbek:
The problem is GitHub will not let me merge the pull request. I would need to get your changed files down somehow and see what needs to be thrown out and test and reformat what could be let in. Is there a way you can resolve the merge conflict and make an other pull request?

src/dialogs/linkdialog.ui
Dalan94: Because I thought that was a bad copy/paste or something like that because "Search note" doesn't seam to be good for the link dialog (where we can also link to an URL)

pbek: Oh, I never saw that one because I overwrite the title in source code. Thank you!

Dalan94:
Ok I fixed the conflict

pbek:
From what I can see you did a great job! I will test it tomorrow and will report back then. :+1: :bow:
Currently there is no French translation added, I'm I correct?

Dalan94:
Thank you.
No there is also the french translation.

pbek:
Sorry, missed the files! Truly an awesome job! Thank you heaps! :+1: :tada:

PR #53, opened 2016-02-04
Released first version of the Polish translation. 

Olhur:
It probably still needs some polishing (see what I did there?), but it is complete.

Known issues:
- was not sure if shortcuts (made with ampersand) should be forced onto some menu options, since translation does not allow easy hooks like this, so they were skipped where not easily introduced.
- there are few things that could've been written better, both in English and Polish, so there will be another proposal some other day to change both.


pbek:
Awesome job! :+1: Thank you a lot!
I now added the Polish language to the UI and the packaging processes!
It will be available with v0.90.

PR #54, opened 2016-02-04
Start in set mode

dasmaetthes:
Added settings to set the view mode for start(edit mode or markdown
mode)


src/mainwindow.cpp
```diff
@@ -146,7 +146,7 @@ MainWindow::MainWindow(QWidget *parent) :
     ui->encryptedNoteTextEdit->setTabStopWidth(width);
 
     // set the edit mode for the note text edit
-    this->setNoteTextEditMode(true);
+    //this->setNoteTextEditMode(true);
```
pbek: why did you have to remove this?

src/dialogs/settingsdialog.cpp
```diff
@@ -164,10 +166,10 @@ void SettingsDialog::readSettings() {
             settings.value("notifyAllExternalModifications").toBool());
     ui->noteSaveIntervalTime->setValue(
             settings.value("noteSaveIntervalTime").toInt());
-    ui->toolbarIconSizeSpinBox->setValue(
```
pbek: why did you have to remove these 2 lines?

src/mainwindow.h
```diff
@@ -172,6 +172,7 @@ private slots:
     SimpleCrypt crypto;
     QDateTime currentNoteLastEdited;
     bool notifyAllExternalModifications;
+    bool setViewMode;
```
pbek: I think `set` isn't a good prefix for a variable. Just `viewMode` might be better...

src/dialogs/settingsdialog.cpp
dasmaetthes: Sorry I just used it for test case to understand better how it works and forgot to reverse it

src/mainwindow.cpp
pbek: Does editing of encrypted notes still work with that?

dasmaetthes: As `setNoteTextEditMode()` is set now in `readSettingsFromSettingsDialog()`

src/dialogs/settingsdialog.cpp
```diff
@@ -109,6 +109,8 @@ void SettingsDialog::storeSettings() {
                       QString::number(ui->toolbarIconSizeSpinBox->value()));
     settings.setValue("MainWindow/showRecentNoteFolderInMainArea",
                       ui->showRecentNoteFolderCheckBox->isChecked());
+    settings.setValue("MainWindow/setViewMode",
+                      ui->setViewMode->isChecked());
```
pbek: can you please name this checkbox something like `defaultViewModeCheckBox`

src/mainwindow.cpp
dasmaetthes: Yes encrypted editing still works

PR #70, opened 2016-02-14
Updated Russian translation

elnull:


pbek:
You really are a quick one!

Issue #168, opened 2016-05-16
Empty space in lists above last indented block


dwu:
QOwnNotes is a really impressive piece of software in terms of both features and development speed. Thank's a lot for developing it!

While importing a number of my previous notes in QOwnNotes I noticed a small inconsitency in the markdown rendering behavior.

If the last bullet of a bullet list is indented, there's an additional empty space added above the first line of the indented items.

![empty_space](https://cloud.githubusercontent.com/assets/91804/15281307/4139fa34-1b3a-11e6-8538-e322f5d612e7.png)

If the last bullet of a bullet list is not indented, the indented items are rendered as expected without the empty space being added.

![no_empty_space](https://cloud.githubusercontent.com/assets/91804/15281316/493786b6-1b3a-11e6-9969-779c8553427f.png)

pbek:
Thanks a lot for your kind words, @dwu!
Could you please also provide an example text (as text, not image).

dwu:
Thanks for the quick reply. Here the examples as text.

Empty space:
```
- Some text (level 1)
	- Some text
	- Some text
- Some text (level 1)
	- Empty space above this line
	- Some text
```

No empty space:
```
- Some text (level 1)
	- Some text
	- Some text
- Some text (level 1)
	- No empty space above this line
	- Some text
- Some text (level 1)
```

pbek:
Sadly the generated HTML block looks the same in both cases, although it gets differently rendered in the preview. I could not get around it by means of modifying HTML or CSS. It seems like an other quirks of the QTextEdit widget.

I will only be able to fix that if we use a full blown browser as preview, but that is currently very inconvenient for a lot of technical reasons... I'll put this issue on my list.
Thanks for reporting.

dwu:
I guess you're right. Browsing through the open bugs for QTextEdit I found the following reports, both reported in 2013.

- https://bugreports.qt.io/browse/QTBUG-32794
- https://bugreports.qt.io/browse/QTBUG-32046

pbek:
Oh well... That's three years ago. :)
Thank you for researching!

dwu:
As the incorrect rendering is also present in exported PDFs with rendered markdown and I don't expect the bug in QTextWidget getting fixed anytime soon, would you be open for a feature request for exporting a note's generated HTML representation to a file?

This way one could open the generated HTML in another application and convert to PDF from there.

pbek:
[...]

PR #173, opened 2016-05-18
Create HTML export from Hoedown output.

dwu:
Pull request for https://github.com/pbek/QOwnNotes/issues/168

Create HTML export output via `Note::toMarkdownHtml()` instead of exporting the content of the note preview as HTML. 

The change turned out to be a bit more invasive than I expected. To allow greatest flexibility for customizing the export stylesheet independently of the preview stylesheet while still keeping code duplication minimal the best solution I could come up with quickly was to extend the stylesheet creation code in `toMarkdownHtml` and add an additional parameter `forExport` (default: `false`) which defines whether the function is used in the preview or export context. I'd be happy to learn about better ways to implement this.

With the ability to adapt the CSS of the export easily, this implementation provides two additional features compared to yesterday's proof-of-concept:

- the font configured in property `MainWindow/noteTextView.font` is used as the font for the body of the exported HTML
- `maxImageWidth` is ignored if `forExport` is true; instead images are exported  with style `max-width: 100%;` to reduce their width if they're wider than the browser window

### DIFF
--- a/src/entities/note.cpp
+++ b/src/entities/note.cpp
@@ -809,9 +809,11 @@ bool Note::removeNoteFile() {
 /**
  * @brief Returns html rendered markdown of the note text
  * @param notesPath for transforming relative local urls to absolute ones
+ * @param maxImageWidth defined maximum image width (ignored if forExport is true)
+ * @param forExport defines whether the export or preview stylesheet
  * @return
  */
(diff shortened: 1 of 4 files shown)


pbek:
looking good, thank you. I will test it tomorrow!

pbek:
thanks a lot, @dwu!

pbek:
I just released v16.05.16 with it...

PR #227, opened 2016-06-22
Decreased default size of todo dialog

guija:
Decreased default size of todo dialog because on small screens e.g. with resolution of 1360x768 using the window manager i3 the dialog was not fully visible.


pbek:
Thank you, I hope Qt Creator or Qt are messing up the sizes later...

Issue #292, opened 2016-08-16
master ([3ce875d]) does not compile with MSVC 19


rwiesenfarth:
## Problem:

Compiling the master branch with MSVC 19 (and Qt 5.6.0, but this is unrelated) leads to an error:

**error C4576:** a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax

## Possible Solution:

replace line 14 in src/widgets/qownnotesmarkdowntextedit.cpp with

`// The initial define causes an error with MSVC 19:`
`// "error C4576: a parenthesized type followed by an initializer list is`
`// a non-standard explicit type conversion syntax"`
`// The replacement works, probably also on other platforms (to be tested)`
`//#define STY(type, format) styles->append((HighlightingStyle){type, format})`
`#define STY(type, format) styles->append({type, format})`

and accordingly line 62 in src/libraries/qmarkdowntextedit/markdownhighlighter.cpp

`// The initial define causes an error with MSVC 19:`
`// "error C4576: a parenthesized type followed by an initializer list is`
`// a non-standard explicit type conversion syntax"`
`// The replacement works, probably also on other platforms (to be tested)`
`//#define STY(type, format) styles->append((HighlightingStyle){type, format})`
`#define STY(type, format) styles->append({type, format})`

pbek:
Could you please make pull requests to qmarkdowntextedit and QOwnNotes, then we could see if the changes are working on other platforms.

rwiesenfarth:
Am 16.08.2016 um 09:35 schrieb Patrizio Bekerle:
> Could you please make pull requests to qmarkdowntextedit and QOwnNotes,
> then we could see if the changes are working on other platforms.

(English version below)

Hi,

kann ich (theoretisch) machen, ist für mich aber Neuland. Der naive
Ansatz (Push Request in Atlassian SourceForge) scheitert mit:

Permission to pbek/QOwnNotes.git denied to rwiesenfarth.

Muss ich dazu zuerst Dein repository bei github clonen?

Und: Ich muss ja dann einen zweiten Push-Request für qmarkdowntextedit
machen, weil das ja nur als submodule eingebunden ist!?

Grüße Rainer

===

Hi,

theoretically, I can, although it'll be a new experience. The naive
approach (pull request in Atlassian SourceForge) failed with:

Permission to pbek/QOwnNotes.git denied to rwiesenfarth.

Do I have to clone your repository at GitHub first?

Furthermore, I'll have to place a second pull request for
[...]

PR #293, opened 2016-08-16
#292 error when compiling with MSVC 19

rwiesenfarth:
Compiling with MSVC19 leads to

error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax


pbek:
looking good, I will merge it

Issue #344, opened 2016-09-30
USE_SYSTEM_BOTAN broken


chronitis:
The source tree includes support for using external libbotan instead of the vendored copy, but setting `USE_SYSTEM_BOTAN=1` currently breaks the build, since the `#ifdef` block this option enables doesn't contain enough header includes. Fixed by:

```
--- a/libraries/botan/botan.h
+++ b/libraries/botan/botan.h
@@ -22,6 +22,8 @@
 #include <botan/pubkey.h>
 #include <botan/rsa.h>
 #include <botan/ui.h>
+#include <botan/pbkdf2.h>
+#include <botan/sha160.h>
 #else
 
 #include <QtGlobal>
```

pbek:
Which source tree includes support for using external libbotan? What build breaks?
If you want to propose changes would you please care to post a pull request and explain why you want it changed?

chronitis:
Sorry, the original report was a bit unclear.

`src/libraries/botan/botan.pri` checks a qmake variable `USE_SYSTEM_BOTAN`, and if set tries to link against the system version of `libbotan` instead of compiling the copy of `libbotan` contained in that directory.

However, if you set the variable and try and build the project (`qmake; make` on a system with botan headers available) you get a bunch of undefined symbol errors - the above patch adds the missing headers to the list used when `USE_SYSTEM_BOTAN` is set - I'll make a pull request.

pbek:
Very interesting. As far as I remember I took the pre-assembled botan library from qtcreator, seems like they will have the same problem then! ;) Thank you for reporting and fixing!

PR #345, opened 2016-09-30
Add missing headers to allow build with USE_SYSTEM_BOTAN=1 to succeed

chronitis:
As per #344, this is a trivial patch that adds missing headers to the list used if the qmake variable `USE_SYSTEM_BOTAN` is set. If this variable is unset, this list is ignored, so there should be no wider consequences. This is (conceivably) relevant for distribution packaging where bundled libraries are not preferred.


pbek:
looking good, thank you!

Issue #450, opened 2017-01-02
No automatic sort when manually renaming a note


dwu:
Manually renaming a note when "Allow note file name to be different from headline" and alphabetical sort order are activated does currently not automatically trigger sorting the note tree. This means that the renamed note might end up in the wrong position. It would be great if sorting could happen automatically after renaming.

As renaming a note does not seem to change `currentNoteLastEdited` I assume that sorting is only required if `sortAlphabetically` is enabled. In this case the easiest solution to get the desired behavior would probably be to do something along the lines of:

```Diff
diff --git a/src/mainwindow.cpp b/src/mainwindow.cpp
index 2d36779..e78addb 100644
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -7534,8 +7534,11 @@ void MainWindow::on_noteTreeWidget_itemChanged(QTreeWidgetItem *item,
                 // handle the replacing of all note urls if a note was renamed
                 Note::handleNoteRenaming(oldNoteName, newNoteName);
 
-                // reload the directory list if note name has changed
-//                loadNoteDirectoryList();
+                // sort notes if note name has changed
+                if (sortAlphabetically) {
+                    ui->noteTreeWidget->sortItems(0, Qt::AscendingOrder);
+                    ui->noteTreeWidget->scrollToItem(item);
+                }
             }
         }
```

However, as there's currently a commented out call to `loadNoteDirectoryList` in `on_noteTreeWidget_itemChanged` it seems that it was originally intended to completely reload the note tree here. If that's still desired the following (with the call to `loadNoteDirectoryList` moved after the last access to `item` to make sure `item` still points to the original object) might probably be a better solution:

```Diff
diff --git a/src/mainwindow.cpp b/src/mainwindow.cpp
index 2d36779..71882db 100644
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -7506,6 +7506,8 @@ void MainWindow::on_noteTreeWidget_customContextMenuRequested(
  */
 void MainWindow::on_noteTreeWidget_itemChanged(QTreeWidgetItem *item,
                                                int column) {
+    bool noteRenamed = false;
+
     Q_UNUSED(column);
     if (item == NULL || !Note::allowDifferentFileName()) {
         return;
@@ -7534,8 +7536,7 @@ void MainWindow::on_noteTreeWidget_itemChanged(QTreeWidgetItem *item,
                 // handle the replacing of all note urls if a note was renamed
[...]

PR #451, opened 2017-01-02
Sort note tree when note is renamed manually

dwu:
Proposed temporary fix for #450. Not required anymore when `loadNoteDirectoryList();` is called.


pbek:
Thank you, I will test it later today.

pbek:
Looking good, thank you!

PR #533, opened 2017-03-03
Added new script

Maboroshy:
This script implements custom action buttons to prepend current note name with a special (or any other) unicode symbols so the note will be "pinned" on top of the list with "alphabetical sorting" on. You can use different symbols, one at a time, to "pin" notes. That allows to make kind of hierarchy of pinned notes.


pbek:
Great job, @Maboroshy!

PR #540, opened 2017-03-08
Add new script

Maboroshy:


pbek:
A small paragraph on how to create a dialog with a link to your script on <http://docs.qownnotes.org/en/develop/scripting/README.html> maybe? :innocent:

Issue #369, opened 2016-10-21
Sub-folders sorting mode


Maboroshy:
Looks like sub-folders are sorted by date rather than by name like the tags do. That makes them jump back and forth and doesn't permits the tricks like starting sub-folder name with "!" to put it to the first line. 

I'm not sure sorting by date is useful for a note app and propose sorting by name, same as tags.

pbek:
Tags are also sorted by date by default. This way the most recently modified sub-folder sits on top.

Maboroshy:
Tags should be sorted by names too then. :smile:

bdouxx:
There is an option to sort tag "by date" or "Alphabetical" can you add the same option for Subfolder?

I use QOwnNotes to stock all my cli command example, and it's more easy to find  what I want when it's sort "Alphabetical"

pbek:
I will look into that.

DenisRybin:
Just vote for sort folders feature, really needed.

pbek:
Pull requests are welcome...

pbek:
## 17.05.12
- a lot of work was done by @derboblan for the note list, tag tree and note
  subfolder panel, a big thank you for that
	- there now is a new page *Panels* in the *Settings dialog*
	- you can configure the sort order of the note list, tag tree and note
	  subfolder panel
	- sorting tags by last use is now possible
	- counting used tags in sub folders was also fixed
	- there are options to hide the search bar for tags and note subfolders
	- for note subfolders there are options to show the root folder name, to
	  display the note subfolders as tree with a root element and to show the
	  full path in the tree

There now is a new release, could you please test it and report if it works for you?

Maboroshy:
I've tried the new features, and they are top notch! Just what the app needs. Great job @derboblan!

I like new "hide search" feature too. I like too keep the app minimalistic, so I've hided everything. But I've encountered an issue with note create/search, which is note list pane issue really. It has no "create new note" in context menu. Sub-folder list have both "create sub-folder" and "create note", and note list doesn't have anything. 

Maybe "create note" better move to the note list context menu with the "enter name" pop-up, like the one sub-folder has.

pbek:
But you can create notes with `Ctrl+N`, can't you?

Maboroshy:
I do, also there's a toolbar button. I just post all the inconsistencies I find in app's UI. If remember correctly, now it has 5 times "stars" there were when I first joined this hub. I want it to have 5 times more! :sunglasses:

[...]

PR #588, opened 2017-05-20
Implementation of issue #369

derboblan:
Hey, saw the help wanted on issue #369 and wanted the feature myself, so here it is.

There are three new configuration options in Note -> Settings -> Interface to specify if the note, note subfolder and tag lists should be ordered alphabetically or by last change.


pbek:
Thanks a lot for your contribution!
A few things came to mind while looking at your code and testing...

You have created a lot of new methods to fetch node, tags or subfolders that basically only differ by the sorting in the query from their old version. Could you please merge those two different methods into one or was there a reason you duplicated the code?

There is already a sorting mechanism for sorting notes in the main menu (Note / Sort by). I think your new settings are conflicting with it. (you previously could also change the sort order to asc / desc there for alphabetically sorted notes).

Currently the tag sorting doesn't seem to do anything for me, regardless of what I choose the tags are always sorted alphabetically. The current method `Tag::fetchAllWithLinkToNoteNamesSortingAlphabetically` seems to do almost the same as `Tag::fetchAllWithLinkToNoteNamesSortingLastChanged`.

Sorting note folders by "Last changed" currently doesn't seem do anything when a note was changed.

derboblan:
> There is already a sorting mechanism for sorting notes in the main menu (Note / Sort by). I think your new settings are conflicting with it. (you previously could also change the sort order to asc / desc there for alphabetically sorted notes).

Ahhhh I never found that one, good to know. I Just had a quick look on friday and it seemed to me that the respecive lists were only build via the sql in their .cpp. Then I just copied and pasted the methods because I didn't want to change to much in your codebase and it was relatively quick. I didn't write gui code in years and were happy this pain was mostly spared. But anyways, I completely rewrote the feature to properly use the widgets sorting mechanisms.

> There is already a sorting mechanism for sorting notes in the main menu (Note / Sort by). I think your new settings are conflicting with it. (you previously could also change the sort order to asc / desc there for alphabetically sorted notes).

They are in sync now, plus you can order the note subfolders and tags asc / desc too (had to put them one level lower in their trees, otherwise the other entries would have gotten sorted aswell)

> Currently the tag sorting doesn't seem to do anything for me, regardless of what I choose the tags are always sorted alphabetically.

That's because you were already ordering priority asc, name asc. I think there should have been a difference at some point with lots of tags with different priorities. I changed that back to priority only for now, because you can set the different order in the settings now.

> Sorting note folders by "Last changed" currently doesn't seem do anything when a note was changed.

This bug was there before; I am able to reproduce it with an older version. Sorting only gets applied when the widget reloads (e.g. after opening+closing the settings, reloading the note folder, or restart) 

regards

pbek:
- I like your new settings panel, you figured out a lot by yourself! great job! :+1: 
	- you added an icon `checkbox_indeterminate.png` for it, that will cause problems
	- I usually take a svg icon (that is scalable) from the Breeze icon theme and add both a "light" and "dark" variation (for the dark mode) to the corresponding resource file, I also set the free desktop icon identifier for it so that the the operating systems can choose there own icon set
		- I could do that myself if you have troubles with it
	- we also need a new constant in the enum `SettingsDialog::SettingsPages`
- it seems that the default note sort order is now alphabetically / ascending, we should not modify the default behaviour of QOwnNotes (that was `date / descending`)
- you did remove the old settings `SortingModeAlphabetically` and `NoteSortOrder`, did you?
	- we will have to write a migration for it so that existing users don't loose their current sort order settings
	- I usually do that in `DatabaseService::setupTables`
	- ping me if you need help with that

- the tag priority isn't used yet and possibly will never be used, so currently there is no other sorting than alphabetically
	- sorting by date (note modification date) would be very difficult because the notes sit in an other in-memory-database and the tags are in a per note folder disk-database (per note folder)
	- I don't know what we should do about your settings in that regard
- a real sorting of note subfolder by change date of notes that always works is a bit more complex and was never implemented
	- currently it only works as you stated
	- I also don't know what we should do about your settings in that regard

derboblan:
> I like your new settings panel, you figured out a lot by yourself! great job!

thanks :smiley: 

> you added an icon checkbox_indeterminate.png for it, that will cause problems

ok, I tried to change that -- is it correct? I'm not sure if the right icon is shown in dark mode.

> it seems that the default note sort order is now alphabetically / ascending, we should not modify the default behaviour of QOwnNotes (that was date / descending). you did remove the old settings SortingModeAlphabetically and NoteSortOrder, did you?

Yes I wanted all the sort and order options to have similar names. Added the conversion and setting of defaults that mimic the old behaviour.

> Sorting by date (note modification date) would be very difficult because the notes sit in an other in-memory-database and the tags are in a per note folder disk-database (per note folder)

I tweaked the tag.cpp to get somewhat close. Under this configuration we can sort at least by the last modification or use of the tag, what do you think?

> a real sorting of note subfolder by change date of notes that always works is a bit more complex and was never implemented

do you mind leaving it at that as it is the old behaviour? for me the alphabetical sort order is the important thing here -- the other one is just to confusing.

regards.

src/services/databaseservice.cpp
```diff
@@ -523,6 +523,46 @@ bool DatabaseService::setupTables() {
...
+            settings.remove("SortingModeAlphabetically");
+        }
+
+        if (settings.contains("NoteSortOrder")) {
+            int order = static_cast<Qt::SortOrder>(settings.value("NoteSortOrder").toInt());
+            settings.setValue("notesPanelOrder", order); // see defines in MainWindow.h
+            settings.remove("NoteSortOrder");
+        }
+
+        // set defaults for now settings iff not set already
```
pbek: you have a typo here ;)

src/entities/tag.cpp
```diff
@@ -178,8 +178,15 @@ QList<Tag> Tag::fetchAll() {
 
     QList<Tag> tagList;
 
-    //query.prepare("SELECT * FROM tag ORDER BY priority ASC, lower(name) ASC");
-    query.prepare("SELECT * FROM tag ORDER BY priority ASC");
+    //query.prepare("SELECT * FROM tag ORDER BY priority ASC, name ASC");
```
pbek: can you please document what your are doing here, and why

src/entities/tag.cpp
```diff
@@ -197,9 +204,17 @@ QList<Tag> Tag::fetchAllByParentId(int parentId) {
     QSqlQuery query(db);
     QList<Tag> tagList;
 
-    query.prepare("SELECT * FROM tag WHERE parent_id = :parentId ORDER BY "
-                          //"priority ASC, lower(name) ASC");
-                            "priority ASC");
+    //query.prepare("SELECT * FROM tag WHERE parent_id = :parentId ORDER BY "
+    //                      "priority ASC, name ASC");
+    query.prepare("SELECT t.id as id, t.name as name, t.priority as priority, "
```
pbek: can you please document what your are doing here, and why

src/entities/tag.cpp
```diff
@@ -469,7 +482,8 @@ bool Tag::store() {
     if (this->id > 0) {
         query.prepare(
                 "UPDATE tag SET name = :name, priority = :priority, "
-                        "parent_id = :parentId, " + colorField + " = :color "
+                        "parent_id = :parentId, " + colorField + " = :color, "
+                        "created = datetime('now') "
```
pbek: I think it's very misleading your your are using the column `created` as a update date, we should use a column `updated`

src/services/databaseservice.cpp
```diff
@@ -523,6 +523,46 @@ bool DatabaseService::setupTables() {
         version = 24;
     }
 
+    if (version < 25) {
+        // migrate old sort and order settings + set defaults iff unset
```
pbek: `iff` typo

src/dialogs/settingsdialog.h
```diff
@@ -48,6 +48,7 @@ Q_OBJECT
         ToolbarPage,
         DebugOptionPage,
         EditorPage,
+        PanelsPage,
```
pbek: `PanelsPage` needs to be behind `GitPage` and your panel needs to have a `whatsThis` of 16 instead of 15 (because  `GitPage` already has 15)

pbek:
You are doing some interesting things here. :)
Currently I seem to have the problem that tags are duplicated in the tag tree (and even the context menu to add tags) when I add (link) an existing tag to a note.

pbek:
Keep in mind that took on the quest of rewriting the most critical parts of the application: listing of notes, tags and subfolders. If anything goes wrong there a big outcry will he the consequence for us so we have to be extra careful. :smile_cat: 

A lot of things can trigger the listing, for example adding / removing / modifying / moving of tags, adding / removing / modifying / moving / copying of notes, adding / removing / modifying / moving of sub folders, external modifications of the note folder, its subfolders or the current note... just to name a few.

src/dialogs/settingsdialog.h
derboblan: I think the order in an enum does not matter; I put it there because in the QT Creators Object Tree the panelsPage appears before the gitPage and so both are aligned.

src/services/databaseservice.cpp
derboblan: iff is the shorthand for "if, and only if" -- still want that changed?

src/dialogs/settingsdialog.h
pbek: the enum is autoincrementing from one constant to the next and the number has to be corresponding to the "whatsThis" number, the position you took was already taken

src/services/databaseservice.cpp
pbek: please don't use abbreviations in documentation :)

src/entities/tag.cpp
derboblan: Yeah, i somewhat abuse that a bit. But I didn't see it in use so far and it seemed perfect for what I wanted to do. One can also argue, that if you rename a tag for instance, you basically create a new one with the same id while deleting the old one, now there's only confusion, when updating a child tag :D. Introducing another column didn't seem that great for me, because i think it will mostly hold duplicate data (except if you make heavy use of managing your tags in a tree, or keep renaming them all the time). I would rather vote to rename created to updated.

pbek: > One can also argue, that if you rename a tag for instance, you basically create a new one with the same id while deleting the old one

where did I do that?
here is where renaming is done: <https://github.com/pbek/QOwnNotes/blob/develop/src/mainwindow.cpp#L6522>

> I would rather vote to rename created to updated.
The problem is that there is no easy way to rename columns in SQLite...

derboblan:
No problem and it's very nice to have someone else test that, who uses the program differently then me. right now i just saw that the note count was broken and I will look into that.

I'm confident this will be working reliably sometime.

src/entities/tag.cpp
derboblan: No, not in your code, I mean more general, philosophically speaking: "If something is change the old is destroyed and something new emerges" :)

I wasn't aware of that problem in sqlite. However, it's possible to create a new table with the desired structure, convert the values and drop the old table. The question stands: do you prefer saving (imho mostly duplicate) data over that?

pbek: The created date currently (in the `pbek` repository) holds the date when the entity was created (set in the table create statement) and should not be misused as `updated` field. I guess the best would be you add a field per entity where you need it that really reflects what you are storing into it. Since you are setting it in the `store` method the name `updated` would in my opinion be appropriate.

derboblan:
Now the tag counts are correct almost all the time after selecting another note subfolder. But for the special item "All Notes" the counts and tag filtering is the same as for the root note folder -- which also applies to your develop branch.

derboblan:
> I didn't really....when i inserted the settings page, I managed to put it before the git page. After you mentioned the issue about the enum, i changed it there and set the tooltip to 16 and ended up getting the git page, when clicking on Panels and vice versa :(. had to moved the settingsPage behind the git page to correct that mistake afterwards.

pbek:
I can't wait to test your changes when you took a look at the adding of the `updated` field. :+1: It will be a wonderful release.

src/services/databaseservice.cpp
```diff
@@ -243,6 +243,40 @@ bool DatabaseService::setupNoteFolderTables() {
         version = 11;
     }
 
+    if (version < 12) {
+        // create new tag table
+        queryDisk.exec("ALTER TABLE tag RENAME TO _tag");
+        queryDisk.exec("CREATE TABLE IF NOT EXISTS tag ("
```
derboblan: Because sqlite does not support to add columns with non-zero default :( And before filling it is forgotten later, I wanted it to be automatically set.

pbek: ah, what a bummer. :( good to know

derboblan:
i always tried to keep the acummulated diff as close to your original source, as possible using https://github.com/pbek/QOwnNotes/compare/develop...derboblan:develop more or less ignoring the direct update diffs for my commits :)

src/services/databaseservice.cpp
pbek: But I think you have created an index wrong, it has to be `CREATE UNIQUE INDEX IF NOT EXISTS idxUniqueTag ON tag (name COLLATE NOCASE, parent_id)` (see version 11). 

And could you please document why you did the migration at version 12 in the code, so it's clear why we didn't just `ALTER TABLE`.

pbek:
So far only the wrong index is left before I'll give it a go again. ;)

derboblan:
I put the collate nocase into the table definition, which does the same (after analysing how the name field was used in other queries) and should be a little bit better performancewise, see https://stackoverflow.com/questions/973541/how-to-set-sqlite3-to-be-case-insensitive-when-string-comparing#973785

pbek:
@derboblan I just tested your latest features and fixes and I'm really impressed, you did an amazing job there! 🎉 

I only found one thing that will cause a small uproar. :)
You turned on the root element for the tag tree and sub folder tree, which is nice in a way and back then I did that too. But then people with very confined screen space complained about the extra space that took and so I optimized everything to make it run without a root element.
I saw you had a setting for to turn that on/off for sub folders, but that setting didn't do anything for me, the root element was always there.
If you really want root elements we need settings for both tag tree and sub folder panel, default has to be *no root element*.

Thank you for all your effort and heart-blood you put into those panels!

pbek:
I really hope to get your code into the repository soon. Maybe it would be good thinking if you stop making further (very welcome) improvements in this pull requests, so I can merge your code. The only thing that is left are settings for the root elements. :smile:

derboblan:
I'm not quite sure I completely understand what you mean. Do you only want the first level to be without parent (like right now in the pbek repository, see the attached screenshot) or should both panels rather be like the notes panel with no hierarchy at all? I guess the tree view will be another option to enable then ;) And yeah, you're right, I can stop my work on the pull after that -- except for bug fixes :) Last night I just added the possibility to hide the searches in the note subfolders and tags panels, since I don't use them and am more concerned with vertical then horizontal space on the screen :D that will be in the push aswell.

image

pbek:
The default behavior should be like it is in the pbek repository. The first level of tags and note subfolders is on the same level as the `All notes` item.

pbek:
Find line edits: I guess you can also do such user interface modifications with the *Custom user interface styling* script from the script repository... ;)

derboblan:
Ok, i've restored the root item behaviour as requested. I've also merged your latest changes, so everything should be up-to-date. Unless there are any bugs left I'm finished here.

pbek:
looking good so far, I will attempt a merge today

pbek:
Damn, under Linux I get random segmentation faults when I create notes or tag a note. Did you do something unhealthy with timers?

pbek:
The crash seems to happen when the tag tree is built.

pbek:
Very strange, I wasn't able to reproduce this on an other Linux system and after a reboot I haven't had it any more on the first system...

pbek:
## 17.05.12
- a lot of work was done by @derboblan for the note list, tag tree and note
  subfolder panel, a big thank you for that
	- there now is a new page *Panels* in the *Settings dialog*
	- you can configure the sort order of the note list, tag tree and note
	  subfolder panel
	- sorting tags by last use is now possible
	- counting used tags in sub folders was also fixed
	- there are options to hide the search bar for tags and note subfolders
	- for note subfolders there are options to show the root folder name, to
	  display the note subfolders as tree with a root element and to show the
	  full path in the tree

pbek:
@derboblan congratulations for your first successful pull request for QOwnNotes! You did a great job there, I really like it! :tada: 
Feel free to suggest more thing!
If you also want to translate your additions: <https://crowdin.com/translate/qownnotes/1/en-de>

derboblan:
Ok, I'll take a look at the translations tomorrow . Thanks, i just really like your program :D I've already been following the issues the last days. Right now I've got something else to finish the next days, but for after that I've already got some ideas what else could be done.

pbek:
Thanks a lot! :)

PR #591, opened 2017-05-26
Desktop file fixes

chronitis:
Fixes a couple of minor deviations of the `QOwnNotes.desktop` file from the desktop file spec.

 * Add `Keywords` (used by some desktop environments to filter applications as you type, eg GNOME)
 * Remove `Encoding` (obsolete [1])
 * Modify the `Categories` list (the given values didn't match the registered values [2])

[1]: https://specifications.freedesktop.org/desktop-entry-spec/1.0/apc.html
[2]: https://specifications.freedesktop.org/menu-spec/1.0/apa.html

### DIFF
--- a/src/QOwnNotes.desktop
+++ b/src/QOwnNotes.desktop
@@ -8,4 +8,4 @@ Type=Application
 Terminal=false
 StartupNotify=true
 Encoding=UTF-8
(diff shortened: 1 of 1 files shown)


pbek:
Thanks a lot for your corrections, @chronitis!

Issue #689, opened 2017-08-22
Cursor relocated before title while editing


SmartBlug:
#### Expected behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
Keep same position

#### Actual behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
Cursor relocated before first char on first line

#### Steps to reproduce                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
Create a note
Edit
Wait until sync while editing
one console log : 
[09:04:14] [status] Stored 1 note(s) to disk
[09:04:15] [status] Current note was modified externally
[09:04:15] [status] Loading external changes from: 2017-08-22 - 08-58-03 - Renault - Espace Initiale .md
and then cursor relocated to 0,0
to reproduce again,
go somewhere on the note
type a key, wait owncloud to save and hop... again 0,0 !

Bug in 17.08.6 and 17.08.7

#### Output from the debug section in the settings dialog

#### Relevant log output (you have to enable enable the log widget)

pbek:
please see #656 ;)

SmartBlug:
True, it's related to external modification as in #656. 
I'll Fork and try to find the bug ASAP.
Thanks

pbek:
That would be great, because I never could reproduce that issue that some users have.

SmartBlug:
I found the problem, it's related to CR/LF on the file when saved on WINDOWS and not on the editor (just LF). 
When you check if notes are identical (line 2264 of mainwindow.cpp), it's false of course.

On Mac, only LF is saved so no pb
I've not checked on Linux yet but should be OK too.
You have a workaround if you save using UNIX newline in Settings/General

I've added the missing part in the code to remove CR before comparison and that solve the problem ;-)

pbek:
Sounds awesome!

So the problem only happened when a note was created on Windows and outside of QON?

pbek:
## 17.08.8
- added a fix by @SmartBlug for the sporadic cursor-jumping to the top in the note edit

pbek:
There now is a new release, could you please test it and report if it works for you?

pbek:
it didn't work for #656, but this could also be something else...

pbek:
It neither seems to fix #659...

ludin:
I have upgrade to 18.08.9 and enabled the option in the general settings:
"restore the current cursor position when opening a note"

The problem with the jumping cursor still exists for me:
[...]

PR #691, opened 2017-08-22
Solve bug #689 - File found to be different just after save on windows if no UNIX newline

SmartBlug:
I found the problem, it's related to CR/LF on the file when saved on WINDOWS and not on the editor (just LF). 
When you check if notes are identical (line 2264 of mainwindow.cpp), it's false of course.

On Mac, only LF is saved so no pb
I've not checked on Linux yet but should be OK too.
You have a workaround if you save using UNIX newline in Settings/General

I've added the missing part in the code to remove CR before comparison and that solve the problem ;-)


pbek:
I now committed your fix myself to include it in today's release to test ist with more users. Thanks a lot so far!

PR #720, opened 2017-09-18
Update link in README.md to link to EN

sanderboom:
Update link in README.md to link to EN page instead of the DE page.


pbek:
thx

Issue #857, opened 2018-02-19
Image import fails when inserting HTML or image from clipboard


echo511:
#### Expected behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
Should store images into media/.

#### Actual behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
In the bar at the bottom it says the image was downloaded but nothing is actually imported - no file in media folder, no snipped in note inserted.
                                                                                                                                                                                                                                                                                                                                                                                          
#### Output from the debug section in the settings dialog
**Current Date**: `po úno 19 19:26:38 2018`
**Version**: `18.02.3`
**Build date**: `Feb 19 2018`
**Build number**: `417`
**Platform**: `linux`
**Operating System**: `Arch Linux`
**Build architecture**: `x86_64`
**Current architecture**: `x86_64`
**Release**: `AUR`
**Qt Version (build)**: `5.10.1`
**Qt Version (runtime)**: `5.10.1`
**Portable mode**: `no`
**Settings path / key**: `/home/.../.config/PBE/QOwnNotes.conf`
**Application database path**: `/home/.../.local/share/PBE/QOwnNotes/QOwnNotes.sqlite`
**Application arguments**: `QOwnNotes`
**Qt Debug**: `0`
**Locale (system)**: `cs_CZ`
**Locale (interface)**: *empty*
**Icon theme**: `Numix-Square`
**Notes in current note folder**: `16`
**Enabled scripts**: `1`


#### Relevant log output (you have to enable enable the log widget)
```
[úno 19 19:29:12] [debug]: insertHtml  - 'imageUrl':  QUrl("https://avatars1.githubusercontent.com/u/658504?s=88&amp;v=4")
[úno 19 19:29:12] [status]: Downloading https://avatars1.githubusercontent.com/u/658504?s=88&amp;v=4
[úno 19 19:29:17] [status]: Downloading images finished
```

pbek:
When I copy your image from this webpage with a bit of other text on Linux I get

```markdown
AssigneesNo one—assign yourself

LabelsNone yetProjectsNone yetMilestoneNo milestoneNotifications UnsubscribeYou’re receiving notifications because you’re subscribed to this repository.

1 participant[![media-AM2270](file://media/1074265970.png)
[...]

PR #858, opened 2018-02-20
fix image import via pasting HTML content from clipboard

echo511:
Solves https://github.com/pbek/QOwnNotes/issues/857

Maybe related to https://bugreports.qt.io/browse/QTBUG-58254.


pbek:
Thank you, but I've already worked on that but not yet committed

PR #862, opened 2018-02-22
do not remove double "code blocks" (<pre><code>)

mdealencar:
I don't see why do that:
```
result.replace("<pre><code>", "<pre>").replace("</code></pre>", "</pre>");
```
It makes html export poorer (not possible to style code blocks) and also prevents assigning a language to the code block.

When entering this markdown:
```
\```python
def myfun():
    pass
\```
```
The tag is actually `<code class="language-python">` and the replacement generates broken html since it removes the closing `</code>` but not the opening one.

### DIFF
--- a/src/entities/note.cpp
+++ b/src/entities/note.cpp
@@ -1650,10 +1650,6 @@ QString Note::textToMarkdownHtml(QString str, QString notesPath,
                     " border-radius: 3px; color: %2; }").arg(
             codeBackgroundColor, codeForegroundColor);
 
-    // remove double code blocks
-    result.replace("<pre><code>", "<pre>")
-            .replace("</code></pre>", "</pre>");
-
     // correct the strikeout tag
     result.replace(QRegularExpression("<del>([^<]+)<\\/del>"), "<s>\\1</s>");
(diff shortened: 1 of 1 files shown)


pbek:
If I remove that lines I get an additional line at the end of the code block in the preview.

pbek:
Don't you get the same result?

pbek:
`<code>` by itself is used for single lines, we could use a regular expression to also remove the `class...`, which of course I also dislike...

pbek:
removing the `pre` from the stylesheet solves the problem, I will do that

mdealencar:
I do get the same results. Seems like a good solution, thanks!

Better yet would be to move the stylesheet out of the code, to some config file.

pbek:
The styles aren't static. They will be generated by code.

mdealencar:
I saw that. It could be a css-like file with placeholders for the configurable settings. The code would then search and replace a dictionary of {placeholder:setting}. It would make it more flexible, but it's just an idea.

pbek:
If you want to overwrite the style sheets you can do so much easier with
the help of the scripting engine. 😉

PR #930, opened 2018-06-04
Preparing for submission to flathub

stobbsm:
Flatpak manifest file added.
Fixed qownnotes.appdata.xml so it will properly validate.

I'm also going to open a pull request to master to fix the appdata.xml file. If it doesn't validate properly for flathub, they will not accept it.


pbek:
thanks a lot!

PR #1012, opened 2018-09-01
Fix the use of snap with Wayland

### LINKED ISSUE #1004: Application failed to start because it could not find or load the Qt platform plugin "wayland-egl"
#### Expected behaviour
The application starts                                                                                                                                                                                                                                                                                                                                                                                                                                                                               

#### Actual behaviour
The application does not start.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               

#### Steps to reproduce
1. Install the qownotes snap.
2. Start qownotes.

#### Relevant log output

```
$ qownnotes 
Fatal: This application failed to start because it could not find or load the Qt platform plugin "wayland-egl".

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, xcb.

Reinstalling the application may fix this problem. ((null):0, (null))
Avortat (s'ha bolcat la memòria)
```

I'm on Ubuntu 18.0.1 and using Gnome + Wayland and the snap has been working OK till the last update (rev 1087)

pachulo:
Add qtwayland5 in the stage packages but disable native wayland use for now with env `DISABLE_WAYLAND=1`
Fixes #1004 
More info here:
* https://forum.snapcraft.io/t/problem-launching-qt-snaps-in-wayland/7055
* https://github.com/ubuntu/snapcraft-desktop-helpers/pull/143

### DIFF
--- a/build-systems/snap/snapcraft/snapcraft.yaml
+++ b/build-systems/snap/snapcraft/snapcraft.yaml
@@ -12,6 +12,8 @@ description: |
 apps:
   qownnotes:
     command: desktop-launch ${SNAP}/usr/bin/QOwnNotes -style=Fusion --snap
+    environment:
+        DISABLE_WAYLAND: 1
     # see https://docs.snapcraft.io/reference/interfaces
     plugs:
       - x11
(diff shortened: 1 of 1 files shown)


pbek:
I hope this doesn't break X11 only systems...

PR #1016, opened 2018-09-07
Create QOwnNotes.svg

ChrisP4:
Created this icon that fits with the elementaryos/HIG stuff. (I like their icons).
I use KDE, so this icon works perfectly on KDE. I created it to use it on my desktop but thought to upload it here, and you can decide.


pbek:
Thank you, it looks quite nice! Thank you! I'm curious if I will use it in some way in the future.

PR #1043, opened 2018-10-11
Fix some spelling errors.

flopp:
This PR fixes some spelling errors found with `codespell`.


pbek:
Thank you for the effort! If you still have that call somewhere you could maybe post it for future reference.

PR #1126, opened 2019-01-27
Fixed gif not playing on Linux/macOS

liulex:
Tested on macOS, but I do not have a Linux environment. Could you please test it on Linux to see if it works now?


pbek:
I will when I'm back on my dev machine

pbek:
It works for me under Linux! :tada:  Great work!

pbek:
macOS also works!

liulex:
🥂

Issue #1142, opened 2019-02-05
[Bug] Note lost after changing its title and switching to another note


liulex:
Steps to reproduce:
1. Change note A's **title**.
2. Before the auto save timer times out, switch to another note, say note B.
3. Wait until the auto save timer fires.
4. Change the content of note B.
5. Wait for the time out of the auto save timer.
6. Boom. Note A is lost!

![note_lost_bug](https://user-images.githubusercontent.com/2010459/52283366-4a05a900-299d-11e9-84b0-0febd67f8bc4.gif)

pbek:
Thank you for reporting, I will try to reproduce your issue.

pbek:
Phew, seems to be a complicated one. Some kind of race condition. At least the removed note appears in the local trash for me.

pbek:
Seems to be this issue: #522

liulex:
Yes, #522 is the same problem.
I just found a fix #1143. It works for me, but I'm not sure if it would cause other problems, could you please have a check?

pbek:
## 19.2.2
- fixed lost note after changing its title and switching to another note
  (for [#522](https://github.com/pbek/QOwnNotes/issues/522) and
  [#1142](https://github.com/pbek/QOwnNotes/issues/1142), a big thank you to @liulex)

pbek:
There now is a new release, could you please test it and report if it works for you?

liulex:
Yes, it's fixed now ;)

pbek:
Thank you again for your help! :+1: You really did a great job in finding and fixing the bug!

Issue #522, opened 2017-02-15
Data loss while switching between notes during header refresh


szotsaki:
Preconditions: 
* Have two longer markdown texts with `=======`-style headers on the top. 
* Headers must totally differ from each other and 
* the content should be easily distinguishable.

1. Copy the first text into a new note
2. Wait until the note name (in the list on the left side) gets refreshed with the header name
3. Paste the second text into a new note
4. Now, quickly, append some random string to the `=====` header and before the note's name gets refreshed, switch to the first note. Try switching back, modify headers and switch again. Try also Alt+Tabbing.

The result: when note-switching occurs during note-name refresh, for some reason, the second note fully gets replaced by the first one, resulting in data loss. Interestingly, the second note's header will also be the same as the first one's but with the current dateTtime appended.

pbek:
Thank you for reporting, I will try to reproduce your issue.

pbek:
Currently I'm not able to reproduce this again (a lot changes were made in the app meanwhile). Are you still able to reproduce it, @szotsaki?

szotsaki:
Unfortunately, I still can reproduce it relatively easily mostly as described in the very first comment.

How to reproduce easier:
- Have two, really big notes with a lot of text and many pictures (40+) inside.
- Create a copy from both of them on the file system and let their filenames be the same just prepend "ALTERNATIVE_" to them
- Quickly iterate between the alternatives back-and-forth with mouse in the note search panel. You'll notice a tiny lag in the application until a note is fully loaded.
- Now, just start typing random gibberish very quickly into the active note while continuing the switching between them.

After a bit of time, you'll see that the notes got a new name and a new content.

OS: I'm using QOwnNotes on Linux.

pbek:
Thank you for testing!

photonxp:
Here is how I could reproduce this problem:

#### STEP 1:
Create 3 notes. 
The first two notes are default new note:
-    "Note 2018-01-03T20.30.20"
-    "Note 2018-01-03T21.07.30"

The third note is a large note:
-    "largenote..2018-01-03T21.16.45"

#### STEP 2:
Then change the title/name section of the third note (large note). 
Meanwhile switch between the first two notes with cursor to change current item. Do this step early enough before the third note (large note) is stored into the disk.

Now the third note is quite likely to be replaced by one of the first two default notes. Also a new time-stamp would be appended.

[...]

PR #1143, opened 2019-02-06
Fixed note lost #522 #1142

liulex:
Before overriding the current note, check if they have the same id. Not sure if this is the right way to do, but it did fix the problem.


pbek:
It really solved the problem for me too, that would be too good to be true! :tada: I hope there are no other "side-effects" :smile: Thanks a lot!

Issue #1140, opened 2019-02-03
QOwnNotes resets all settings when notes folder is unavailable


toxpal:
Using QOwnNotes on antergos KDE for some time, and noticed that every single time QOwnNotes resets all settings to default when notes folder is unavailable. I really mean all - theme, toolbars, layout, every single option is reset.

I run antergos on SSD, and store notes on Luks RAID (2 x HDD in mirror mode, encrypted with Luks and auto-mounted during boot). Sometimes, after system update (like 1 out of 50 times), antergos fails to automount encrypted RAID during first boot. If I launch QOwnNotes then, it displays default startup window, and all settings are lost.

I then reboot OS, try to run QOwnNotes again, but default startup window is still here, and it doesn't allow me to start using app unless I choose default folder for notes, specify theme, etc. In other words, I must do everything from scratch.

I did some tests and noticed the same happens if I manually delete/rename default notes folder and run QOwnNotes then - all settings are reset again. Restoring original folder doesn't help - I still need to go through default startup wizard. It really drives me crazy...

I found a similar issue here - https://github.com/pbek/QOwnNotes/issues/370 - but it's not exactly the same.

`QOwnNotes Debug Information
===========================

## General Info

**Version**: `19.2.0`
**Build date**: `Feb  1 2019`
**Build number**: `517`
**Platform**: `linux`
**Operating System**: `Antergos Linux`
**Build architecture**: `x86_64`
**Current architecture**: `x86_64`
**Release**: `AUR`
**Qt Version (build)**: `5.12.0`
**Qt Version (runtime)**: `5.12.0`
**Portable mode**: `no`
**Settings path / key**: `/home/myusername/.config/PBE/QOwnNotes.conf`
**Application database path**: `/home/myusername/.local/share/PBE/QOwnNotes/QOwnNotes.sqlite`
**Application arguments**: `/usr/bin/QOwnNotes`
**Qt Debug**: `0`
**Locale (system)**: `en_US`
**Locale (interface)**: *empty*
**Icon theme**: `breeze-qownnotes`
**Notes in current note folder**: `3`
**Enabled scripts**: `0`

## Server Info

**serverUrl**: *empty*
**appIsValid**: `no`
**notesPathExists**: *empty*
**connectionErrorMessage**: *empty*

## Note folders

**currentNoteFolderId**: `0`

### Note folder `QOwnNotes`

**id**: `1`
**isCurrent**: `no`
**activeTagId**: `-2`
[...]

PR #1145, opened 2019-02-09
Fixed layout settings lost if note folder is not found #1140

liulex:
This PR does two things:
1. Do not always reset layout settings before showing up the welcome dialog.
2. If note folder is not found at startup while user settings are present, allow the user to only re-select the note folder and then finish the welcome dialog, so that all other settings keeps untouched.


pbek:
Sounds good, will test later.

Issue #235, opened 2016-07-02
Add checkbox rendring


chriswininger:
On github in your markdown you can do
`- [X] Does it work?`

The resulting text will be rendered using a checkbox (as you see below)
- [X] Does it work?

This is really useful for being able to create todo lists, I know QOwnNotes already has a todo list featured tied to the calendar, but I don't necessarily need calendar integration, I just want to be able create simple grocery style lists in the text editor.

chriswininger:
great product by the way, I've tried several other markdown based note taking tools recently, this one seems to outshine the rest.

pbek:
Thank you for your kind words!

> inline todo list

I'm sorry. As already thought over in #211 that doesn't is possible technically and doesn't make a lot of sense to me neither...

chriswininger:
Cool, thanks for feedback, I understand if it's not a desired feature. Technically it seems like you should be able to do it through hoedown extension. I found this https://github.com/kjdev/hoextdown which claims to do it. My C++ is a bit rusty, I'm a JS guy. I made a first stab at using that extension in the QOwnNotes project. I got the project to build but did not see any lovely checkboxes in the preview I had hoped. Oh well.

pbek:
I don't think the QTextEdit, that is used for the preview is able to view html checkboxes. I will put that on my list if we use a full browser component some day. 

You can also experiment by altering the generated html code for the preview in a Script.
You might take a look at `function noteToMarkdownHtmlHook` in 
<https://github.com/pbek/QOwnNotes/blob/develop/doc/scripting/example.qml>.
You could try to replace `- [X]` with a html checkbox to test it...

chriswininger:
Ah, you are correct. Looks like QTextEdit can not render them.

pbek:
We have to wait then... ;)

BackSeat:
Another vote for this when possible.

[Zim](http://www.zim-wiki.org/) does this nicely, and converts '()' to an unchecked checkbox, so nice in meetings to just type:

() do this

pbek:
Currently you would have to vote at Qt for supporting it in QTextEdit first... ;)

tito:
In the meantime, for everybody that want this feature, unicode can safe your time. Just put in your document at the top both symbol ☑☐. Then just use it.

tito:
And maybe one thing in qownnotes would be cool is to have an option that when we click on that kind of box, it would switch with the other one. It's a low cost rendering, but still in text and would work ^^

[...]

PR #1155, opened 2019-02-20
Support pdf export with checkboxes #235

liulex:
But I cannot reproduce this:
> for me all checkboxes are visible, but **unchecked**

Anyway, please test with this patch.


pbek:
Thank you for your quick implementation!

For some reason I can only see unchecked checkboxes, the checked ones are not present.

[Checkbox Test.pdf](https://github.com/pbek/QOwnNotes/files/2888127/Checkbox.Test.pdf)

liulex:
It should be a font issue. Copy this character
default
and paste it into browser, you will get ☑.

pbek:
Now, that is funny. I can see them on Manjaro, but not on KDE Neon. Could really be a font issue.

pbek:
Anyway, great work! I love it!

Issue #1174, opened 2019-04-08
 build-tuxfamily-src.sh question


dandelionred:
```
# checkout the source code
git clone --depth=50 git@github.com:pbek/QOwnNotes.git QOwnNotes -b $BRANCH
cd QOwnNotes

# checkout submodules
git submodule update --init

# remove huge .git folder
rm -Rf .git
```

`.git` dir is removed after `git clone` so what is the point of **--depth=50**?

pbek:
this is just one of my internal deployment scripts

> .git dir is removed after git clone so what is the point of --depth=50?

the point is to not get all commits, but only the last few

dandelionred:
You've got me wrong. I meant why is it not --depth=**1** but **50**?

pbek:
no particular reason, `1` would be fine too, does it make any difference to you?

dandelionred:
> no particular reason, 1 would be fine too, does it make any difference to you?

I see depth 50, init submodules below and wipe .git altogether below. And I think maybe it is related to submodules somehow I'm not aware of?
If it was just 1 (which is what you mean by "the point is to not get all commits, but only the last ~~few~~") there we'd not spend a few minutes discussing it :)

pbek:
I need to do a git checkout to fetch the sub-modules.

Forget about the `depth 50` I just took a parameter example and never changed it, the overhead is minimal.

dandelionred:
The overhead is the time spend on the "why 50?!" question :)

It is not in a single place btw and not 50 everywhere. There are depth=5 as well. Mby make it all depth=1 for a cleaner code and less stupid questions?

```bash
$ rgrep -F depth= 
ubuntu-launchpad/build-for-launchpad.sh:git clone --depth=50 git@github.com:pbek/QOwnNotes.git $PROJECT_PATH -b $BRANCH
ubuntu-launchpad/build-for-launchpad-trusty.sh:git clone --depth=50 https://github.com/pbek/QOwnNotes.git $PROJECT_PATH -b $BRANCH
obs/build-for-obs.sh:git clone --depth=50 git@github.com:pbek/QOwnNotes.git QOwnNotes -b $BRANCH
build-systems/tuxfamily/build-tuxfamily-src.sh:git clone --depth=50 git@github.com:pbek/QOwnNotes.git QOwnNotes -b $BRANCH
[...]

PR #1176, opened 2019-04-08
Make all git clone of depth=1 in build scripts

dandelionred:
As per https://github.com/pbek/QOwnNotes/issues/1174


pbek:
thank you

Issue #1248, opened 2019-07-26
Narrow down the search results with count


prog20901:
Number of hits found should be part of the search. Ex: 
Branham - 1000 hits
Branham is a propher of God - 500 hits
Branham is a prophet of God for this generation - 230 hits
This will help the user to narrdown the search more..

might be bit sluggish however the narrow down of results are very important when the number of hits found are huge....Any user would always like to search and lookup for few records and not all always.

pbek:
Thank you for your suggestion.

For me it usually is more important to see in how many notes the keyword is found.
Do you have any suggestions in what way (in the user interface) this information should be presented?

prog20901:
Just like Google Search which shows the results and in the brackets give the count found for the given text

pbek:
Google search is a web page 😄  QON is a desktop application. I have no really good place in the user interface to put such information.

prog20901:
Why don't you display in the search bar itself in brackets the count. Example, go to http://table.branham.org/#/main and type "holy ghost" and you will see the list with count. Just an example

pbek:
Thank you for your example but this still is a web page and no Qt application 😁

prog20901:
Example of desktop application - https://raw.githubusercontent.com/xiaoniu-578fa6bff964d005/docfetcher/type-ahead/dev/screenshots/type-ahead-search.gif

prog20901:
Another example of desktop application. However you need to bring the search in a new window - https://www.prabhupada.com/Vedabase/images/VB-programscr4.jpg

pbek:
> Example of desktop application - https://raw.githubusercontent.com/xiaoniu-578fa6bff964d005/docfetcher/type-ahead/dev/screenshots/type-ahead-search.gif

this one might work, but I would have to do some experiment if I can properly add and remove new columns and persuade the user interface to resize the other column to show the amount of hits in a note

the latter example seems to need a different "advanced" dialog

pbek:
...was a bit tricky

## 19.7.9
- if you **search for text in all notes** you will now **see how many times**
  the first word of your **search term occurred** in each note in the note list
  in a new column
    - the new column will be automatically resized to be visible next to the
      note name without scrolling

pbek:
There now is a new release, could you please test it and report if it works for you?

prog20901:
I have tested. one issue is

typed a word and it shows the list of files the occurances found. However i
am unable to find the exact match. it goes with or condition

[...]

PR #1305, opened 2019-10-10
Add a setting to disable the search result count calculation

smokris:
The search result count calculation (#1248) significantly slows down searching on my system (2,200 notes).  This PR adds a setting allowing the user to disable it.  (It remains enabled by default.)


pbek:
Looking good. Thank you!

Issue #1263, opened 2019-08-12
Switch to botan2 from botan


tim77:
#### Expected behaviour
Using modern, safe, well maintained libs.

#### Actual behaviour
Using old, legacy stuff.

pbek:
Thank you for your concern. Any pull request coming up?

tim77:
Unfortunately can promise this for now. We trying to package it according to Fedora guidelines and push in official repos and there some [issues](https://src.fedoraproject.org/rpms/compat-openssl10/blob/master/f/compat-openssl10.spec#_403) with this old `botan` happens in Rawhide.

thmo:
For now, botan has been rebuilt without OpenSSL in Fedora 31 and Rawhide.

Nonetheless, botan v1 has gone EOL 2018-12-31 and is not supported anymore: https://botan.randombit.net/handbook/support.html

pbek:
Back then I took botan the Botan source (one big CPP file with no dependencies) from QtCreator (see https://github.com/pbek/QOwnNotes/tree/develop/src/libraries/botan), so it should compile just fine in the future.

pbek:
Qt Creator seems to have switched to Botan 2 with 4.8 and removed it at all in 4.9.

thmo:
Embedding such an ancient version of a crypto lib doesn't seem to be a good idea to me:
https://botan.randombit.net/security.html

pbek:
There seems to be no "single file" botan.cpp in <https://github.com/qt-creator/qt-creator/tree/4.8/src/libs/3rdparty/botan>

pbek:
> Embedding such an ancient version of a crypto lib doesn't seem to be a good idea to me.

We just need it for AES encryption and decryption.
Suggestions and pull requests are welcome.

The new solution needs to work with qmake and cmake, build on all currently supported platforms and distributions and be still able to decrypt encrypted notes from Botan 1.

tim77:
Also cant [build](https://koji.fedoraproject.org/koji/taskinfo?taskID=37038590) for s390x and  ppc64le arches because of bundled `botan` lib.

thmo:
I would not expect a compatibility issue regarding notes encrypted with Botan v1.

pbek:
> Also cant build for s390x and ppc64le arches because of bundled botan lib.

I think I also saw the same on the OpenBuildService...

Waqar144:
[...]

PR #1320, opened 2019-10-19
Botan 2: Remove future internal header warnings

Waqar144:
Remove the future internal header warnings from botan.h

#1263


pbek:
Thank you!

PR #1346, opened 2019-10-31
Use QStringLiteral instead of normal string literals

Waqar144:
- Using `QStringLiteral` instead of string literals to create `QString` at compile time to reduce runtime overhead of QString creation and copy. Improves performance [WIP]
 - Prevent useless calls to `setNoteTextFromNote()` from `updatePanelMenu()` when note preview widget is not visible

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1215,7 +1215,9 @@ void MainWindow::updatePanelMenu() {
         }
 
     // update the preview in case it was disable previously
-    setNoteTextFromNote(&currentNote, true);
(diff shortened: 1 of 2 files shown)


pbek:
Nice work!

pbek:
Haha, but your PR completely broke my un-commited changes.

Waqar144:
> Haha, but your PR completely broke my un-commited changes.

Hope it wasn't too bad :laughing: 

Note loading performance has increased a lot. The same note which was loading in 1000ms is now loading in 390ms~

pbek:
Wow, that's awesome!

PR #1347, opened 2019-11-01
Fix spellchecker only checking plain text

Waqar144:
Spellchecker was not checking any type of markdown formatted text
#125


pbek:
Looking good. Question is if we want spell check in code blocks...

PR #1356, opened 2019-11-04
Fix an optimization regression from #7848f4d

Waqar144:
Reviewed the whole commit again. Seems like everything was fixed last night except this one line.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2477,7 +2477,7 @@ void MainWindow::updateTreeWidgetItemHeight(QTreeWidget *treeWidget,
 
     // remove the old height stylesheet
     styleText.remove(QRegularExpression(
-            QStringLiteral("QTreeWidget::item \\{height: \\d+px\\}"),
+            QStringLiteral("\nQTreeWidget::item \\{height: \\d+px\\}"),
             QRegularExpression::CaseInsensitiveOption));
 
     // add the new height stylesheet


pbek:
Thank you! Was there no other file beside mainwindow.cpp?

Waqar144:
> Thank you! Was there no other file beside mainwindow.cpp?

there was another file in that PR: `databaseservice.cpp`. I checked it as well. Seems good.

pbek:
Great! And that commit was the only commit with (major) refactoring?

Waqar144:
Yes. That has been the only commit with refactoring from my side till now.
You should review this change though, just to be sure:

https://github.com/pbek/QOwnNotes/blob/7848f4de14fd152be8a8ffc7c8d97f8f37d1a431/src/mainwindow.cpp#L1218-L1220

pbek:
I wasn't sure about that one, but I let it in for now...

Waqar144:
> I wasn't sure about that one, but I let it in for now...

The same thing is being checked inside the `setNoteTextFromNote(&currentNote, true);` function:

https://github.com/pbek/QOwnNotes/blob/be629652e1b831d42b61dc066f3d96b6729c01dc/src/mainwindow.cpp#L3997-L3998

So, theoretically it shouldn't cause any sort of problems.

pbek:
Did you change those two lines?

Waqar144:
> Did you change those two lines?

Nah. They are there.

PR #1360, opened 2019-11-04
Fix memory leak

Waqar144:
Fix memory leak when creating dictionary menu items

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -11384,6 +11384,7 @@ void MainWindow::loadDictionaryNames() {
         }
     }
     connect(languageGroup, SIGNAL(triggered(QAction*)), this, SLOT(onLanguageChanged(QAction*)));
+    delete speller;
 }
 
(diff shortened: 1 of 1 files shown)


Waqar144:
@pbek 
There are two other memory leaks:
- In `ScriptingService`:
https://github.com/pbek/QOwnNotes/blob/111fff2f429aa87896702bfd74029597a1d13e70/src/services/scriptingservice.cpp#L104
- In `NoteHistoryItem`:
https://github.com/pbek/QOwnNotes/blob/111fff2f429aa87896702bfd74029597a1d13e70/src/entities/notehistory.cpp#L149-L152

pbek:
>  There are two other memory leaks

Something you want to tackle?

Waqar144:
> > There are two other memory leaks
> 
> Something you want to tackle?

I am looking into it, first one is a bit more complicated. Second one is easier if we just inherit from Qobject. `delete` is not going to work in both these cases.

Issue #1359, opened 2019-11-04
Spell checking performance issues on macOS


smokris:
Since spell checking (#125) was added, QOwnNotes takes a very long time to display a note after I select it in the note list (tested with QOwnNotes 19.11.3 on macOS 10.14.6).  It grows linearly with the size of the note:

   - 3KB note — 1 second
   - 18KB note — 5 seconds
   - 46KB note — 11 seconds
   - 159KB note — 58 seconds

I ran a time profile while waiting for a note to load, and most of the time is spent doing language detection:
![Screen Shot 2019-11-04 at 09 09 42](https://user-images.githubusercontent.com/190164/68127782-87737480-fee4-11e9-9125-f8f043886eb4.png)

If I go to Edit > Languages and pick (for example) American English, note selection is somewhat faster:

   - 46KB note — 2 seconds
   - 159KB note — 12 seconds

And if I disable spell checking entirely, it's back to normal (before QOwnNotes added spell checking) speed:

   - 46KB note — < 1 second
   - 159KB note — 1 second

Perhaps QOwnNotes could run language detection on only a subset of each note, to speed it up?  And maybe perform spell checking asynchronously and/or only on the visible text in the current viewport?

Waqar144:
Thank you very much for the profiling, it gives a better idea of what's happening on Mac, since I have never been able to build or try it on mac and I have no one to work with me interactively so that we can try to improve the performance on Mac.

The culprit in this case seems to be the native spell checker for Mac. Either it is being used wrong or there is something else. I am not sure.

I have given some possible solutions in #125, one of them being switching to Hunspell for Mac. The decision rests with @pbek

Waqar144:
If anyone has better suggestions or ideas or is willing to help in any way, feel free to write please.

pbek:
> I have given some possible solutions in #125, one of them being switching to Hunspell for Mac. The decision rests with @pbek

I'm ok with using hunspell, if the native spellchecker support doesn't work out... Or I wonder if you want to go multithreaded, @Waqar144

Waqar144:
> > I have given some possible solutions in #125, one of them being switching to Hunspell for Mac. The decision rests with @pbek
> 
> I'm ok with using hunspell, if the native spellchecker support doesn't work out... Or I wonder if you want to go multithreaded, @Waqar144

[...]

PR #1362, opened 2019-11-05
Call updateCurrentNote() only once when setting the note

Waqar144:
Huge performance improvement especially for big notes.
#1359

### DIFF
--- a/src/helpers/qownnotesmarkdownhighlighter.cpp
+++ b/src/helpers/qownnotesmarkdownhighlighter.cpp
@@ -50,8 +50,12 @@ QOwnNotesMarkdownHighlighter::~QOwnNotesMarkdownHighlighter()
     delete wordTokenizer;
 }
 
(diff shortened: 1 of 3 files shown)


pbek:
Great! :tada:

Waqar144:
Also,
I think if we override the setText method in QOwnNotesMarkDownTextEdit, we can check for stuff like CodeBlocks, Comments, Front Matter and Broken Links.. before setting the text. We can then turn off these highlighting options based on the results. After setting the text, we can turn everything on again so that new text gets highlighted properly.

pbek:
Sounds like a good idea, but what happens if a text block changes later (aka. "the user edits the note")?

pbek:
Sounds good. :smile_cat:

Waqar144:
> Sounds good. smile_cat

I will try it, if it works good, I will send in the code for testing

PR #1365, opened 2019-11-05
Improve intial note loading performance

Waqar144:
Huge performance improvements when loading notes for the first time.
#1359


src/widgets/qownnotesmarkdowntextedit.cpp
```diff
@@ -308,6 +308,59 @@ QMargins QOwnNotesMarkdownTextEdit::viewportMargins() {
...
+    }
+
+    //check for code blocks
+    if (!text.contains(QStringLiteral("```"))) {
+        h->setCodeHighlighting(false);
+    }
+
+    //check for front matter
+    if (h->document()->firstBlock().text() != "---") {
+
```
pbek: Did you miss a `setFrontmatterHighlighting`?

Waqar144: ahh yes. Fixed

Issue #1378, opened 2019-11-08
Spellchecker dictionary download manager


pbek:
There will be a dialog for downloading dictionaries.
The dictionaries will be stored in the application's data path so they also will be available in portable mode.

pbek:
@Waqar144, does our spellchecker currently store some data locally? If yes, we need to make it so that it does it in the data path, so the rest of the system doesn't get "polluted" in portable mode.

Waqar144:
> @Waqar144, does our spellchecker currently store some data locally? If yes, we need to make it so that it does it in the data path, so the rest of the system doesn't get "polluted" in portable mode.

I guess  *personal dictionary* is stored locally at ~/.hunspell_langCode
it could be stored anywhere else

pbek:
Good question, are those shared with other applications like LibreOffice?

I guess at least in portable mode (`Utils::Misc::isInPortableMode()`) they must reside in a directory in `Utils::Misc::appDataPath()`.

Waqar144:
> Good question, are those shared with other applications like LibreOffice?

- LibreOffice stores its *personal dictionary* data in ~/.config/libreoffice on linux. On windows/Mac it will be something similar.
- Chrome/ Firefox also store their data in separate places. I just tried and they are not detecting the words I added to my personal hunspell dict created using QOwnNotes/Kde apps
- In KDE land, since everything is using sonnet(i think), the personal dictionary is shared between all kde apps, as well as QOwnNotes.
- In Gnome land, I can't be sure, since I haven't tested it. I think it doesn't using hunspell, maybe enchant or something else. So I am 90% sure, QOwnNotes's personal dicts aren't shared there, unless of course you install some kde app.

> I guess at least in portable mode (`Utils::Misc::isInPortableMode()`) they must reside in a directory in `Utils::Misc::appDataPath()`.

Sounds good. Because what's the use of *personal dict* if you are not going to carry it with you. So, yeah I agree, in portable mode the path for personal dict should be set using `portableDataPath()`

pbek:
> Sounds good. Because what's the use of personal dict if you are not going to carry it with you. So, yeah I agree, in portable mode the path for personal dict should be set using portableDataPath()

Could you do that please? :smile_cat:

Waqar144:
> > Sounds good. Because what's the use of personal dict if you are not going to carry it with you. So, yeah I agree, in portable mode the path for personal dict should be set using portableDataPath()
> 
[...]

PR #1380, opened 2019-11-08
Allow portable mode to store dic in portable data

Waqar144:
Personal dictionary will now be stored in `portableDataPath()` when using QOwnNotes in portable mode

#1378


pbek:
Thank you!

Issue #1379, opened 2019-11-08
Spell checking bug: italics causes red wavy lines


sbrl:
#### Expected behaviour
When using italics on more than one word, QOwnNotes should not identify it as a spelling mistake. Example:

```markdown
Some text _some text_
```

In the above example, `text_` will be marked as a spelling mistake. Screenshot:

![image](https://user-images.githubusercontent.com/9929737/68472447-7db97c00-0218-11ea-95f2-7898ebe5d7ed.png)

#### Actual behaviour
It identifies the last word of multi-word italics as a spelling mistake

#### Steps to reproduce
1. Make sure spell checking is turned on (I have it set to en_GB)
2. Type some text
3. Put italics on multiple words, as above

#### Output from the debug section in the settings dialog

<details>
<summary>Debug information</summary>
QOwnNotes Debug Information
===========================

## General Info

**Current Date**: `Fri Nov 8 11:13:10 2019`
**Version**: `19.11.5`
**Build date**: `Nov  5 2019`
**Build number**: `610`
**Platform**: `linux`
**Operating System**: `Ubuntu 18.04.3 LTS`
**Build architecture**: `x86_64`
**Current architecture**: `x86_64`
**Release**: `Launchpad PPA`
**Qt Version (build)**: `5.9.5`
**Qt Version (runtime)**: `5.9.5`
**Portable mode**: `no`
**Settings path / key**: `/home/bryan-smithl/.config/PBE/QOwnNotes.conf`
**Application database path**: `/home/bryan-smithl/.local/share/PBE/QOwnNotes/QOwnNotes.sqlite`
**Application arguments**: `QOwnNotes`
**Qt Debug**: `0`
**Locale (system)**: `en_GB`
**Locale (interface)**: *empty*
**Icon theme**: `breeze-dark-qownnotes`
**Notes in current note folder**: `62`
**Calendar items**: `24`
**Enabled scripts**: `3`

## Server Info

**serverUrl**: `https://nextcloud.starbeamrainbowlabs.com`
**appIsValid**: `no`
**notesPathExists**: *empty*
**connectionErrorMessage**: *empty*

## Note folders

**currentNoteFolderId**: `1`

### Note folder `default`

**id**: `1`
**isCurrent**: `yes`
**activeTagId**: `-1`
**localPath**: `/home/bryan-smithl/Documents/Nextcloud/Notes`
**remotePath**: `Notes`
**cloudConnectionId**: `1`
[...]

PR #1381, opened 2019-11-08
Fix underscore at end of word not being tokenized

Waqar144:
Underscore at the end of a word wasn't being tokenized thus causing it to be misspelled.

#1379


pbek:
OBS seems to be breaking with the changes...

pbek:
e.g. https://build.opensuse.org/package/live_build_log/home:pbek:QOwnNotes/desktop/openSUSE_13.2/i586

PR #1393, opened 2019-11-09
Refactoring: Use more constness and const refs

Waqar144:


src/helpers/qownnotesmarkdownhighlighter.cpp
```diff
@@ -55,12 +55,9 @@ QOwnNotesMarkdownHighlighter::~QOwnNotesMarkdownHighlighter()
     delete wordTokenizer;
 }
 
-void QOwnNotesMarkdownHighlighter::updateCurrentNote(Note *_note) {
-    if (_note == nullptr) {
-        _currentNote = Note::fetch(qApp->property("currentNoteId").toInt());
```
pbek: Why did you remove these lines? They are pretty important. :grin:

Waqar144: This function is only being called once for every note from the `setNoteTextFromNote` function. The note can never be null, and using a ptr here seemed useless.

The null check is now being performed in the `setNoteTextFromNote` for safety.

As for the second line, it's pretty much useless if we pass the note directly to `_currentnote`, no?

src/mainwindow.cpp
```diff
@@ -5692,7 +5696,7 @@ void MainWindow::openLocalUrl(QString urlString) {
             Note note = Note::fetch(noteId);
             if (note.isFetched()) {
                 // set current note
-                setCurrentNote(note);
+                setCurrentNote(std::move(note));
```
Waqar144: This, I think you should review.
after an `std::move(note)`, the `note` object will become empty, so it can't be used again in the function. It seems to me that after the first this `if`, it's not likely that we can fall into any other condition branch, so we are safe. but better be sure.

I can remove these in any case.

src/helpers/qownnotesmarkdownhighlighter.cpp
pbek: Phew, I just saw you disabled the call from `QOwnNotesMarkdownHighlighter::highlightBlock` in `Call updateCurrentNote() only once when setting the note (#1362)`. That maybe will break note linking and maybe other things in certain circumstances...

pbek: I guess we'll wait and see.

src/mainwindow.cpp
pbek: Looks like it's safe. :smiling_imp:

src/helpers/qownnotesmarkdownhighlighter.cpp
Waqar144: > Phew, I just saw you disabled the call from `QOwnNotesMarkdownHighlighter::highlightBlock` in `Call updateCurrentNote() only once when setting the note (#1362)`. That maybe will break note linking and maybe other things in certain circumstances...

Yes that was done a while ago.

It doesn't need to be called for every block since the note is always same. But when the note changes we update the currentNote().
Calling it in highlightBlock is very expensive.

PR #1397, opened 2019-11-10
Sonnet: Load dictionaries concurrently to improve performance

Waqar144:
Loading dictionaries asynchronously to improve performance.
Using this, we cut almost 200-250ms that were wasted in loading dictionaries.
Autodetect should be much faster now.

#1359

### DIFF
--- a/src/libraries/sonnet/src/plugins/hunspell/hunspelldict.cpp
+++ b/src/libraries/sonnet/src/plugins/hunspell/hunspelldict.cpp
@@ -31,6 +31,9 @@
 #include <QTextStream>
 #include <QStandardPaths>
 #include <utils/misc.h>
+#include <QStringBuilder>
+#include <QtConcurrent/QtConcurrent>
+#include <thread>
 
 using namespace Sonnet;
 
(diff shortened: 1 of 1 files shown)


Waqar144:
Don't merge this yet.

pbek:
Nicely done!

pbek:
Lol, didn't see the comment before merging. :laughing:

Waqar144:
> Lol, didn't see the comment before merging. 😆

Well did you try it out yet? How is it working for you?

pbek:
Seems to work fine on Linux and macOS. Anything I should keep an eye on?

Waqar144:
Try with multiple languages, see if it detects them correctly the first time you load the note?
If it doesn't, switch notes to see if that makes a difference 🤔

pbek:
Should the Automatic mode work for each note individually? That almost never works for me. I'm currently testing with Text from Wikipedia (English and German). I've notes with just English text and notes with just German text. The language is not detected by the spellchecker.

pbek:
But it also doesn't work without this commit (I reverted it locally).
I also get the warning `Missing trigrams for languages: QSet("de_AT")`. Austrian German Dictionary is installed (twice actually)...

pbek:
After I installed `ger_DE` (Germany German) the auto-detection worked for each note (also with this PR). But I German notes usually need a 2nd activation for it to be detected correctly.

```markdown
Kulturapfel
=================

Der Kulturapfel (Malus domestica Borkh., Syn. Pyrus malus L.) ist eine weithin bekannte Art aus der Gattung der Äpfel in der Familie der Rosengewächse (Rosaceae). Er ist eine wirtschaftlich sehr bedeutende Obstart. Die Frucht des Apfelbaumes wird Apfel genannt.

Äpfel werden sowohl als Nahrungsmittel im Obstbau als auch zur Zierde angepflanzt. Außerdem wird ihnen eine Wirkung als Heilmittel zugeschrieben. Als die Frucht schlechthin symbolisieren Apfel und Apfelbaum die Themenbereiche Sexualität, Fruchtbarkeit und Leben, Erkenntnis und Entscheidung sowie Reichtum. 
```

pbek:
On the official 19.11.9 release the automatic detection works on the first time with the same notes if I install the German dictionary for Germany, For the Austrian I also get `Missing trigrams for languages: QSet("de_AT")`.

Waqar144:
> But it also doesn't work without this commit (I reverted it locally).
> I also get the warning `Missing trigrams for languages: QSet("de_AT")`. Austrian German Dictionary is installed (twice actually)...



> But it also doesn't work without this commit (I reverted it locally).
> I also get the warning `Missing trigrams for languages: QSet("de_AT")`. Austrian German Dictionary is installed (twice actually)...

Yes because in the trigrams there's only one type of every language. See:
https://github.com/KDE/sonnet/tree/master/data/trigrams

I wonder what would happen if we added de_AT to that. That would surely solve this problem.



> After I installed `ger_DE` (Germany German) the auto-detection worked for each note (also with this PR). But I German notes usually need a 2nd activation for it to be detected correctly.
> 
> Example note:
> 
> ```gfm
> Kulturapfel
> =================
> 
> Der Kulturapfel (Malus domestica Borkh., Syn. Pyrus malus L.) ist eine weithin bekannte Art aus der Gattung der Äpfel in der Familie der Rosengewächse (Rosaceae). Er ist eine wirtschaftlich sehr bedeutende Obstart. Die Frucht des Apfelbaumes wird Apfel genannt.
> 
> Äpfel werden sowohl als Nahrungsmittel im Obstbau als auch zur Zierde angepflanzt. Außerdem wird ihnen eine Wirkung als Heilmittel zugeschrieben. Als die Frucht schlechthin symbolisieren Apfel und Apfelbaum die Themenbereiche Sexualität, Fruchtbarkeit und Leben, Erkenntnis und Entscheidung sowie Reichtum. 
> ```
> 
> If this note is the current note for the first time the detection fails. On the 2nd time it works.

On the second time, you mean after switching notes?


This PR is basically not a very good effort to load dictionaries concurrently. The auto detection basically happens like this:
- Read a sentence
- Check trigrams for a match? Found a match? Yes.
  - Go tell the spellchecker and load the dictionary for it.
- Found a match? No.
  - Go load the dictionaries and brute force a match.

What happens in this PR:
- We found a match in trigrams so lets check with the dictionary if the word is correct. Send request to Hunspell, Ooops dictionary wasn't loaded yet. return false -> hence every word gets underlined red.
Then when you switch the notes and come back, the dictionaries are loaded and everything works as expected.

What this PR gives:
- better Initial note loading performance (since the time required to load a dictionary or a second dictionary is skipped)
- poor spellchecking the first time you open QON

pbek:
Without this PR the automatic detection also works on the first time.

pbek:
> On the second time, you mean after switching notes?

correct, switching to other note and switching back

Waqar144:
> > On the second time, you mean after switching notes?
> 
> correct, switching to other note and switching back

Exactly. So I guess this needs to be reversed.

Waqar144:
Basically hunspell is slow[1] and auto detection makes it even more slower. The bigger the dictionary, the longer it's gonna take to start. The more dictionaries you have on your drive, the more its gonna impact initial performance.

And then there's this: https://github.com/wolfgarbe/SymSpell


[1]http://wiki.languagetool.org/hunspell-support

Waqar144:
Also, worth sharing:
https://wiki.qt.io/Using_QString_Effectively
Now i know where the % was coming from
:rofl:

pbek:
> https://github.com/wolfgarbe/SymSpell

Wow, that fast. :grin: 

> https://wiki.qt.io/Using_QString_Effectively

Do we need `QT_USE_QSTRINGBUILDER` now? :laughing:

Waqar144:
> > https://github.com/wolfgarbe/SymSpell
> 
> Wow, that fast. grin
> 
> > https://wiki.qt.io/Using_QString_Effectively
> 
> Do we need `QT_USE_QSTRINGBUILDER` now? laughing

just `#include<QStringBuilder>` :wink:

Waqar144:
> > https://github.com/wolfgarbe/SymSpell
> 
> Wow, that fast. grin

I am very interested in this.
Pros:
- Probably a lot faster if it works
- Much much smaller than hunspell

And it seems there are a lot of dictionaries available for it too

pbek:
You really want to re-implement everything? :laughing:

Waqar144:
> You really want to re-implement everything? laughing

First I will do a little bit, see if it builds and works as claimed. Then we can think about re-implementing. But yeah it's gonna be tedious integrating it into Sonnet.

pbek:
And new dictionaries will also be needed...

Waqar144:
> And new dictionaries will also be needed...

There are a lot of dictionaries available for it,
plus it provides a dictionary creation option from a large text corpus.

Well anyways, its not happening anytime soon :laughing:

pbek:
:grin:

Waqar144:
Dictionaries are plenty.
Provided it works, it can be a better choice but the fact that no one's using it makes me a bit skeptical.

https://github.com/hermitdave/FrequencyWords

Waqar144:
> I wonder what would happen if we added de_AT to that. That would surely solve this problem.

You can however make a trigram for it using de_DE file, if you want de_AT autodetection :wink:

pbek:
Do we generate the trigrams at runtime? I can't find the files in our repository.

Waqar144:
> Do we generate the trigrams at runtime? I can't find the files in our repository.

We have it pre-compiled.

pbek:
Ah I see, so currently it's "insider information" that only certain languages will work with auto-detection. :grinning:

Waqar144:
> Ah I see, so currently it's "insider information" that only certain languages will work with auto-detection. grinning

Hahah kind of :laughing:

pbek:
I wonder if I should remove those languages from the dictionary manager...

Waqar144:
> I wonder if I should remove those languages from the dictionary manager...

Which ones?
All languages will work in specific mode in any case

Waqar144:
The trigrams have most of the languages covered, even some languages which don't have a dictionary.
It just doesn't contain variation for languages such as only en_US and de_DE.

pbek:
>  such as only en_US and de_DE.

I guess quite some people are using en_US :grin:

Waqar144:
> > such as only en_US and de_DE.
> 
> I guess quite some people are using en_US grin

Haha or in case they don't have en_US dic installed, auto detection will probably fail for them.
Should I recompile with en_GB and de_AT and so on?
The trigrams for en_GB will be same as en_US since the language is practically the same. I don't know about de

pbek:
For the auto-detection they are quite the same. AT and DE only differ in a few words...

pbek:
>  Should I recompile with en_GB and de_AT and so on?

that would be great! :tada:

Waqar144:
> > Should I recompile with en_GB and de_AT and so on?
> 
> that would be great! tada

ok let's do this.

PR #1401, opened 2019-11-11
Fix etc being marked misspelled for en_GB

Waqar144:
#1399

### DIFF
--- a/src/helpers/qownspellchecker.cpp
+++ b/src/helpers/qownspellchecker.cpp
@@ -54,9 +54,13 @@ void QOwnSpellChecker::setCurrentLanguage(const QString &lang) {
     spellchecker->setLanguage(lang);
 }
 
-bool QOwnSpellChecker::isWordMisspelled(const QString &word) {
-    return spellchecker->isMisspelled(word);
-}
+//bool QOwnSpellChecker::isWordMisspelled(const QString &word) {
+
+//    if (word == QLatin1String("etc") && language == QLatin1String("en_GB")) {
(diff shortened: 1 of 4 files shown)


pbek:
I don't like that fix at all... Hardcoded string comparisons that will hit every word for every language... :crying_cat_face:

Waqar144:
> I don't like that "fix" at all... Hardcoded string comparisons that will hit every word for every language... crying_cat_face

There are two other solutions:
- User adds it to their personal dictionary. It will get ignored
- User repairs the en_GB dic to not mark etc. as misspelled
- en_GB fixes itself

On our side that's all we can do.

pbek:
Other languages and other abbreviations are affected too, see https://github.com/pbek/QOwnNotes/issues/1399#issuecomment-552582204

Waqar144:
> Other languages and other abbreviations are affected too, see [#1399 (comment)](https://github.com/pbek/QOwnNotes/issues/1399#issuecomment-552582204)

I did.
Closing the PR. Fixable on the user side.

pbek:
We cannot "fix" every abbreviation anyway...

Waqar144:
> We cannot "fix" every abbreviation anyway...

Yes, and there's going to be a lot of users with such problems. We can't handle dictionary problems. Unless we were hosting the dictionaries ourselves

Waqar144:
How about we host dictionaries here in a repo?

pbek:
https://github.com/qownnotes/dictionaries

PR #1402, opened 2019-11-11
Update trigrams for en_GB and de_AT

Waqar144:
Dedicated to @pbek 

Be sure to delete de_DE, otherwise be ready for the performance overhead ;)
I am not sure yet why that happens but it happens.


pbek:
Thank you!

PR #1405, opened 2019-11-12
Update dictionary manager to target qownnotes/dictionaries

Waqar144:
#1399


pbek:
You can also update the documentation if you want.

Waqar144:
> You can also update the documentation if you want.

Ah yes,
Will do that in a while

PR #1411, opened 2019-11-13
tagapi.cpp: minor refactor

Waqar144:


src/api/tagapi.cpp
```diff
@@ -1,8 +1,8 @@
 #include <utility>
 #include "tagapi.h"
 
-TagApi* TagApi::fetch(int id) {
-    Tag tag = Tag::fetch(id);
+TagApi* TagApi::fetch(int _id) {
```
pbek: I don't like the `_` a lot. `_` kind of was reserved for class variables.

Waqar144: Ah okay. I did this because of the warning `Member shadowing parent class variable`

pbek: You rather could rename all class member variables to underline-something, like `_id`. How do other projects solve this?

Waqar144: Underscore is the usual quick way of getting around this.
It would work without it in any case. This is just for clarity.

pbek: Currently underscore is used for new variables in classes (I haven't refactored all old variables yet). Now using them also in parameters would be the first time and also confusing, don't you think? :grin:

Waqar144: Yes of course. Consistency is important. Better not merge this then

pbek: ...or give the variables other name or refactor the names of the internal class variables that are shadowed by the parameters. :grin:

Waqar144: but tagapi doesn't have variables :thinking: 
But yeah we can give the parameters different names, cause they are the ones giving shadow warnings.
Something like `tagId`, `tagName`, `tagParentId`?

pbek:
> but tagapi doesn't have variables

they are at the top in the header file and I don't think we want to change them in this case

> Something like tagId, tagName, tagParentId?

for example, if the warning is bothering us :)

Issue #1423, opened 2019-11-17
Increasing font size does not increase plain text on Windows 10 if interface font size is overridden


raelb:
As you can see in attached video, increasing the font size does not increase plain (unformatted) text. However it is increasing Header size, Bold and Italics text.

I would expect this to apply to all elements equally.

[increase font size.zip](https://github.com/pbek/QOwnNotes/files/3855553/increase.font.size.zip)

QOwnNotes Debug Information
===========================

## General Info

**Current Date**: `Sun Nov 17 12:42:57 2019`
**Version**: `19.11.13`
**Build date**: `Nov 14 2019`
**Build number**: `4886`
**Platform**: `windows`
**Operating System**: `Windows 10 (10.0)`
**Build architecture**: `i386`
**Current architecture**: `x86_64`
**Release**: `AppVeyor`
**Qt Version (build)**: `5.13.2`
**Qt Version (runtime)**: `5.13.2`
**Portable mode**: `yes`
**Settings path / key**: `C:/Downloads/Editors/QOwnNotes/Data/PBE/QOwnNotes.ini`
**Application database path**: `C:\Downloads\Editors\QOwnNotes\Data\QOwnNotes.sqlite`
**Application dictionaries path**: `C:\Downloads\Editors\QOwnNotes\Data\dicts`
**Application arguments**: `QOwnNotes.exe`, `--portable`
**Qt Debug**: `0`
**Locale (system)**: `en_ZA`
**Locale (interface)**: *empty*
**Icon theme**: `breeze-qownnotes`
**Notes in current note folder**: `6`
**Calendar items**: `0`
**Enabled scripts**: `0`

## Server Info

**serverUrl**: *empty*
**appIsValid**: `no`
**notesPathExists**: *empty*
**connectionErrorMessage**: *empty*

## Note folders

**currentNoteFolderId**: `1`

### Note folder `default`

**id**: `1`
**isCurrent**: `yes`
**activeTagId**: `0`
**localPath**: `C:\Downloads\Editors\QOwnNotes\Data\Notes`
**remotePath**: `Notes`
**cloudConnectionId**: `1`
**isShowSubfolders**: `yes`
**isUseGit**: `no`
**activeNoteSubFolder name**: `Folder 2`
**database file**: `C:\Downloads\Editors\QOwnNotes\Data\Notes\notes.sqlite`

### Cloud connection `Default`

**id**: `1`
**isCurrent**: `yes`
**serverUrl**: *empty*
[...]

PR #1425, opened 2019-11-18
Disable spellchecking in logTextEdit

Waqar144:
#1423

### DIFF
--- a/src/dialogs/issueassistantdialog.cpp
+++ b/src/dialogs/issueassistantdialog.cpp
@@ -21,6 +21,12 @@ IssueAssistantDialog::IssueAssistantDialog(QWidget *parent) :
     on_issueTypeComboBox_currentIndexChanged(ProblemIssueType);
     ui->logOutputPlainTextEdit->setHighlightingEnabled(false);
 
+    //enable spellchecking
+    ui->expectedBehaviourPlainTextEdit->enableSpellChecker();
+    ui->actualBehaviourPlainTextEdit->enableSpellChecker();
+    ui->questionPlainTextEdit->enableSpellChecker();
+    ui->stepsPlainTextEdit->enableSpellChecker();
+
     QObject::connect(ui->titleLineEdit, SIGNAL(textChanged(QString)),
                      this, SLOT(allowIssuePageNextButton()));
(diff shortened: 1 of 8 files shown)


pbek:
I wonder if wouldn't be better to set a flag or something to disable the spellchecker and not checking the the object name. 😁 
I guess we should also disable spellchecking in the *Debug settings*.

Waqar144:
> I wonder if wouldn't be better to set a flag or something to disable the spellchecker and not checking the the object name....

There seems to be no simple way to do that. 
I will find a way around it.

pbek:
> There seems to be no simple way to do that.

I thought as much. :grin: 

>  I will find a way around it.

:tada:

pbek:
If it's too bad let's stay at checking the object names... (and add a check for the debug information text edit, maybe via a QStringList with all the object names)

Waqar144:
> If it's too bad let's stay at checking the object names... (and add a check for the debug information text edit, maybe via a QStringList with all the object names)

For now,
`Spellchecker` doesn't get initialized except when we set the note text using `setText()`. Note text is set only for `noteTextEdit` so spell checking stays disabled for everything else. What do you think?

pbek:
Sounds great. Thank you!

pbek:
Ah, you also need to add `encryptedNoteTextEdit`. :laughing:

Waqar144:
> Ah, you also need to add `encryptedNoteTextEdit`. laughing

It works for that too. I mean spellchecking is enabled inside it. :wink: Though there's not any point in spellchecking the encrypted text :laughing:

pbek:
> It works for that too. I mean spellchecking is enabled inside it. 

are you sure?

> Though there's not any point in spellchecking the encrypted text laughing

no, it's the other way around. `encryptedNoteTextEdit` is used to actually edit the encrypted notes. The encrypted note text is then written to `noteTextEdit` to be stored in the note file.

Waqar144:
> > It works for that too. I mean spellchecking is enabled inside it.
> 
> are you sure?

Yeah it's working,
I just encrypted a note and then decrypted it. Spell checking is working. 

However, 
Post issue, feature request, menu action is not working. Crashes the app

Waqar144:
> Issue Assistant is not working. Crashes the app

Fixed

PR #1440, opened 2019-11-22
Major performance improvements

Waqar144:
Note loading speed has now doubled

### DIFF
--- a/src/helpers/qownnotesmarkdownhighlighter.cpp
+++ b/src/helpers/qownnotesmarkdownhighlighter.cpp
@@ -153,7 +153,7 @@ void QOwnNotesMarkdownHighlighter::highlightMarkdown(const QString& text) {
             highlightCodeBlock(text);
         }
     }
-    if (frontMatterHighlightingOn) {
+    if (currentBlock().position() == 0) {
         highlightFrontmatterBlock(text);
     }
 }
--- a/src/helpers/qownnotesmarkdownhighlighter.cpp
+++ b/src/helpers/qownnotesmarkdownhighlighter.cpp
@@ -130,7 +130,11 @@ void QOwnNotesMarkdownHighlighter::highlightMarkdown(const QString& text) {
 
         // highlight broken note links
         if (highlightBrokenNotesLinkOn) {
+            if (text.contains("note://") || text.contains(".md") || text.contains("[") ||
+                text.contains("]") || text.contains("<") || text.contains(">") ||
+                text.contains("<")) {
             highlightBrokenNotesLink(text);
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


pbek:
This breaks frontmatters, like a note with:

```markdown
---
name: test
---

Frontmatter test 01
========================

asdasd
asdasd
```

src/helpers/qownnotesmarkdownhighlighter.cpp
```diff
@@ -127,23 +118,36 @@ void QOwnNotesMarkdownHighlighter::highlightMarkdown(const QString& text) {
         highlightAdditionalRules(_highlightingRulesPre, text);
 
         // needs to be called after the horizontal ruler highlighting
-        highlightHeadline(text);
+        const QString &next = currentBlock().next().text();
+        if (text.contains(QStringLiteral("# ")) || text.contains(QStringLiteral("## ")) ||
```
pbek: I guess you `startsWith` would be more correct instead of all `contains`.
And instead of all the `# ` checks a regular expression that checks one to six occurrences would also do, wouldn't it?

Waqar144:
> This breaks frontmatters, like a note with:
> 
> ```gfm
> ---
> name: test
> ---
> 
> Frontmatter test 01
> ========================
> 
> asdasd
> asdasd
> ```

Yep. I will disable this one. It's not as beneficial for the performance anyways.

src/helpers/qownnotesmarkdownhighlighter.cpp
Waqar144: I was having problems with startsWith initially. Just tired it again, and it seems to be working fine. No differences in performance.

Using regex is a lot more expensive. It would be more concise and clean I agree, but the performance penalty is going to be much bigger.

As of now, a 16K line note (which is almost equal to a 400 page book) is loading in almost 0.4s with spellchecking enabled(without autodetect).  With autodetect it's 1 sec.

This is a lot faster than many other applications, and most if not all of the Qt based markdown readers and certainly faster than all the electron based apps. (Have tested ghostwriter, focus writer, marktext, Libreoffice, typora). Not to mention that we are doing full markdown highlighting, and not skipping anything.

The slowest function of them all is `highlightBrokenNotesLink()` followed by `highlightHeadline()` and then `highlight-additionalRulesAfter`.

We can speed this up even more, but for that we will have to seperate functionality into smaller functions. For e.g., taking some options out of `highlightAdditionalRulesAfter` and putting them in seperate functions and then placing checks on them to decide whether it needs to be run or not. Separating functionality makes it easier to place checks for individual elements (tables, lists etc.) Same could be done with `headings` and `brokenNotelinks`.

pbek: You are doing an awesome job here! I'm always amazed that improvements are made!

Waqar144: ;)

pbek: I meant "what improvements are made" :smile:

Waqar144: Both sentences are correct I think :rofl:

PR #1441, opened 2019-11-23
Use startsWith() instead of contains

Waqar144:
- Change the order of `if` statement check for more performance
- Use startsWith() instead of contains
- Disable checks for `frontMatter`


pbek:
Works great! You did an awesome job at #1440 and #1441! :+1: :tada:

PR #1443, opened 2019-11-23
Fix autodetection for some languages

Waqar144:
Fixes autodetection for some languages that do not have trigrams for e.g. Malayan

KDE Bug: https://bugs.kde.org/show_bug.cgi?id=176537
Changes sent upstream at: https://phabricator.kde.org/D25495


pbek:
Very well!

Issue #1432, opened 2019-11-20
Add 'navigate to sub-dir' to note-contextmenu


sanderboom:
#### Expected behaviour

When searching and opening a search-result-note it would be cool to have a right-mouse-click option on the search result to be able to move to the subdirectory where the note lives. A bit like sublime's 'reveal in sidebar'.

pbek:
Thank you for your suggestion.

pbek:
## 19.11.21
- added a note list context menu item to jump to the note sub-folder of the
  selected note (for [#1432](https://github.com/pbek/QOwnNotes/issues/1432), thank you @Waqar144)

sanderboom:
Nice!

pbek:
There now is a new release, could you please test it and report if it works for you?

iolivier87:
The subfolder jump works for me.

But after the jump, i can't select the root folder. I have to select another subfolder before to go to the root.

pbek:
I'll look at that (if @Waqar144 isn't before me)

Waqar144:
> I'll look at that (if @Waqar144 isn't before me)

Was very hard to reproduce this issue as it wasn't happening to me under my default setup. I will send in the PR in a moment for review.

pbek:
Will be fixed in the next release.

pbek:
Also affected the scripting function to jump to a sub-folder. Happens only if you want to switch back to the last subfolder by clicking on it.

pbek:
## 19.11.22
- fixed unable to switch back to last note subfolder after jumping to a subfolder
  with the `Jump to the note's sub-folder` note list context menu item
  (for [#1432](https://github.com/pbek/QOwnNotes/issues/1432))

pbek:
There now is a new release, could you please test it and report if it works for you?

iolivier87:
Yes, It works now ! 👍

pbek:
Great, thank you for testing!

PR #1447, opened 2019-11-24
Feature req #1432: Navigate to sub dir from context menu

Waqar144:


src/mainwindow.cpp
```diff
@@ -9400,6 +9400,12 @@ void MainWindow::openNotesContextMenu(
     }
 
     QAction *removeAction = noteMenu.addAction(tr("&Remove notes"));
+
+    QAction *moveToThisSubFolderAction = nullptr;
+    if (ui->noteTreeWidget->selectedItems().count() == 1) {
+        moveToThisSubFolderAction = noteMenu.addAction(tr("Go to this note's folder"));
```
pbek: Can you please call it like "Jump to the note's sub-folder". And you need to check if sub-folders are activated for the current note folder. (there is a method hanging around somewhere that does that)

Waqar144: > Can you please call it like "Jump to the note's sub-folder".

Sure. I was unsure about the label

And you need to check if sub-folders are activated for the current note folder. (there is a method hanging around somewhere that does that)

Will do!

pbek: Awesome!

PR #1454, opened 2019-11-25
Fix dictionaries being displayed incorrectly

Waqar144:


pbek:
I had to revert your last PR because I was preparing a release. It seems this PR doesn't contain your last PR.

Waqar144:
> I had to revert your last PR because I was preparing a release. It seems this PR doesn't contain your last PR.

Yes, I will send in a fresh one.

PR #1458, opened 2019-11-26
Fix unable to switch to root folder directly

Waqar144:


pbek:
Every time I click on a note sub-folder it gets loaded twice now.

pbek:
I fixed that now. I just needed to remove this PR and remove `NoteSubFolder::setAsActive(subFolderId);` from your last PR. Why did you need `NoteSubFolder::setAsActive(subFolderId);`?

Waqar144:
> I fixed that now. I just needed to remove this PR and remove `NoteSubFolder::setAsActive(subFolderId);` from your last PR.

ok

Why did you need `NoteSubFolder::setAsActive(subFolderId);`?

I thought setting as 'active' and then jumping to that folder was the way. I just didn't test without it :(

pbek:
`jumpToNoteSubFolder` should do fine on it's own, it's also used in the scriptping engine ;)

pbek:
> I thought setting as 'active' and then jumping to that folder was the way. I just didn't test without it :(

note sub-folder are pretty complex already, because all the things you can do in the *Panel settings* (e.g. like also show all notes of the sub-folders recursively)

Waqar144:
> `jumpToNoteSubFolder` should do fine on it's own, it's also used in the scriptping engine ;)

Yes,

I just tried this. It works, but the root folder switching isn't working

pbek:
> I just tried this. It works, but the root folder switching isn't working

Oh, then it's not fixed yet, because I can't reproduce it. :( I just added your `setAsActive` again and I'm still able to jump to the root folder, what settings did you use to reproduce it?

Waqar144:
> > I just tried this. It works, but the root folder switching isn't working
> 
> Oh, then it's not fixed yet, because I can't reproduce it. :( I just added your `setAsActive` again and I'm still able to jump to the root folder, what settings did you use to reproduce it?

It works all the time. Except when you have these 'panel settings' checked:
- Show all notes in a folder including their subfolders
- Display note subfolders as full tree

pbek:
Strange. I have (and had) those two activated.

Waqar144:
This is how my subFolderTreeWidget looks like to reproduce this issue:

image

Waqar144:
And my panel settings:

image

pbek:
Settings are the same, folders almost (I have a deeper tree). And you can't click on "Notes" after you jumped to a sub-folder?

Waqar144:
> Settings are the same, folders almost (I have a deeper tree). And you can't click on "Notes" after you jumped to a sub-folder?

Yeah exactly. Clicking on it doesn't reload the `noteTreeWidget`

pbek:
> Clicking on it doesn't reload the noteTreeWidget

But you can click on "Notes" and it is then selected?

pbek:
For me the noteTreeWidget gets the correct notes.

Waqar144:
> > Clicking on it doesn't reload the noteTreeWidget
> 
> But you can click on "Notes" and it is then selected?

Yeah it does get selected. And after you click 5 or 6 times, the notes get loaded

Waqar144:
Here's a small gif

Peek 2019-11-26 12-48

pbek:
> Yeah it does get selected. And after you click 5 or 6 times, the notes get loaded

I see, you need to previously be on the "Notes" folder, not on any other folder. I can reproduce it now.

Waqar144:
> I see, you need to previously be on the "Notes" folder, not on any other folder. I can reproduce it now.

Great. Welcome to the party :laughing:

pbek:
Tricky, this needs to be fixed in `MainWindow::jumpToNoteSubFolder`.

pbek:
That did the trick for me:

```cpp
bool MainWindow::jumpToNoteSubFolder(int noteSubFolderId) {
    QTreeWidgetItem* item =
            Utils::Gui::getTreeWidgetItemWithUserData(
                ui->noteSubFolderTreeWidget, noteSubFolderId);

    if (item != Q_NULLPTR) {
        ui->noteSubFolderTreeWidget->setCurrentItem(item);

        return true;
    }

    return false;
}
```

Waqar144:
> That did the trick for me:
> 
> ```c++
> bool MainWindow::jumpToNoteSubFolder(int noteSubFolderId) {
>     QTreeWidgetItem* item =
>             Utils::Gui::getTreeWidgetItemWithUserData(
>                 ui->noteSubFolderTreeWidget, noteSubFolderId);
> 
>     if (item != Q_NULLPTR) {
>         ui->noteSubFolderTreeWidget->setCurrentItem(item);
> 
>         return true;
>     }
> 
>     return false;
> }
> ```

That makes the function a lot cleaner.

We can also just directly set the Item in the customContextMenu function. But that would lead to duplicate code.

pbek:
> We can also just directly set the Item in the customContextMenu function.

we don't have the item there, don't we?

Waqar144:
> > We can also just directly set the Item in the customContextMenu function.
> 
> we don't have the item there, don't we?

We can create it though

pbek:
why? `MainWindow::jumpToNoteSubFolder` already does that and is also used elsewhere

Waqar144:
> why? `MainWindow::jumpToNoteSubFolder` already does that and is also used elsewhere

I also mean it's better to use the function.

Issue #327, opened 2016-09-18
Selecting multiple sub-folders/tags to view notes in all of these sub-folders/tags


Maboroshy:
There's a multiple selection available for tags and sub-folders but it's not connected to the note list. 

I think that selecting multiple sub-folders/tags should show a list of all notes for them in note list pane. It's logically expected behaviour and also a cool feature to have.

Can't say I really do need it personally, but that may open some interesting note grouping possibilities for guys with many notes.

pbek:
Currently there only can be one "selected tag" or "selected subfolder" and a lot of pieces of code depend on that selection in their operation. I don't think how wise it would be to mingle with that. It would be an awesome amount of work for sure...

Is there a use case for such a scenario?

Maboroshy:
My user case:
I was rearranging notes migrating to sub-folder structure and wanted to move all notes with several tags to specific sub-folder. Logically I've tried to select multiple tags with no avail. So I had to go through each tag, select all notes and move them with context menu. 

With proposed feature that would be N times faster, with N = number of tags. 

But this feature is not something I'd use on a daily or maybe even weekly basis.

livmion:
Hi all,

Actually that’s my typical way to use tags and how it works in other similar softwares, like Zotero or Zim. I use them in my daily work and I think this would be very useful to help finding the right note refered specifically to a sub-subject between many others with a similar thema.

bepolymathe:
Hello,

I agree with that. I often search for notes that mix several criteria (meeting + david + barcelona) ... To do this by selecting several tags would be great to find the right note.
In my case, this would be a daily use

milan-rusev:
+1 for that. It would be awesome to show the notes from all selected tags and all selected folders.

SuryaLuna:
:+1:  !!! PLEASE implement this :-)
I have the hope, that I found with QOwnNotes a software which fits my need almost completely, it's great!- but without this feature, the software would be useless for me...
Thanks a lot in advance :-)
(I need it especially for the tags)

iolivier87:
I need this feature too. Displaying matches between subtags and subfolders is very important to me.
I'm waiting for it impatiently in order to use QOwnNotes !
Thanks

zorglups:
Oh yes !
Please let me select tags 'a' and 'b' to show only the notes that do have BOTH tags 'a' AND 'b'.
This makes the tag meaningful and allow a really powerful search approach (ala taglocity).

pbek:
[...]

PR #1463, opened 2019-11-27
Add multiple folder selection to view notes in all selected folders[WIP]

Waqar144:
- Initial commit.

#327

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -9673,6 +9673,13 @@ void MainWindow::on_noteSubFolderTreeWidget_currentItemChanged(
         QTreeWidgetItem *current, QTreeWidgetItem *previous) {
     Q_UNUSED(previous)
 
+    auto items = ui->noteSubFolderTreeWidget->selectedItems();
+    //if multiple items are selected, we just set the last selected
+    //as active and return;
+    if (items.size() > 1) {
+        return;
+    }
+
     if (current == nullptr) {
         return;
     }
@@ -9698,6 +9705,43 @@ void MainWindow::on_noteSubFolderTreeWidget_currentItemChanged(
     }
 
     reloadTagTree();
+}
+
+void MainWindow::on_noteSubFolderTreeWidget_itemSelectionChanged()
+{
+    auto items = ui->noteSubFolderTreeWidget->selectedItems();
+    //if no items selected
+    if (items.count() < 1) {
+        return;
+    }
+
+    QList<int> noteSubFolderIds;
+    QList<int> noteIdList;
+    //get all the folder ids
+    Q_FOREACH(QTreeWidgetItem *item, items) {
+        noteSubFolderIds << item->data(0, Qt::UserRole).toInt();
+    }
+
+    // get the notes from the subfolders
+    Q_FOREACH(int noteSubFolderId, noteSubFolderIds) {
+        // get all notes of a note sub folder
+        QList<Note> noteList = Note::fetchAllByNoteSubFolderId(
+                    noteSubFolderId);
+        noteIdList << Note::noteIdListFromNoteList(noteList);
+    }
+
+    QTreeWidgetItemIterator it(ui->noteTreeWidget);
+
+    while (*it) {
+        // hide all notes that are not in selection
+        if (!noteIdList.contains((*it)->data(0, Qt::UserRole).toInt())) {
+            (*it)->setHidden(true);
+        } else {
+            (*it)->setHidden(false);
+        }
+
+        ++it;
+    }
 }
 
 /**
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -5176,11 +5176,12 @@ void MainWindow::filterNotes(bool searchForText) {
         filterNotesByNoteSubFolders();
     }
 
+    filterNotesByMultipleNoteSubFolders();
+
     // moved condition whether to filter notes by tag at all into
     // filterNotesByTag() -- it can now be used as a slot at startup
     filterNotesByTag();
 
-    filterNotesByMultipleNoteSubFolders();
 
     if (searchForText) {
         // let's highlight the text from the search line edit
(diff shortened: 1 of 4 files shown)


src/mainwindow.cpp
```diff
@@ -9700,6 +9707,43 @@ void MainWindow::on_noteSubFolderTreeWidget_currentItemChanged(
...
+                    noteSubFolderId);
+        noteIdList << Note::noteIdListFromNoteList(noteList);
+    }
+
+    QTreeWidgetItemIterator it(ui->noteTreeWidget);
+
+    while (*it) {
+        // hide all notes that are not in selection
+        if (!noteIdList.contains((*it)->data(0, Qt::UserRole).toInt())) {
+            (*it)->setHidden(true);
```
pbek: You cannot simply show or hide notes in the list. Don't forget about everything MainWindow::filterNotes() does! :smile:

Waqar144: Really? :thinking: 
But it does hide notes, and shows the ones from the selected folders :thinking: 
I don't know the implications this will have for all other modules though and the side effects...

`MainWindow::filterNotes()` would require a lot of changes to work with multi folder selection, right? or maybe I didn't understand you

pbek: You can always for example also search for text at the same time, working with `MainWindow::filterNotes()` is mandatory or you will immediately get bug reports. :grimacing:

pbek: Your new filters need to be part of the `MainWindow::filterNotes()` filtering process.

Waqar144: > You can always for example also search for text at the same time, working with `MainWindow::filterNotes()` is mandatory or you will immediately get bug reports. grimacing

I just tried this.
Multi select + search in noteTreeWidget. Works, but if you delete the searched text, the selection isn't retained.

Waqar144: > Your new filters need to be part of the `MainWindow::filterNotes()` filtering process.

Ok, let me try this.

Waqar144: Yes, Now its working much better I think

pbek: and if someone changes the search text after you selected note sub-folders? :grin:

Waqar144: :grimacing: 
The notes aren't getting filtered

Waqar144: Ok, so now the filtering is working with multi folder selection

pbek: Searching for text seems to work, but if I also select a tag than no notes are shown if more than one sub-folder was selected.

pbek: And if multiple sub-folders are selected then only the notes from the selected sub-folders are shown, not also their sub-folders (if the "including sub-folders" checkbox is checked in the settings)

Waqar144: > And if multiple sub-folders are selected then only the notes from the selected sub-folders are shown, not also their sub-folders (if the "including sub-folders" checkbox is checked in the settings)

Yes, I didn't do that because I thought the user might only want to see notes from the selected folder and note the subfolders too :thinking: 
Ok then, I will enable this if 'including-subfolders' is checked

> Searching for text seems to work, but if I also select a tag then no notes are shown if more than one sub-folder was selected.

Yes. That is not done yet. This is for round 2. :grin: 

Round 2 beings now.

Waqar144: Notes of subfolders are now also visible if show recursive setting is on.

Tags are working, but not as well as they should.
Tag count doesn't get updated for multiple folders.

The only way to fix that would be.. to modify `NoteSubFolder` class to take account of 'multiple-active-note-subfolders` and have another option like `isMultipleActive`

What do you think?

pbek: > Notes of subfolders are now also visible if show recursive setting is on.

https://imgur.com/a/ESYI0w7

With `Ctrl + Click` I was able to remove a selected sub-folder but the not the notes from the still selected sub-folder was shown. (there should be shown 15 notes in the end)

pbek: > Tags are working, but not as well as they should.

https://imgur.com/a/03QgQVq

Doesn't work for me, one sub-folder and one tag selected shows one note. when I add a 2nd sub-folder to the multi-selection no notes are shown

> Tag count doesn't get updated for multiple folders.

that would be ok for me, we can just note that limitation. of course if you also want to fix that you are free to do so, would save us the bug-report handling :laughing:

Waqar144: > Doesn't work for me, one sub-folder and one tag selected shows one note. when I add a 2nd sub-folder to the multi-selection no notes are shown

Yes, that's because when you click the second folder, that folder becomes active and since the second folder has  no associated tags with 'art', no notes are shown.

Tag only fetches notes for a certain selected folder at a time, right?
 
> > Tag count doesn't get updated for multiple folders.
> 
> that would be ok for me, we can just note that limitation. of course if you also want to fix that you are free to do so, would save us the bug-report handling laughing

Waqar144: Alright, tags are now working quite well I think

pbek: > Alright, tags are now working quite well I think

clicking on tags *after* selecting multiple su-folders doesn't seem to reduce the amount of notes shown in the note list :grin:

Waqar144: Ah Yes. Ok let me fix that.

pbek: > clicking on tags after selecting multiple sub-folders doesn't seem to reduce the amount of notes shown in the note list

I still have that issue.

Waqar144: Can you make a small vid like before?

Waqar144: Do you maybe have the 'recursive tagging' setting on? 🤔

Waqar144: Peek-2019-11-30-12-11

Waqar144: Please say something here too :grin:

pbek: Need more time to test that again, maybe tomorrow. I'm out the whole day today. 😬

Waqar144: Alright :grin:

pbek: > Do you maybe have the 'recursive tagging' setting on? thinking

yes

https://imgur.com/a/7cr6b2a

Multiple sub-folders selected. Clicking on tags didn't change the note list.

pbek: I am sorry, I tested one more time and didn't have the issue, I guess tray icon support shot me in the leg (an old version of the app may not have been closed). :grimacing:

Waqar144: Cool. One more feature down. :1st_place_medal: :grin:

PR #1465, opened 2019-11-29
Remove broken note links check to improve performance

Waqar144:
This check is now slowing down performance a lot so I am removing it. Note loading speed should increase a lot now, especially for notes larger than 200 lines.

### DIFF
--- a/src/helpers/qownnotesmarkdownhighlighter.cpp
+++ b/src/helpers/qownnotesmarkdownhighlighter.cpp
@@ -44,7 +44,6 @@ QOwnNotesMarkdownHighlighter::QOwnNotesMarkdownHighlighter(
 
     commentHighlightingOn = true;
     codeHighlightingOn = true;
-    highlightBrokenNotesLinkOn = true;
 }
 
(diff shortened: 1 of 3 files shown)


pbek:
But this breaks current functionality. If you want you can create a setting to disable those checks.

Waqar144:
Which functionality will it break? 🤔
This is just an extra check, that's unnecessary now.

pbek:
> that's unnecessary now.

where is the broken note links highlighting done otherwise?

Waqar144:
> where is the broken note links highlighting done otherwise?

The actual highlighting function is still in the highlighter. 

This was just a precheck we were doing on the whole document before loading it and  checking whether broken note link highlighting was required or not. Then our note loading speed was much slower so it was beneficial. Now we are much faster, so this check is slowing things down.

pbek:
Sorry, I didn't read the code well enough, I just read "Remove broken note links check" and saw code was being removed from the highlighter. :smile:

Waqar144:
> Sorry, I didn't read the code well enough, I just read "Remove broken note links check" and saw code was being removed from the highlighter. smile

:laughing:

PR #1468, opened 2019-11-30
Add GNU ASpell support for Linux

Waqar144:
Added **GNU ASpell** backend for faster and better spellchecking support especially for English language (Only on Linux for now)


pbek:
If it is breaking the current spellchecking I'd rather not want it in the develop branch :laughing:

Waqar144:
Seems like Travis has no Lib Aspell

Waqar144:
> If the PR it is breaking the current spellchecking I'd rather not want it in the develop branch 😆

Ahhahaha. No, it will not Break any existing functionality. It is an addition. Only one backend is used at a time. The option for switching backends is right above Languages menu. 
You should give it a try😉😜

Waqar144:
Dictionaries available at: https://ftp.gnu.org/gnu/aspell/dict/0index.html

This will provide much better spell checking for English language. It will also allow us to spell check languages like Finnish, which cannot be checked with Hunspell.

Easiest way to install dictionaries is using the package manager.

pbek:
> Easiest way to install dictionaries is using the package manager.

are we talking about "currently Linux-only support", right? are now external libraries needed to build QOwnNotes under Linux (that wouldn't be great)?

Waqar144:
> are we talking about "currently Linux-only support", right? 

Yes, it's linux only till now. Mac can be added easily (i think)

> are now external libraries needed to build QOwnNotes under Linux (that wouldn't be great)?

Yeah, libaspell is going to be necessary (it is available on all linux distros + mac).

What do you think?

src/mainwindow.ui
```diff
@@ -2541,6 +2547,11 @@ p, li { white-space: pre-wrap; }
     <string>Manage dictionaries</string>
    </property>
   </action>
+  <action name="actionasd">
+   <property name="text">
+    <string>asd</string>
```
pbek: What does that action? :) Is it a placeholder? The way you currently adding it it well need translation.

.travis.yml
```diff
@@ -43,7 +43,7 @@ install:
       sudo apt-add-repository -y ppa:ubuntu-toolchain-r/test
       sudo apt-add-repository -y ppa:beineri/opt-qt562-trusty
       sudo apt-get -qq update
-      sudo apt-get -qq install g++-4.8 libc6-i386 qt56tools qt56svg qt56script qt56websockets
+      sudo apt-get -qq install g++-4.8 libc6-i386 qt56tools qt56svg qt56script qt56websockets libaspell-dev
```
pbek: I'm very reluctant to now have an external dependency beside Qt to build at all 😞

src/mainwindow.ui
Waqar144: Nah, I didn't even add that. :thinking: 

The one I added was above that. I wonder how it got in there.

pbek: If it is a dummy-action that will be removed please name it that way and set the "no tranlation" flag on the text 😁

.travis.yml
Waqar144: Well,
this is optional. So I can abandon this if you want. No problems.

Embedding aspell, is going to be very very difficult.

I chose external lib because, aspell is available everywhere (except windows)

pbek:
> 
> Yeah, libaspell is going to be necessary (it is available on all linux distros + mac).
> 
> What do you think?

I don't like it. Beside me advertising that Qt is free of external dependencies beside Qt it will break all build systems not managed by myself. 😢

Waqar144:
> I don't like it. Beside me advertising that Qt is free of external dependencies beside Qt it will break all build systems not managed by myself. cry

Yeah for custom builds the user would have to `sudo xxx install libaspell-dev(or any of the 10 other names for aspell)` to build QON

.travis.yml
pbek: > Embedding aspell, is going to be very very difficult.

I wouldn't want to embed it neither for just having an optional 2nd spellchecker backend...
At best aspell support could be an optional flag that can be enabled at build time and if building doesn't take much longer this flag could be enabled by us in the different build systems.

Would users now also need to have the aspell installed to even run QOwnNotes?

Waqar144:
Alright I will leave this one :smile:

pbek:
> Yeah for custom builds the user would have to sudo xxx install libaspell-dev(or any of the 10 other names for aspell) to build QON

Some distros (like Fedora) have their own build system. Flatpack also isn't done by me.

pbek:
> Alright I will leave this one 😄

It could be a build flag.

Waqar144:
> It could be a build flag.

Like how?

pbek:
> Like how?

I guess like a env variable or qmake variable that could be set at build time, like I tried to do it with the initial Sonnet support I tried to add (I removed that code yesterday)

Waqar144:
> > Like how?
> 
> I guess like a env variable or qmake variable that could be set at build time, like I tried to do it with the initial Sonnet support I tried to add (I removed that code yesterday)

Ah yes, That's possible indeed.
But if the user or the packager wants aspell, or if we want to package it, it would still have to be included in all the build systems and scripts

Waqar144:
Maybe.. let me try and embed it. It's kind of a big library though.

Embedding it inside of QON (especially with windows support) would be a really big achievement.

src/mainwindow.ui
Waqar144: Done

.travis.yml
Waqar144: I have added the flags.

Let me test and confirm that it builds on mac/Linux with aspell.
Then I will add a global flag to turn it on/off completely.

> At best aspell support could be an optional flag that can be enabled at build time and if building doesn't take much longer this flag could be enabled by us in the different build systems.

Sounds good to me.

pbek:
> But if the user or the packager wants aspell, or if we want to package it, it would still have to be included in all the build systems and scripts

of course, but the build of users who don't include it will not break if they are not setting the flag  :grin:

Waqar144:
> > But if the user or the packager wants aspell, or if we want to package it, it would still have to be included in all the build systems and scripts
> 
> of course, but the build of users who don't include it will not break if they are not setting the flag 😁

The flag is in place now 🚀

Even with the flag on, the build system will check if aspell was found on the system and only then will the aspell backend get enabled.

src/mainwindow.cpp
```diff
@@ -102,6 +102,7 @@
 #include <utility>
 #include <QScreen>
 #include "libraries/sonnet/src/core/speller.h"
+#include "loader_p.h"
```
pbek: does this also need the #ifdef?

src/mainwindow.ui
```diff
@@ -2541,6 +2547,11 @@ p, li { white-space: pre-wrap; }
     <string>Manage dictionaries</string>
    </property>
   </action>
+  <action name="actionasd">
```
pbek: Do we need this action?

build-systems/aur/PKGBUILD
```diff
@@ -9,7 +9,7 @@ arch=('i686' 'x86_64' 'armv7h' 'aarch64')
 url='http://www.qownnotes.org/'
 license=('GPL2')
 groups=('qownnotes')
-depends=('qt5-base' 'qt5-svg' 'qt5-declarative' 'openssl' 'qt5-xmlpatterns' 'qt5-websockets')
+depends=('qt5-base' 'qt5-svg' 'qt5-declarative' 'openssl' 'qt5-xmlpatterns' 'qt5-websockets' 'aspell')
```
pbek: Do we want all those depends without the flag?

Waqar144: No.
I will disable these

src/mainwindow.cpp
Waqar144: Nopes. That's a relic

src/mainwindow.ui
Waqar144: I don't think so.
Will remove it as well

src/libraries/sonnet/src/core/sonnet-core.pri
```diff
@@ -35,6 +35,29 @@ HEADERS += $$PWD/client_p.h \
...
+        include($$PWD/../plugins/aspell/aspell.pri)
+        LIBS += -laspell
+    }
+}
+}
+
+#windows
+CONFIG(with_aspell) {
+win32: {
+    exists ("C:\\Qt\\5.13\\mingw73_32\\include\\aspell.h") {
```
pbek: That will break on AppVeyor as soon Qt 5.13 will not be used any more. And for the rest of the users it will probably never work. :grimacing:

Waqar144: Yes, this was just for the sake of completeness. :grimacing: 
It isn't supposed to work lol :grimacing: 

Aspell on windows has poor support.

pbek: Ok :)

pbek:
I turned on `CONFIG += with_aspell` in the project file but wasn't able to see the new menu item for selecting aspell. Did I miss anything?

Waqar144:
> I turned on `CONFIG += with_aspell` in the project file but wasn't able to see the new menu item for selecting aspell. Did I miss anything?

On Linux?

Make sure you have aspell installed.
and have aspell headers at `/usr/include/aspell.h`

pbek:
Works great now.

1) I didn't have libaspell-dev. I saw you actively check if `/usr/include/aspell.h` exists. Does that mean we can enable aspell support by default?

2) The dictionary manager doesn't work if aspell is selected, right? shouldn't we disable it in that case?

Waqar144:
> Works great now.
> 
>     1. I didn't have libaspell-dev. I saw you actively check if `/usr/include/aspell.h` exists. Does that mean we can enable aspell support by default? That would mean you could also enable all those repo depends again, right?

Well, yeah. We can enable it by default and it would only include `aspell` if we find aspell in the system. I put the extra check in there so that build systems don't fail.

>     2. The dictionary manager doesn't work if aspell is selected, right? shouldn't we disable it in that case?

Doesn't work? :thinking: 
You mean it doesn't have much purpose then, right?
It can be disabled, but `hunspell` is always enabled, always available so, I think if we leave it as it is, its probably fine. We can put a note on top saying, these dictionaries are for hunspell only.

pbek:
1) Awesome!
2) Hunspell is also enabled even if the backend is set to "Aspell"? Why the setting then? :grin:

Waqar144:
>     2. Hunspell is also enabled even if the backend is set to "Aspell"? Why the setting then? 😁

Haha, I mean, the hunspell backend is going to stay, it's always there as a fallback. But yeah, I think we should hide the dictionaries manager in case *aspell* is enabled.

Waqar144:
> Works great now.
> 
>     1. I didn't have libaspell-dev. I saw you actively check if `/usr/include/aspell.h` exists. Does that mean we can enable aspell support by default? That would mean you could also enable all those repo depends again, right?
> 
>     2. The dictionary manager doesn't work if aspell is selected, right? shouldn't we disable it in that case?

Should I re-enable those repo depends?

pbek:
> Should I re-enable those repo depends?

Yes, you can (in good hope that they will not break the build process :laughing:)

Waqar144:
> > Should I re-enable those repo depends?
> 
> Yes, you can (in good hope that they will not break the build process laughing)

I will send in the PR in a short while

PR #1471, opened 2019-12-01
notehistory: Use note id directly instead of querying database

Waqar144:
Performance improvements

### DIFF
--- a/src/entities/notehistory.cpp
+++ b/src/entities/notehistory.cpp
@@ -155,7 +155,7 @@ QDataStream &operator>>(QDataStream &in, NoteHistoryItem &item) {
  */
 
 NoteHistory::NoteHistory() {
-    noteHistory = new QList<NoteHistoryItem>;
+    noteHistory = new QVector<NoteHistoryItem>;
     currentIndex = 0;
 }
 
@@ -163,7 +163,7 @@ NoteHistory::NoteHistory(const NoteHistory &h) :
     currentIndex(h.currentIndex),
     currentHistoryItem(h.currentHistoryItem)
 {
-    noteHistory = new QList<NoteHistoryItem>;
+    noteHistory = new QVector<NoteHistoryItem>;
     *noteHistory = *h.noteHistory;
 }
 
@@ -172,7 +172,7 @@ NoteHistory &NoteHistory::operator=(const NoteHistory &rhs) {
         return *this;
 
     delete noteHistory;
-    noteHistory = new QList<NoteHistoryItem>;
+    noteHistory = new QVector<NoteHistoryItem>;
     *noteHistory = *rhs.noteHistory;
 
     currentIndex = rhs.currentIndex;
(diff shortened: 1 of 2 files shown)


pbek:
Does the note history still work if you restart the application (are the correct notes of the last session found in the history at the correct position)?

pbek:
Because it seems you are now relying on note ids (which are not persisted on disk!) instead of the note name and sub-folder!

Waqar144:
Yes, that's not happening. I was going to use the existing implementation as fallback, so that the note history is correctly loaded, and with time after the first time, note id would be used. But I wanted you to take a look at this before I went any further (as this may have other side effects that I may not be aware of)

pbek:
> and with time after the first time, note id would be used.

and if the note folder is reloaded (and the note index is regenerated)?

Waqar144:
> and if the note folder is reloaded (and the note index is regenerated)?

If the index is regenerated then yeah, the old note_ids would be pretty much useless for this purpose.

pbek:
> If the index is regenerated then yeah, the old note_ids would be pretty much useless for this purpose.

do you have to adapt your implementation?

Waqar144:
> do you have to adapt your implementation?

```cpp
             if (item.getNoteId() == note.getId()) {
                 return item;
            } else if (item.getNote().getId() == note.getId()){
               return item;
             }
```

This would work, but its pointless, since we are doing the same thing as before.

We can have something like:
```cpp
     //first for loop
        for (int i = 0; i < noteHistory->count(); ++i) {
         ....
        }
//fall back.
        for (int i = 0; i < noteHistory->count(); ++i) {
            const NoteHistoryItem &item = noteHistory->at(i);

            if (item.getNote().getId() == note.getId()){
                return item;
            }
        }
```

Waqar144:
Are you getting a lag while switching notes (happens sometimes, often with small notes)?

pbek:
> Are you getting a lag while switching notes (happens sometimes, often with small notes)?

you mean with this PR or in general? in general I can't say that I see any lags.

pbek:
> This would work, but its pointless, since we are doing the same thing as before.

You could just also wipe the NoteIds in the history when all notes are wiped off the database... But the problem is notes can also be removed individually, you would also need to handle that...

Is this all really worth the effort? Is there such a high note switching speed improvement?

Waqar144:
> You could just also wipe the NoteIds in the history when all notes are wiped off the database... But the problem is notes can also be removed individually (e.g. deleted or moved), you would also need to handle that...
> 
> Is this all really worth the effort? Is there such a high note switching speed improvement?

Not really, It just gives me a 12ms lag at most.
This was a small lag I noticed.

Going after this is going to cause more regressions and a lot of effort. I will close this now.

Waqar144:
> you mean with this PR or in general? in general I can't say that I see any lags.

It is happening with 2 or 3 specific notes. I have tried really hard to find out where the lag is happening but have been unable to do so till now. Maybe it is something related to the note file.

pbek:
12ms for a in-memory sqlite query by string? Seems high. How many notes do you have? Maybe we should use more sqlite indices.

Waqar144:
> 12ms for a in-memory sqlite query by string? Seems high. How many notes do you have? Maybe we should use more sqlite indices.

about 34 or so where I am testing. 

Each call to `getNote()` is taking about a 300 - 600 microseconds usually on my pc.

pbek:
You could try to add an index for the note table.

PR #1474, opened 2019-12-03
Allow syntax highlighting

Waqar144:


src/helpers/qownnotesmarkdownhighlighter.cpp
```diff
@@ -134,6 +134,7 @@ void QOwnNotesMarkdownHighlighter::highlightMarkdown(const QString& text) {
     if (codeHighlightingOn) {
         if (previousBlockState() == HighlighterState::CodeBlock ||
             previousBlockState() == HighlighterState::CodeBlockEnd ||
+            previousBlockState() >= 200 ||
```
pbek: Don't you want to use the constant here?

Waqar144: Ah yes, just a sec

pbek:
Looks great already, but of course you can't see a lot in QOwnNotes when using dark mode.
That would need addition to `FontColorWidget::initTextTreeWidgetItems()` and `QOwnNotesMarkdownTextEdit::setStyles()` and integration in all configs of `schemes.conf`.

Waqar144:
> Looks great already, but of course you can't see a lot in QOwnNotes when using dark mode.
> That would need addition to `FontColorWidget::initTextTreeWidgetItems()` and `QOwnNotesMarkdownTextEdit::setStyles()` and integration in all configs of `schemes.conf`.

Yes,

The normal code block font color is still being followed (for everything that doesn't get highlighted)

I have divided the code words into 3 types:
- keywords
- basic types
- other/special/preprocessor 

This would work for a lot of languages I think. But it may not for some others.

Also my choice of colors wasn't the most aestheic I guess :grimacing: :rofl:

pbek:
Yes, there need to be three new colors to be configured. Plus default colors for light and for dark themes in `schemes.conf`.

pbek:
https://ci.appveyor.com/project/pbek/qownnotes/builds/29273832/job/yrp5onwjiyoba34y is breaking.

pbek:
My fault, need to merge qmarkdowntextedit later...

Waqar144:
> Yes, there need to be three new colors to be configured. Plus default colors for light and for dark themes in `schemes.conf`.

Is `schemes.conf` meant to be editted by hand? or is there a function which I can use to  write the styles?

pbek:
> Is schemes.conf meant to be editted by hand? or is there a function which I can use to write the styles?

Unfortunately all schemas have to be edited by hand. Of course you could create a copy of a schema in the QON settings and then export and copy the settings into `schemes.conf` :grimacing:

Waqar144:
> > Is schemes.conf meant to be editted by hand? or is there a function which I can use to write the styles?
> 
> Unfortunately all schemas have to be edited by hand. Of course you could create a copy of a schema in the QON settings and then export and copy the settings into `schemes.conf` grimacing

That's what I am doing lol :rofl:

pbek:
Usually I just check which keys I need and copy the new settings to the different schemata. But you have to do that for light and dark ones individually.

Waqar144:
> Usually I just check which keys I need and copy the new settings to the different schemata. But you have to do that for light and dark ones individually.

Should I set the default `CodeBlock` colors to black and white for light and dark themes, since green looks weird in between the colors

pbek:
I quite like the green. :grinning:

Waqar144:
> I quite like the green. grinning

Well then I left it as green :grin:

Waqar144:
## Some syntax highlighting comparisons with `Highlight.js`

 (Left: QON, Right: Highlight.js)

### Cpp
cpphi

### JS
jshi
jsanotherhi

### PHP
phphi

### Python
pythonhi

### QML
qml

## Performance
10K lines in 330~ ms

## Missing features
- very small number of langs supported
- No support for multi line comments or # comments
- No lib function highlighting (something that may never happen)

pbek:
Very nice already! Great work!

pbek:
You really also can swap the green default code block color if you want.

Waqar144:
> You really also can swap the green default code block color if you want.

Only reason I didn't do that was because we don't support a lot of languages yet. So for the ones we don't support, green is still kind of nice

gerroon:
Hi

Is this feature in? I have some py blocks and all I see is green and white.

pbek:
```python
    import test
    ```

...works for me

gerroon:
Thanks this is what I see 

https://i.imgur.com/BKgoFof.png

Waqar144:
Seems to me as if color schemas are not working correctly for you, the highlighting is working

gerroon:
@Waqar144  Any speculation on why that might be the case?

Waqar144:
> @Waqar144 Any speculation on why that might be the case?

Can't say for sure. Please check if there are editor schemas (themes) available in settings and you have selected a style that actually does some coloring. You can see the available Schemas in Settings -> Editor Fonts

PR #1475, opened 2019-12-03
Add code syntax coloring options to settings

Waqar144:


pbek:
I guess the other schemata would also need those colors. :grimacing:

Waqar144:
> I guess the other schemata would also need those colors. grimacing

Yeah :grimacing:

Waqar144:
5 done

pbek:
The schema "Light" also code colors which background, was that on purpose?

pbek:
If you want you can change the general colors for code blocks also away from green. :grin:

pbek:
3) JS "support" isn't done yet, right? looks a bit funny with the white function declaration. :grin: 

image

Waqar144:
>     1. The schema "Light" also code colors with background, was that on purpose?

I am just copying the background color of `codeblock` so that it looks uniform

Waqar144:
>     1. JS "support" isn't done yet, right? looks a bit funny with the white function declaration. grin
> 


That's a bug. Fixing

PR #1478, opened 2019-12-03
Enable aspell on linux build systems

Waqar144:


pbek:
I keep my fingers crossed :grin:

pbek:
Releasing now...

Waqar144:
> Releasing now...

I have my eyes on OBS :eyes:

pbek:
SLE doesn't find the aspell library...

Waqar144:
> SLE doesn't find the aspell library...

I think the name for aspell there is `aspell-devel`

pbek:
> I think there is no aspell dev library on SLE. 

Can you come up with a "if-clause" to don't fetch aspell for SLE?

Waqar144:
Looking into it

PR #1479, opened 2019-12-04
Attempt fix SLE build on OBS

Waqar144:
Maybe `aspell-devel` would work there. Not sure

### DIFF
--- a/obs/qownnotes.spec
+++ b/obs/qownnotes.spec
@@ -24,6 +24,13 @@ BuildRequires:  qt5-qtwebsockets-devel
 BuildRequires:  desktop-file-utils
 BuildRequires:  aspell-devel
 Requires:       qt5-qtsvg qt5-qtxmlpatterns qt5-qtwebsockets
+%endif
+
+%if 0%{?sle_version} == 150000 || 0%{?sle_version} == 150100 && !0%{?is_backports}
+BuildRequires:  libqt5-qtbase-devel libqt5-qtdeclarative-devel libQt5Svg-devel libQt5XmlPatterns-devel libqt5-qtwebsockets-devel
+#BuildRequires:  aspell-devel
+BuildRequires:  update-desktop-files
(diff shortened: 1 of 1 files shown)


obs/qownnotes.spec
```diff
@@ -24,6 +24,13 @@ BuildRequires:  qt5-qtwebsockets-devel
...
 Requires:       qt5-qtsvg qt5-qtxmlpatterns qt5-qtwebsockets
+%endif
+
+%if 0%{?sle_version} == 150000 || 0%{?sle_version} == 150100 && !0%{?is_backports}
+BuildRequires:  libqt5-qtbase-devel libqt5-qtdeclarative-devel libQt5Svg-devel libQt5XmlPatterns-devel libqt5-qtwebsockets-devel
+#BuildRequires:  aspell-devel
+BuildRequires:  update-desktop-files
+Requires:       libQt5Svg5 libQt5Sql5 libQt5Sql5-sqlite libQt5Gui5 libQt5Network5 libQt5Widgets5 libQt5Xml5 libQt5XmlPatterns5 libQt5PrintSupport5 libQt5WebSockets5
 
 %else
```
pbek: This else-clause will now also be execute for all other distributions other then SLE, don't you think? I'll try to put your new if-statement inside the old else-clause.

pbek: so just

```
%if 0%{?sle_version} == 150000 || 0%{?sle_version} == 150100 && !0%{?is_backports}
#BuildRequires:  aspell-devel
%else
BuildRequires:  libaspell15
%endif
```

Waqar144: maybe it's aspell-devel for both. I think libaspell15 is something different.. not really sure though

pbek: we'll see, I manually added my lines in OBS, currently building

pbek: SLE 15 is already built

pbek: Looking good. Thanks for the if-statement!

Waqar144: Great :tada:

PR #1481, opened 2019-12-05
Implement filtering by multiple tags

Waqar144:
#327

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -8309,7 +8309,13 @@ void MainWindow::on_tagTreeWidget_itemSelectionChanged() {
 
     int count = ui->tagTreeWidget->selectedItems().count();
 
-    if (count <= 1) return;
+    if (count <= 1) {
+        if (count == 1) {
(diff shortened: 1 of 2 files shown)


pbek:
I knew you will take tags up too! :grin: 
Doesn't seem to work in every situation for me when unselecting tags.

https://imgur.com/a/6fKXtNO

pbek:
After single selecting I selected two tags, that worked. But when un-selecting tags notes wasn't filtered correctly. I guess that's because only one tag was selected any more and I clicked on the tag that I wanted un-selected. I don't know how to get around this. :grimacing:

Waqar144:
> I knew you will take tags up too! grin
> Doesn't seem to work in every situation for me when unselecting tags.
> 
> https://imgur.com/a/6fKXtNO

It seemed pretty straightforward after I looked into it :laughing:

Waqar144:
> After single selecting I selected two tags, that worked. But when un-selecting tags notes wasn't filtered correctly. I guess that's because only one tag was selected any more and I clicked on the tag that I wanted un-selected. I don't know how to get around this. grimacing

Fixed, I think

pbek:
Awesome!

Issue #569, opened 2017-05-02
Feature request: Moving paragraph one up or down


L-R-S:
Will be nice and helpful if will be possible to move paragraphs one up and down. Like in Kate (Ctrl + Shift + ↑/↓) or Libre Office (Ctrl + Alt + ↑/↓).

pbek:
Thank you for your suggestion.

Maboroshy:
It looks like a request for a proper note text editing by scripting engine to me. ;-)

quailiff:
![move_text_up_down](https://cloud.githubusercontent.com/assets/16974626/26664085/00d7795e-465d-11e7-9a4b-b4e506c234cf.png)

I was able to get this working at [quailiff/feature/text-manipulation](https://github.com/quailiff/QOwnNotes/tree/feature/text-manipulation) (click [here](https://github.com/pbek/QOwnNotes/compare/develop...quailiff:feature/text-manipulation) to see changes against pbek/develop). Comments:

- So far, I have only tested this on Linux (not Windows or macOS).
- I can probably combine the logic in each function into one function (maybe in Utils::Misc) that accepts the offset to move the current selection (negative to move up and positive to move down).
- I was only able to get the application to call `on_actionMove_text_up_one_line_triggered` and `on_actionMove_text_down_one_line_triggered` to work by adding two new menu items as shown above. When I removed the menu items, the shortcuts didn't work and there is no other way to trigger the functions. Maybe shortcuts can be set in the constructor? However, if there are no menu items, will it still be possible to customize the shortcuts in the Settings dialog?
- It looks like I can only use [`QTextDocument::toPlainText()`](https://doc.qt.io/qt-5/qtextdocument.html#toPlainText), which replaces "some" (whatever that means) Unicode characters with ASCII.
- **The worst problem with this approach:** Moving text blocks up and down works. However, I use [`QTextEdit::setPlainText`](https://doc.qt.io/qt-5/qtextedit.html#setPlainText) to update the text, which erases the undo history.

pbek:
Hey, great you took on an other challenge, @quailiff!
New menu entries would be ok in that case, I guess. In the worst we can think about creating a submenu for such operations.
Hmmm, maybe you can try using the text cursor and select the text block, then removing the selected text, moving the cursor one text block above/below and insert the text again. That would surely keep the history and you don't need `toPlainText`. There should be plenty of examples of how to work with the text cursor in `mainwindow.cpp`.

[...]

PR #1482, opened 2019-12-05
Update shortcuts

Waqar144:
#569


pbek:
There also is a 2nd block for macOS. :grin:

PR #1483, opened 2019-12-05
Fix deselection of multi folder selection

Waqar144:
#327

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -9845,6 +9845,14 @@ void MainWindow::on_noteSubFolderTreeWidget_currentItemChanged(
 }
 
 void MainWindow::on_noteSubFolderTreeWidget_itemSelectionChanged() {
+    auto items = ui->noteSubFolderTreeWidget->selectedItems();
+    //if no items selected or only one selected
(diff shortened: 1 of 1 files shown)


pbek:
Works great!

pbek:
I was testing the multi-tag deselection. But your PR wasn't about that. :) Why does the the multi-tag deselection work now? :laughing:

pbek:
Hm, and if I select one sub-folder. The tag multi-selection will be removed and only one tag will be selected.

Waqar144:
> Hm, and if I select one sub-folder. The tag multi-selection will be removed and only one tag will be selected.

Ah yes, that's because tag tree is reloaded right after `filternotes`, 
There might be a way around that I think. 
But if you select multiple folders first and then select multiple tags, that works.

Waqar144:
> I was testing the multi-tag deselection. But your PR wasn't about that. :) Why does the the multi-tag deselection work now? 😆

It was working previously, no? 🤔😅

pbek:
Yes it does... I think I misread https://github.com/pbek/QOwnNotes/issues/327#issuecomment-562126920

pbek:
I reverted this PR because of #1489.

Issue #1489, opened 2019-12-07
Notes are not shown if no note sub-folder was selected


thomascrisan:
Allowed application to update itself to windows.  Upon restart, all of my notes disappeared.

Reinstalled v5038 manually by copying downloaded zip file contents and obtained same result.  Checked setup directories and server entries and all was okay.  Re-read directories but again, all notes were missing

Reverted to v5030 and restored functionality in Windows.  No notes were altered by new version, they simply were not being read and made visible.

[QOwnNotes.log](https://github.com/pbek/QOwnNotes/files/3935201/QOwnNotes.log)

jebrosen:
Same issue here with version 19.12.3 (Arch Linux, via AUR) - the sidebar on the left side with the list of notes always appears empty.

thomascrisan:
That's odd; my usual desktop is Arch Linux and when I installed version 19.12.3 via AUR earlier, everything works as expected.  The left sidebar is fully populated.

Waqar144:
Seems like note directory isn't read, everything is empty in your log. Very weird...

pbek:
I neither can reproduce this on Manjaro (AUR). @Waqar144, could #1483 do this?

dm2912:
Same problem here in Mint 19.2

Waqar144:
> I neither can reproduce this on Manjaro (AUR). @Waqar144, could #1483 do this?

Seems like it.

pbek:
I was able to reproduce it. I had to click on a note sub-folder to see any notes. If sub-folders weren't activated I never got any notes in the list. #1483 was the culprit. It worked after I reverted it.

pbek:
## 19.12.4
- fixed a problem with loading notes if no note sub-folder was selected

pbek:
There now is a new release, could you please test it and report if it works for you?

thomascrisan:
release 5050 fixed the issue for me on the Windows version.  Thanks.

pbek:
Great, thank you for testing!

dm2912:
fine here in mint now

PR #1492, opened 2019-12-08
Fix notes not visible when subfolders are off

Waqar144:
#1489

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -5188,8 +5188,6 @@ void MainWindow::filterNotes(bool searchForText) {
         filterNotesByNoteSubFolders();
     }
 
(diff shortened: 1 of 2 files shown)


pbek:
I think you pushed the reference to an older version of qmarkdowntextedit.

Waqar144:
I did run git submodule update and pushed again. But it doesn't seem to change anything . What do I do?

pbek:
Go into the qmarkdowntextedit directory and switch to the develop branch, then you can commit that in the QOwnNotes directory.

pbek:
I'll do it for you. :)

Waqar144:
> I'll do it for you. :)

Thanks :grimacing:

Issue #943, opened 2018-06-16
Slow navigation between folders with @tag script


OlivierHobbes:
#### Expected behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
I expect to browse folders and sub-folders in a fluid way.

#### Actual behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
Each time I select a folder, I must to wait a few seconds to see what it contains.

#### Steps to reproduce  
My Note folder contains 488 notes in 4 main folders and 17 sub-folders.
I measure the time to go through the 4 main folders in various conditions .

PC1 : Windows 10 + i3-6100U @2.3 GHz + HDD 7200 rpm + 8GB RAM
PC2 : Windows 10 + i3-2120 @3.3 GHz + SSD + 8GB RAM

#1 : PC1 + Encrypted Note folder with Cryptomator : 20 seconds
#2 : PC1 + No encryption : 11 seconds           
#3 : PC2 + Encrypted Note folder with Cryptomator : 7 seconds
#4 : PC2 + No Encryption : 4 seconds (not instantaneous but quite usable)

NB : External modifications are ignored and Note save interval time is 30s.

The script "@tag tagging in note test (experimental)" was active during this test. So, I start again without it : 

#5 : PC1 + Cryptomator + No script : 4 seconds
Ok, the @tag script seems guilty !

#### Comment
The loading time of the folders seems to depend a lot on storage type.
However, browsing notes in the same folder is instantaneous in all above conditions.

I'm a noob but perhaps an optional parameter to preload folders in RAM could make it more fluid ?
Another suggestion, the @tag script might be triggered after an adjustable delay.

pbek:
Maybe @Maboroshy can help. :)

iolivier87:
Thank you for your reply.
So I think I will run the script once a day. Is there a simple way to have a button in the toolbar that parses the folders and syncs the tags with the database ? (not important, but it could be a workaround)

pbek:
You would need to write a different script for that.

On 16 Jun 2018 16:39, "iolivier87" <notifications@github.com> wrote:

Thank you for your reply.
So I think I will run the script once a day. Is there a simple way to have
a button in the toolbar that parses the folders and syncs the tags with the
database ? (not important, but it could be a workaround)

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
[...]

PR #1499, opened 2019-12-11
Optimize folder switching when tags are enabled

Waqar144:
checking for broken `tagnotelinks` can take a lot of time and doing that every time we switch folders doesn't seem like a good idea. `Tagnotelinks` are automatically taken care of whenever a tag or note is removed.

This function is good in case the user deletes the notes using file manager. In that case we just run this once during startup and it remains disabled after that.

If the user still deletes notes (and QON is open), it would be taken care of in the next run of QON.

#943

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -7294,6 +7294,7 @@ void MainWindow::on_noteFolderComboBox_currentIndexChanged(int index) {
 
     if (noteFolder.isFetched()) {
         changeNoteFolder(noteFolderId);
+        areBrokenTagNoteLinksRemoved = false;
     }
 
     // hide the noteSubFolderDockWidget menu entry if sub-folders are
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1725,6 +1725,9 @@ void MainWindow::changeNoteFolder(int noteFolderId, bool forceChange) {
     // store the note history of the old note folder
     noteHistory.storeForCurrentNoteFolder();
 
+    //recheck for broken tag note links
+    resetBrokenTagNotesLinkFlag();
+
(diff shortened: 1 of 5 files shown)


pbek:
> If the user still deletes notes (and QON is open), it would be taken care of in the next run of QON.

It would only take care of it for the note folder that is active while starting QOwnNotes... The tags of other note folders will never be cleaned.

Potentially every time an external process removes notes or folders the tags could be affected.

Waqar144:
> It would only take care of it for the note folder that is active while starting QOwnNotes... The tags of other note folders will never be cleaned.

We can reset the switch when a user switches note folders. 

> Potentially every time an external process removes notes or folders the tags could be affected.

Yes, and the next run of QON should be able to take care of that. Running this everytime we switch subfolders is too expensive

pbek:
For speed's sake I'm OK with just doing it at every folder switch and initial start.

src/mainwindow.cpp
```diff
@@ -7294,6 +7294,7 @@ void MainWindow::on_noteFolderComboBox_currentIndexChanged(int index) {
 
     if (noteFolder.isFetched()) {
         changeNoteFolder(noteFolderId);
+        areBrokenTagNoteLinksRemoved = false;
```
pbek: you can also switch the note folder via the settings dialog and the `Note / Note folders` menu (you can hide the note folder selector panel), is this line also triggered in those cases?

pbek: And can you please start all new global variables with an underline, `_areBrokenTagNoteLinksRemoved` in this case. :grin:

Waqar144: Done

pbek:
Looking good!

PR #1501, opened 2019-12-11
Performance optimizations

Waqar144:
- Use QStringLiteral in note.cpp everywhere
- Use reserve to reserve memory before appending elements to `QLists` for better performance.

### DIFF
--- a/src/entities/note.cpp
+++ b/src/entities/note.cpp
@@ -332,9 +332,9 @@ bool Note::copyToPath(const QString& destinationPath, QString noteFolderPath) {
             // find a new filename for the note
             QDateTime currentDateTime = QDateTime::currentDateTime();
             destinationFileName =
-                    destinationPath + QDir::separator() + this->name + " " +
-                    currentDateTime.toString(Qt::ISODate).replace(":", "_") +
-                    "." + defaultNoteFileExtension();
+                    destinationPath + QDir::separator() + this->name + QChar(' ') +
+                    currentDateTime.toString(Qt::ISODate).replace(QChar(':'), QChar('_')) +
+                    QChar('.') + defaultNoteFileExtension();
 
             qDebug() << "New file name:" << destinationFileName;
         }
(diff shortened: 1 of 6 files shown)


pbek:
Awesome! You really take care of performance!

Waqar144:
> Awesome! You really take care of performance!

Haha, I do. And the reasons are
- I have an older pc (2 core) :(
- Have seen a lot of users(including myself) abandon something just because it's slow
- And there's not much point in using c++ if we aren't faaaaaast. :grin:

pbek:
> I have an older pc (2 core) :(

It's great to use older hardware for such things!

> Have seen a lot of users(including myself) abandon something just because it's slow

Which might rule out to use a fully fledged browser in QON. :grimacing: 

> And there's not much point in using c++ if we aren't faaaaaast. grin

in contrast to Electron apps :laughing:

pbek:
> And there's not much point in using c++

but I'm a big Qt fanboy! :grin:

Waqar144:
> > And there's not much point in using c++
> 
> but I'm a big Qt fanboy! grin

Me too :rofl:

> Which might rule out to use a fully fledged browser in QON. grimacing

With syntax highlighting, we are feature complete so in my opinion, a person doesn't even need to go to preview. Everything is being rendered correctly in the text editor :grin:

pbek:
> With syntax highlighting, we are feature complete so in my opinion, a person doesn't even need to go to preview. Everything is being rendered correctly in the text editor grin

True. But still the first thing that people ask me when they see your beautiful code block syntax highlighting is if it is possible to also have that in the preview. :grinning:

Waqar144:
> > With syntax highlighting, we are feature complete so in my opinion, a person doesn't even need to go to preview. Everything is being rendered correctly in the text editor grin
> 
> True. But still the first thing that people ask me when they see your beautiful code block syntax highlighting is if it is possible to also have that in the preview. grinning

Yeah :grimacing: 
some day we'll have it there too maybe :grin: 

It wouldn't be slow. with QWebEngine... It would just take more memory

pbek:
> It wouldn't be slow. with QWebEngine... It would just take more memory

Or we write our own markdown2html converter. :grin:

PR #1502, opened 2019-12-12
Use QRegularExpression for newer Qt versions when searching

Waqar144:
`QRegularExpression` is supported since Qt 5.13 in `QPlaintextEdit`. Its quite a bit faster so lets use it where we can.


pbek:
As far as I remember the syntax of QRegExp and QRegularExpression was not 100% compatible, but I hope I remember wrongly. :grin:

Waqar144:
> As far as I remember the syntax of QRegExp and QRegularExpression was not 100% compatible, but I hope I remember wrongly. grin

They are different :grimacing: 
But this regex is quite simple and seems to be working so hopefully there will not be any problems

PR #1503, opened 2019-12-12
Fix some warnings being emitted by clazy[WIP]

Waqar144:
Lots of QString changes
Almost all the signal/slot connection is now done in the modern style.
Some constness changes

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -4792,7 +4792,7 @@ void MainWindow::handleTextNoteLinking() {
  * @brief Sets the current note from a CurrentNoteHistoryItem
  * @param item
  */
-void MainWindow::setCurrentNoteFromHistoryItem(NoteHistoryItem item) {
+void MainWindow::setCurrentNoteFromHistoryItem(const NoteHistoryItem &item) {
     qDebug() << item;
     qDebug() << item.getNote();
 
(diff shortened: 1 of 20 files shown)


src/mainwindow.cpp
```diff
@@ -1648,7 +1627,7 @@ void MainWindow::loadNoteFolderListMenu() {
     int c = 0;
     Q_FOREACH(QAction* action, actions) {
             // start with the 2nd item, the first item is the menu itself
-            if (c++ > 0) {
+            if (++c > 0) {
```
pbek: Are you sure you can do this? Value of `c` will be increased before assignment.

Waqar144: ah, no. Thanks. I will revert this.

pbek: I will undo that change.

pbek: Ok, I'll wait a minute more. :grin:

Waqar144: eeh. Don't merge this yet :grimacing: 
I am preparing some more changes :grimacing:

pbek: Done?

src/mainwindow.h
```diff
@@ -689,6 +689,13 @@ private slots:
     QActionGroup *languageGroup;
     QActionGroup *spellBackendGroup;
     bool _brokenTagNoteLinksRemoved = false;
+    const QIcon tagIcon = QIcon::fromTheme(
```
pbek: can you please use `_` as prefix for global variables

src/mainwindow.cpp
Waqar144: Done.

src/mainwindow.h
Waqar144: Yes, sorry I am really forgetful 😬

pbek:
I accepted the PR, you can do the global variables in another PR. :grin:

PR #1505, opened 2019-12-13
Use _ with global variable names

Waqar144:
-  Use QString() instead of creating empty strings
#1503

### DIFF
--- a/src/utils/gui.cpp
+++ b/src/utils/gui.cpp
@@ -30,7 +30,7 @@
  * Checks if there is at least one child that is visible
  */
 bool Utils::Gui::isOneTreeWidgetItemChildVisible(QTreeWidgetItem *item) {
-    for (int i = 0; i < item->childCount(); ++i) {
+    for (int i = 0; i < item->childCount(); i++) {
         QTreeWidgetItem *child = item->child(i);
         if (!child->isHidden() || isOneTreeWidgetItemChildVisible(child)) {
             return true;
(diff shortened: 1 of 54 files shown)


src/utils/gui.cpp
```diff
@@ -30,7 +30,7 @@
  * Checks if there is at least one child that is visible
  */
 bool Utils::Gui::isOneTreeWidgetItemChildVisible(QTreeWidgetItem *item) {
-    for (int i = 0; i < item->childCount(); i++) {
+    for (int i = 0; i < item->childCount(); ++i) {
```
pbek: This also does not the same than `i++` in this case!

src/utils/gui.cpp
```diff
@@ -69,7 +69,7 @@ void Utils::Gui::searchForTextInTreeWidget(QTreeWidget *treeWidget,
                 bool show = false;
 
                 // look in all columns that we want to search
-                for (int index = 0; index < searchColumnCount; index++) {
+                for (int index = 0; index < searchColumnCount; ++index) {
```
pbek: Here too!

src/utils/gui.cpp
Waqar144: Really? :thinking: 

Okay done.

src/utils/gui.cpp
Waqar144: done

src/utils/gui.cpp
pbek: Maybe I'm wrong in this case. I'm not 100% sure, but why are you changing those anyway? :grimacing: :grin:

Waqar144: They are not so important. The compiler can optimize them but in general it is preferred to use the pre increment because its faster

PR #1507, opened 2019-12-15
Revert back to old syntax to fix debian 8 build

Waqar144:


pbek:
Thank you! We'll see if there is more in the next release. 😁

Waqar144:
> Thank you! We'll see if there is more in the next release. grin

Though I highly urge you to rethink debian 8 support now. It's going EOL soon. :grin:

PR #1508, opened 2019-12-15
Fix pkg-config name for the botan library.

travankor:


pbek:
@travankor, what is this change fixing?
@Waqar144, will that make any troubles with Botan?

travankor:
On my system, the pkgconfig for botan is located at `/usr/lib/pkgconfig/botan-2.pc`. Note that the botan library is `/usr/lib/libbotan-2.so` and it symlinks to `/usr/lib/libbotan-2.so.12.12.1`. The minor version of the library is unnecessary and confuses pkg-config.

I can add this info to the git commit message if needed.

Also, something that might be useful to others would be to allow users to set `USE_SYSTEM_BOTAN` directly from qmake instead of editing the file manually (or with `sed`), but I wasn't sure how to proceed on that one and if you guys actually wanted to do something like that.

travankor:
I should also say clearly that `PKGCONFIG += botan2.12` fails to build.

Waqar144:
> @Waqar144, will that make any troubles with Botan?

Nah. It's better this way if someone wants to use the system provided botan. Should compile without worries.



> I should also say clearly that `PKGCONFIG += botan2.12` fails to build.

Yes, no wonder :laughing:

Thanks @travankor :+1:

pbek:
Great, thank you 😁

PR #1554, opened 2020-01-07
Add Two Dark color scheme

ScottKillen:
This is the two dark color scheme, based on the VS Code's One Dark theme.

2020-01-07 08_44_42-Welcome to QOwnNotes - QOwnNotes 20 1 3


Waqar144:
This will replace the existing `Dark` theme.

src/configurations/schemes.conf
```diff
@@ -1585,3 +1585,147 @@ FallbackSchema=EditorColorSchema-cdbf28fc-1ddc-4d13-bb21-6a4043316a2f
 Underline_0=true
 Underline_18=true
 Underline_26=true
+
+[EditorColorSchema-cdbf28fc-1ddc-4d13-bb21-6a4043316a2f]
```
Waqar144: This will replace the existing Dark theme.

ScottKillen: Sorry!

I used the UUID produced by QOWNNotes and didn't check for conflicts. I did begin with a copy of the Dark theme.

ScottKillen: Resolved in e638c93a6554b3b38e0509dc5db88d10808d2cca

pbek:
Great. :grin:  `FallbackSchema` is also missing, I'll add that.

pbek:
I cannot see any difference to the `Dark` theme (even the new headline color variations are the same). Are you sure you added the right schema?

pbek:
I'll revert for the next release.

Issue #1559, opened 2020-01-09
Numeric literal preview problem in code block


samael205:
#### Expected behaviour

The same text on the source Markdown and the rendered markdown pannel

#### Actual behaviour

Hi, copying and pasting CSS code, QOwnnotes Do that:

![image](https://user-images.githubusercontent.com/649740/72065883-8ba28080-32df-11ea-864b-8f86fada7926.png)

#### Steps to reproduce

Paste this:

Botones
=======

```
.boton2 
    {
        background: #898989;
        border:0px;
        color:White;
        border-radius: 3px;
        padding: 6px 20px;
        font-size: 14px;
    }
    .boton2:focus 
    {
        outline:0px;
    }
    .boton2:hover 
    {
        background: #6E6E6E;   
    }
```

#### Output from the debug section in the settings dialog

<details><summary>Expand</summary>

</details>

#### Relevant log output in the Log panel

<details><summary>Expand</summary>

</details>

pbek:
Funny one. Thank you for reporting.

@Waqar144, a preview bug again. 😁

pbek:
## 20.1.6
- fixed a numeric literal code block highlighting problem (for [#1559](https://github.com/pbek/QOwnNotes/issues/1559), thank you @Waqar144)

Waqar144:
You can also use ` ```css ` style code blocks for better highlighting 😉

pbek:
@Waqar144, now it looks like this::

![image](https://user-images.githubusercontent.com/1798101/72153452-4c8f3080-33ae-11ea-948e-a83608df6300.png)

    
```
    .boton2 
    {
        background: #898989;
        border:0px;
        color:White;
        border-radius: 3px;
        padding: 6px 20px;
        font-size: 14px;
    }
    .boton2:focus 
    {
        outline:0px;
    }
    .boton2:hover 
    {
        background: #6E6E6E;   
    }
```

Waqar144:
fixed

pbek:
> fixed

I think your PR is for the wrong branch.

pbek:
There now is a new release, could you please test it and report if it works for you?

PR #1564, opened 2020-01-10
Fix highlighting for unsupported langs

Waqar144:
#1563
#1559


pbek:
In this PR you are undoing the md4c support in the md4c branch. :laughing:

Waqar144:
:grimacing:

PR #1565, opened 2020-01-10
Fix CMake build

Waqar144:


pbek:
Works for me, thank you.

PR #1591, opened 2020-01-16
Fix build for older qt versions

Waqar144:
Lets hope that was the only error. :laughing:


pbek:
:crossed_fingers:

pbek:
worked! thank you!

PR #1594, opened 2020-01-18
Fix memory leak in preview

Waqar144:


pbek:
Much better, but QON is still leaking if markdown highlighting is turned on.

PR #1600, opened 2020-01-21
Fix crash

Waqar144:


src/helpers/qownnotesmarkdownhighlighter.cpp
```diff
@@ -110,7 +110,7 @@ void QOwnNotesMarkdownHighlighter::highlightMarkdown(const QString& text) {
                             text.startsWith(QLatin1String("```")) ||
                             text.startsWith(QLatin1String("~~~"));
     const QString &next = currentBlock().next().text();
-    const bool isHeading = text.at(0) == QChar('#');
+    const bool isHeading = !text.isEmpty() && text.at(0) == QChar('#');
```
pbek: is it really `text` and not `next`?

pbek: ah, yes. `text` seems ok :grin:

pbek:
yes, fixed

src/helpers/qownnotesmarkdownhighlighter.cpp
Waqar144: ;)

Issue #1558, opened 2020-01-08
Markdown code blocks are no longer formatted


smokris:
#### Expected behaviour
In 20.1.2 and prior, Markdown code blocks (4 spaces indented) are properly formatted (font and colors are applied):
![20 1 2](https://user-images.githubusercontent.com/190164/72001897-5da43a00-3214-11ea-8197-66d6fcc4bb23.png)

#### Actual behaviour
In 20.1.3 and 20.1.5, the same text is no longer formatted:
![20 1 3 and 20 1 5](https://user-images.githubusercontent.com/190164/72001914-68f76580-3214-11ea-8623-934dfd124dd2.png)

#### Steps to reproduce
Create a note with 4-space-indented text.

pbek:
Confirmed. Thank you for reporting.

@Waqar144, another regression?

pbek:
## 20.1.6
- fixed four spaces code blocks (for [#1558](https://github.com/pbek/QOwnNotes/issues/1558), thank you @Waqar144)

pbek:
Btw. it would be much better to use the code blocks with "ticks", then you can also profit from code block syntax highlighting.

pbek:
There now is a new release, could you please test it and report if it works for you?

smokris:
Yes, I confirmed that in 20.1.6, four-space code blocks are formatted again.  Thanks, @pbek and @Waqar144!

> Btw. it would be much better to use the code blocks with "ticks", then you can also profit from code block syntax highlighting

The syntax highlighting looks nice!  I've started using it for some code blocks.  (Though for very short snippets, like in my screenshot above, I prefer the more concise four-space syntax.)

pbek:
Great, thank you for testing!

smokris:
I just updated to 20.1.13, and four-space code blocks are broken again.  😇

Waqar144:
Can you share a screenshot?

smokris:
(It's exactly the same as the screenshots at the top of this case.)

Waqar144:
Ah yes,
The Library got updated, QON didn't. I guess it will be working in the next release.

smokris:
Sounds good.  Thanks.

pbek:
## 20.1.14
- fixed four spaces code blocks again (for [#1558](https://github.com/pbek/QOwnNotes/issues/1558), thank you @Waqar144)

There now is a new release, could you please test it and report if it works for you?

smokris:
[...]

PR #1601, opened 2020-01-22
Use parent class highlightMarkdown method

Waqar144:
This is redundant, I think its better to use the parent class method directly.

### DIFF
--- a/src/helpers/qownnotesmarkdownhighlighter.cpp
+++ b/src/helpers/qownnotesmarkdownhighlighter.cpp
@@ -87,12 +87,19 @@ void QOwnNotesMarkdownHighlighter::setCodeHighlighting(bool state)
  * @param text
  */
 void QOwnNotesMarkdownHighlighter::highlightBlock(const QString &text) {
+    if (currentBlockState() == HeadlineEnd) {
+        currentBlock().previous().setUserState(NoState);
+        addDirtyBlock(currentBlock().previous());
+    }
     setCurrentBlockState(HighlighterState::NoState);
     currentBlock().setUserState(HighlighterState::NoState);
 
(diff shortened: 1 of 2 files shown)


pbek:
Does this also already fix #1558?

Waqar144:
> Does this also already fix #1558?

The library still needs to be updated. But with this change we don't have to worry about updating QON everytime the lib changes

pbek:
I updated the library. Highlighting works now. One interesting aspect with dark schema (light schema seems ok). Strings have no background color. :grin: In the fenced code blocks they have the background color if the code block.

image

    <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
    <verzeichnis>
         <titel>Wikipedia Städteverzeichnis</titel>
         <eintrag>
              <stichwort>Genf</stichwort>
              <eintragstext>Genf ist der Sitz von ...</eintragstext>
         </eintrag>
         <eintrag>
              <stichwort>Köln</stichwort>
              <eintragstext>Köln ist eine Stadt, die ...</eintragstext>
         </eintrag>
    </verzeichnis>

Waqar144:
Will fix that.
However it can kind of fixed if you turn of code block bg color off ;)

Waqar144:
We can't draw rectangles on this. Two solutions:
1. We lose the syntax highlighting in it.
2. Turn off code block background (or alternatively turn on background for all elements of type code)

pbek:
I vote for 1) since the type can't be specified and so there is no context. `"` doesn't always mean "string"

Waqar144:
Alright, will do that.

Waqar144:
Try now

pbek:
Works great, thank you!

PR #1603, opened 2020-01-23
New script function noteTextEditSelectCurrentWord()

jfx2006:
Selects the current word in the editor. It can then be read with
noteTextEditSelectedText(), processed and replaced with
noteTextEditWrite("New Text").

Useful when writing weekly status reports.


pbek:
Looking good, thank you!

PR #1608, opened 2020-01-27
Tests and improvements for Utils::Misc

Waqar144:


src/utils/misc.cpp
```diff
@@ -961,13 +964,13 @@ int Utils::Misc::getDefaultSearchEngineId() {
...
  */
 QList<int> Utils::Misc::getSearchEnginesIds() {
-    QList<int> list;
-    list << SearchEngines::DuckDuckGo << SearchEngines::Google
-         << SearchEngines::Bing << SearchEngines::Yahoo
-         << SearchEngines::GoogleScholar << SearchEngines::Yandex
-         << SearchEngines::AskDotCom << SearchEngines::Qwant
-         << SearchEngines::Startpage;
-    return list;
+    return QList<int> {
```
pbek: Will this work with older Qt versions?

QOwnNotes-Project.pro
```diff
@@ -0,0 +1,13 @@
...
+
+SUBDIRS = \
+    src \
+    tests
+
+src.file = src/QOwnNotes.pro
+tests.file = tests/QOwnNotesTests.pro
+
+test.depends = src
+
```
pbek: Why the name `QOwnNotes-Project.pro`? Wouldn't `QOwnNotes.pro` be better?

Waqar144: Just to differentiate from the inner QOwnNotes.pro file :grimacing:

src/utils/misc.cpp
Waqar144: Yeah totally.

QOwnNotes-Project.pro
Waqar144: Renamed inner files to `src.pro` and `tests.pro`. Outer file is now `QOwnNotes.pro`

pbek: But I still don't like that `QOwnNotes-Project.pro`. Why do we need that again?

Waqar144: Just for utility. Tests and Project are opened as one overall project with two subprojects(app, tests) so you can quickly switch from one to another

pbek: Isn't that almost more confusing than before? Are there other projects who are using one project file for the application and the tests?

Waqar144: Yeah most of them

But I will remove it.

pbek: I'm not against it if it is considered "best practice". :laughing: And that was what I tried to find out.

Waqar144: I don't know if its best practice lol :rofl: , but it does kinda make it easier to run the tests(from Qt creator, i.e., just switch the target)

Waqar144:
We just hit 5000 commits :fireworks:

QOwnNotes-Project.pro
pbek: I just switch project.

pbek:
Let's have a commit party. :tada: :laughing:

Issue #1611, opened 2020-01-28
Crash when showing note versions


pbek:
@Waqar144, I'm getting hard to reproduce crash in QON when showing the note versions of a note in the current release. But it's only happening in release mode and it seems it's happening in the highlighter. 

```
1   Note::getFileURLFromFileName(QString) const                                                            0x5555556ebeb4 
2   Note::fetchByRelativeFileName(QString const&) const                                                    0x5555556ec132 
3   QOwnNotesMarkdownHighlighter::highlightBrokenNotesLink(QString const&)                                 0x5555557401af 
4   QOwnNotesMarkdownHighlighter::highlightBlock(QString const&)                                           0x55555574107f 
5   ??                                                                                                     0x7ffff68e2660 
6   ??                                                                                                     0x7ffff68e2820 
7   QMetaObject::activate(QObject *, int, int, void * *)                                                   0x7ffff4b29805 
8   QTextDocument::contentsChange(int, int, int)                                                           0x7ffff6b550ca 
9   QTextDocumentPrivate::finishEdit()                                                                     0x7ffff689d8a1 
10  ??                                                                                                     0x7ffff7187913 
11  QMarkdownTextEdit::setPlainText(QString const&)                                                        0x5555557f1010 
12  VersionDialog::on_versionListWidget_currentRowChanged(int)                                             0x5555556d15df 
13  VersionDialog::qt_metacall(QMetaObject::Call, int, void * *)                                           0x55555594bbae 
14  QMetaObject::activate(QObject *, int, int, void * *)                                                   0x7ffff4b298c9 
15  QListWidget::currentRowChanged(int)                                                                    0x7ffff7243b71 
16  ??                                                                                                     0x7ffff7244662 
17  QMetaObject::activate(QObject *, int, int, void * *)                                                   0x7ffff4b29805 
18  QItemSelectionModel::currentChanged(QModelIndex const&, QModelIndex const&)                            0x7ffff4aa212a 
19  QItemSelectionModel::setCurrentIndex(QModelIndex const&, QFlags<QItemSelectionModel::SelectionFlag>)   0x7ffff4aa23b6 
20  QListWidget::setCurrentRow(int)                                                                        0x7ffff7244aa6 
... <More>
```

[...]

PR #1612, opened 2020-01-28
Attempt fix crash when checking note versions

Waqar144:
#1611


pbek:
Yes, it's fixed now! Great, thank you!

PR #1628, opened 2020-02-14
Minor optimizations for QOwnNotes

Waqar144:


pbek:
@Waqar144 this seems to break a lot of builds on https://build.opensuse.org/package/show/home:pbek:QOwnNotes/desktop

Waqar144:
> @Waqar144 this seems to break a lot of builds on https://build.opensuse.org/package/show/home:pbek:QOwnNotes/desktop

Seems like we need a separate code path for `constFirst() ` for older Qt versions

pbek:
Qt 5.6, I see :grimacing:

PR #1636, opened 2020-02-19
Switch to QVector for Note

Waqar144:


pbek:
Failed:

- https://ci.appveyor.com/project/pbek/qownnotes/builds/30912203/job/uc1k67lvp2nqmu1r
- https://travis-ci.org/pbek/QOwnNotes/jobs/652548476

pbek:
What are the advantages (and disadvantages?) from QVector over QList?

Waqar144:
> What are the advantages (and disadvantages?) from QVector over QList?

This link in the docs explains it the best:
https://doc.qt.io/qt-5/qlist.html#details

and this one goes in a lot of detail:
https://marcmutz.wordpress.com/effective-qt/containers/#containers-qlist

Some benchmarks by KDAB:
https://www.kdab.com/~marc/effective-qt/containers/results-append-tickcounter-linux_x86-64_473_gcc43-zoom-qt.html

Pros
----

Efficient when prepending / inserting an item

Cons
---

- Allocates on the heap if the size of the object is ` > sizeof(void*)` which would be 8 bytes usually. Items don't get strored sequentially hence the loss in performance and unnecessary heap allocations.

pbek:
Thank you!

pbek:
OBS is breaking apart again...
e.g.: https://build.opensuse.org/package/live_build_log/home:pbek:QOwnNotes/desktop/Debian_8.0/i586

Waqar144:
I fear this doesn't have a simple straightforward solution.

pbek:
What does that mean? Reverting the pull request?

pbek:
Because that would be pretty sad. 😞

pbek:
https://doc.qt.io/qt-5/qvector.html#append-1 was added in Qt 5.6, but https://doc.qt.io/qt-5/qvector.html#append should be in since before...

pbek:
Ah, it's rather a https://doc.qt.io/qt-5/qvector.html#append-2 which came with Qt 5.5

pbek:
I guess in the case of `noteIdList.append(searchInNotes(QChar('<') + altLinkText + QChar('>'), true));` that appends a QVector to a QVector there could a fallback be written for Qt < 5.5 where the items are appended one by one.

Waqar144:
> Ah, it's rather a https://doc.qt.io/qt-5/qvector.html#append-2 which came with Qt 5.5

I don't understand why they didn't put all those methods just once instead of sprinkling them over time in different versions.

I think std::copy might be able to fix this.

Or we can just use the good old std::vector and be happy forever(it is the fastest ;)) 😅🤷‍♂️

Or maybe write our own dynamic array to rule them all 😂

pbek:
> I think std::copy might be able to fix this.

But there is nothing below Qt 5.5 that can add a vector to a vector, isn't it? I currently attempting a `Utils::Misc::appendVectorToVector` to fix that.

Waqar144:
I think `<<` will work. I can see it in the Qt 4.8 docs.

pbek:
Oh, really? Sounds like: https://doc.qt.io/qt-5/qvector.html#operator-lt-lt-1

Meanwhile https://github.com/pbek/QOwnNotes/commit/a9924c009c0fdaeb06c1f22e37066c64b6876538 will do...

Waqar144:
> Oh, really? Sounds like: https://doc.qt.io/qt-5/qvector.html#operator-lt-lt-1
> 
> Meanwhile [a9924c0](https://github.com/pbek/QOwnNotes/commit/a9924c009c0fdaeb06c1f22e37066c64b6876538) will do...

Yes exactly. I think this is exactly == `append(QVector &)`, though the latter didn't become available until after Qt 5.3

PR #1651, opened 2020-03-08
Update desktop-qt5 support for the snap

pachulo:
Following the tips from here:

- https://forum.snapcraft.io/t/desktop-app-support-qt5/11703

I've updated the `snapcraft.yaml` so the snap looks better integrated with the system.

### DIFF
--- a/build-systems/snap/snapcraft/snapcraft.yaml
+++ b/build-systems/snap/snapcraft/snapcraft.yaml
@@ -15,6 +15,7 @@ apps:
     command: desktop-launch ${SNAP}/usr/bin/QOwnNotes -style=Fusion --snap
     environment:
         DISABLE_WAYLAND: 1
+        QT_QPA_PLATFORMTHEME: gtk3
     # see https://docs.snapcraft.io/reference/interfaces
     plugs:
       - x11
@@ -27,6 +28,20 @@ apps:
       - removable-media
       - opengl
       - cups-control
+
+plugs: # plugs for theming, font settings, cursor and to use gtk3 file chooser
+  gtk-3-themes:
+    interface: content
+    target: $SNAP/data-dir/themes
+    default-provider: gtk-common-themes:gtk-3-themes
+  icon-themes:
+    interface: content
+    target: $SNAP/data-dir/icons
+    default-provider: gtk-common-themes:icon-themes
+  sound-themes:
+    interface: content
+    target: $SNAP/data-dir/sounds
+    default-provider: gtk-common-themes:sounds-themes
 
 parts:
   qownnotes:
@@ -66,6 +81,7 @@ parts:
       plugin: make
       make-parameters: ["FLAVOR=qt5"]
       build-packages:
+        - build-essential
         - qtbase5-dev
         - dpkg-dev
       stage-packages:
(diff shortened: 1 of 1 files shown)


build-systems/snap/snapcraft/snapcraft.yaml
```diff
@@ -15,6 +15,7 @@ apps:
     command: desktop-launch ${SNAP}/usr/bin/QOwnNotes -style=Fusion --snap
     environment:
         DISABLE_WAYLAND: 1
+        QT_QPA_PLATFORMTHEME: gtk3
```
pbek: Thank you, but is there something that would enable some KDE look and feel instead of a GTK look and feel? 😅

pachulo: I don't know of any specific configuration for that, but as I understand it, this should enable the application to look more «at home» on any DE.
I don't use KDE, but if you do you can test some snaps with these changes already applied, like:
- https://snapcraft.io/keepassxc
- https://snapcraft.io/musescore

To see how they look like in there.

pbek: I've keepassxc installed and all the dialogs are GTK dialogs, despite that it's a Qt app and I'm running KDE. I don't like that as a default-experience a lot. :grinning: 

They also use `QT_QPA_PLATFORMTHEME=gtk3` in <https://github.com/keepassxreboot/keepassxc/blob/develop/snap/local/launchers/gtk3-env-launch>

I think QOwnNotes currently is using the "Fusion" theme in snaps.
I almost would prefer different snaps for KDE and Gnome, like `qownnotes-kde` and `qownnotes-gtk`.
What's your opinion?

pachulo: > I almost would prefer different snaps for KDE and Gnome, like qownnotes-kde and qownnotes-gtk.
>  What's your opinion?

I haven't seen any snap doing that, so I don't really thought about it till now...I'm not really sure it's a good idea.

> Or is there another option to incorporate different look and feels?

Not that I know, but that is a question better asked on the forum: https://forum.snapcraft.io/ do you mind opening a tread?

pbek: Phew, there are a lot of snaps out there using many integrations. That topic needs more research.

pbek: Nevertheless I asked https://forum.snapcraft.io/t/kde-and-gtk-integration-for-qt-app/15892

pbek:
I tested your PR and got weird font problems in the editor (also see log on the left).

image

I wasn't even able to resize the font size with `Ctrl + Plus` and in the settings dialog the default font size for the editor wasn't set (and thus no size could be set with `Ctrl + Plus`).

image

Do you also have a similar problem?

pbek:
I'm trying to work around the problem, but I still have the same issue. Don't you also have it?

pbek:
After many hours of testing and building I settled for an "in between" thing between Breeze and GKT...

pbek:
The Breeze styling will now be used by default along with the GTK file selectors.

pachulo:
Sorry @pbek , I didn't have time to test anything this week...

I can see that the last version of the snap looks quite different now (and much better if you ask me)! Thanks for improving this!

pbek:
Thank you for testing, lets roll with them then. In theory I could also check for KDE env variables to decide which style I use.

PR #1660, opened 2020-03-15
Updating the location of localization files

DeadMozay:
It seems that I didn’t forget anything, I checked appimage and openSUSE, localization works, as in other distributions, packaging also went to fixed directories, I think there should be no problems
https://build.opensuse.org/package/show/home:Dead_Mozay:branches:home:pbek:QOwnNotes/desktop
building everywhere was successful except Arch, but I think this is due to some kind of error on the obs


obs/qownnotes.spec
```diff
@@ -206,98 +142,15 @@ rm -rf $RPM_BUILD_ROOT
 %defattr(-,root,root)
 %doc LICENSE README.md CHANGELOG.md shortcuts.md
 %{_bindir}/QOwnNotes
-%{_datadir}/pixmaps/QOwnNotes.png
```
pbek: Can that line really go?

obs/qownnotes.spec
```diff
@@ -114,88 +123,15 @@ popd
...
-install -D -m 0644 images/icons/16x16/apps/QOwnNotes.png $RPM_BUILD_ROOT/%{_datadir}/icons/hicolor/16x16/apps/QOwnNotes.png
-install -D -m 0644 images/icons/24x24/apps/QOwnNotes.png $RPM_BUILD_ROOT/%{_datadir}/icons/hicolor/24x24/apps/QOwnNotes.png
-install -D -m 0644 images/icons/32x32/apps/QOwnNotes.png $RPM_BUILD_ROOT/%{_datadir}/icons/hicolor/32x32/apps/QOwnNotes.png
-install -D -m 0644 images/icons/48x48/apps/QOwnNotes.png $RPM_BUILD_ROOT/%{_datadir}/icons/hicolor/48x48/apps/QOwnNotes.png
-install -D -m 0644 images/icons/64x64/apps/QOwnNotes.png $RPM_BUILD_ROOT/%{_datadir}/icons/hicolor/64x64/apps/QOwnNotes.png
-install -D -m 0644 images/icons/96x96/apps/QOwnNotes.png $RPM_BUILD_ROOT/%{_datadir}/icons/hicolor/96x96/apps/QOwnNotes.png
-install -D -m 0644 images/icons/128x128/apps/QOwnNotes.png $RPM_BUILD_ROOT/%{_datadir}/icons/hicolor/128x128/apps/QOwnNotes.png
-install -D -m 0644 images/icons/256x256/apps/QOwnNotes.png $RPM_BUILD_ROOT/%{_datadir}/icons/hicolor/256x256/apps/QOwnNotes.png
-install -D -m 0644 images/icons/512x512/apps/QOwnNotes.png $RPM_BUILD_ROOT/%{_datadir}/icons/hicolor/512x512/apps/QOwnNotes.png
```
pbek: How are those lines substituted?

pbek:
Looks pretty complete (beside my comments).
It will be hard to test all that beforehand... How many distros have you actually tested?

Can you please attempt to build the Arch Packages again, maybe OBS fixed the build already.

And the application will check `/usr/share/qt5/translations/` without adding it to `main.cpp`?

obs/qownnotes.spec
DeadMozay: The icon itself will be found

obs/qownnotes.spec
DeadMozay: these files are automatically installed using a macro `%make_install`

DeadMozay:
I checked the finished package for arch before the building broke, at first glance the packaging was completed as expected, but I will check again.
without editing `main.cpp`, the application will look for translations not where they actually lie, it will search for them along the old path

pbek:
> it will search for them along the old path

Which old path do you mean? So do we need to include `/usr/share/qt5/translations/` as path to look in?

DeadMozay:
> > it will search for them along the old path
> 
> Which old path do you mean? So do we need to include `/usr/share/qt5/translations/` as path to look in?

Checked on openSUSE, Fedora, Arch, Ubuntu, Appimage, everything works

pbek:
Thank you, we'll wait for the feedback from the next release then. 😬

PR #1688, opened 2020-04-13
Fix QSet deprecation warnings

Waqar144:


src/mainwindow.cpp
```diff
@@ -3298,18 +3300,19 @@ bool MainWindow::buildNotesIndex(int noteSubFolderId, bool forceRebuild) {
...
         // check for removed note subfolders
-//#if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
+        QList<int> removedNoteSubFolderIdList =
+            QSet<int>(_buildNotesIndexBeforeNoteSubFolderIdList.begin(),
+                      _buildNotesIndexBeforeNoteSubFolderIdList.end())
+                .subtract(QSet<int>(_buildNotesIndexAfterNoteSubFolderIdList.begin(),
+                          _buildNotesIndexAfterNoteSubFolderIdList.end())).values();
+#else
 //        const QList<int> removedNoteSubFolderIdList =
```
pbek, line 3310: The "else" part is: Variable `removedNoteSubFolderIdList` will not be set at all? :laughing:

Waqar144: ah yes. Done :p

pbek:
I hope you tested this. This is the most vital part of the application. :fearful:

pbek:
Seems ok, so thank you. I'll keep my fingers crossed :crossed_fingers: :grin:

Waqar144:
> I hope you tested this. This is the most vital part of the application. fearful

Dont worry. It will work. And if for some unexplainable reason it doesn't then the future of Qt is grim lol

PR #1689, opened 2020-04-13
Make db non const when querying

Waqar144:


pbek:
Doesn't it work with const?

Waqar144:
> Doesn't it work with const?

It does, but there is a hint of problem we had, so how about revert constness just for this?

pbek:
You mean https://github.com/pbek/QOwnNotes/commit/00f16e3e998e1102c8103e51fd4e5456280f4af5? No, that was an actual warning I got under Windows, because the DB connection was closed on a const DB. You didn't "see" that because you only work on Linux. :grin: 

I don't think we need this PR unless there is an actual problem.

pbek:
I'll close this and keep it until there is an actual issue.

Waqar144:
> You mean [00f16e3](https://github.com/pbek/QOwnNotes/commit/00f16e3e998e1102c8103e51fd4e5456280f4af5)? No, that was an actual warning I got under Windows, because the DB connection was closed on a const DB. You didn't "see" that because you only work on Linux. grin
> 
> I don't think we need this PR unless there is an actual problem.

Alright.

PR #1714, opened 2020-04-27
Refactor app properties setting into separate func

Waqar144:


pbek:
Almost the same I wanted to do. 😁 

Interesting that (only) the macOS build breaks:

```
main.cpp:562:31: error: use of undeclared identifier 'translators'

        loadTranslations(app, translators, locale);

                              ^

main.cpp:606:31: error: use of undeclared identifier 'translators'

        loadTranslations(app, translators, locale);
```

btw. you can also switch the template comments away from the old "Macro" text 😉

PR #1728, opened 2020-05-03
Caja: open folder fix

r00tr4v3n:


src/utils/misc.cpp
```diff
@@ -155,9 +155,11 @@ void Utils::Misc::openFolderSelect(const QString &absolutePath) {
                 QStringLiteral("nautilus"),
                 QStringList{"--no-desktop", QDir::toNativeSeparators(path)});
         } else if (output == QStringLiteral("caja-folder-handler.desktop")) {
+            QDir dir = QFileInfo(path).absoluteDir();
+            QString absDir=dir.absolutePath();
```
Waqar144, line 159: Thanks for the patch.
Can you make these both `const`?
and fix the formatting on second line?

r00tr4v3n: Like this?

pbek:
Thank you for the fix.

PR #1731, opened 2020-05-06
Fix crash due to QString overlapping count

Waqar144:
This will fix the crash,

the commit in MD4C still needs to be reverted.

The preview of the note is still incorrect so i will investigate further as to what is causing it. My guess is the CodeHighlighter and MD4C don't agree on something.

If you increase/decrease the indentation in the second para of the test note, everything becomes visible.

### DIFF
--- a/src/entities/note.cpp
+++ b/src/entities/note.cpp
@@ -2053,6 +2053,18 @@ void highlightCode(QString &str, const QString &type, int cbCount) {
     }
 }
 
+static int nonOverlapCount(const QString &str, const QChar c = '`') {
+  const auto len = str.length();
+  int count = 0;
+  for (int i = 0; i < len; ++i) {
+    if (str[i] == c && i + 2 < len && str[i + 1] == c && str[i + 2] == c) {
+      ++count;
+      i += 2;
+    }
+  }
+  return count;
+}
+
 /**
  * Converts a markdown string for a note to html
  *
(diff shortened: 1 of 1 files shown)


src/entities/note.cpp
```diff
@@ -2053,6 +2053,18 @@ void highlightCode(QString &str, const QString &type, int cbCount) {
     }
 }
 
+static int nonOverlapCount(const QString &str, const QChar c = '`') {
```
pbek, line 2056: Did you put this method *not* into the header file on purpose?

Waqar144: yes, its not being used anywhere, so i wanted to limit this to note.cpp only

Waqar144: does this fix the crash on your side?

pbek: will test  when I have access to my Manjaro box

pbek:
> the commit in MD4C still needs to be reverted.

my "fix"? I already did, didn't that went that through?

Waqar144:
> > the commit in MD4C still needs to be reverted.
> 
> my "fix"? I already did, didn't that went that through?

ah yes, its gone

src/entities/note.cpp
pbek: Crash gone on Manjaro, great job!

PR #1738, opened 2020-05-10
file management for scripts, script menu

r00tr4v3n:
[https://github.com/pbek/QOwnNotes/issues/1737#issuecomment-626306054](https://github.com/pbek/QOwnNotes/issues/1737#issuecomment-626306054)

PR #1739, opened 2020-05-10
file management docs

r00tr4v3n:


docs/content/scripting/methods-and-objects.md
```diff
@@ -1328,6 +1328,48 @@ Opening an input dialog with a line edit
     script.log(result);
     ```
 
+Checking for a file
```
pbek, line 1331: I guess `Checking if file exists` would be a better description. :grin:

pbek:
I guess this PR should be part of #1738, right? But never mind that now. :grin:

docs/content/scripting/methods-and-objects.md
pbek: I will update it...

PR #1749, opened 2020-05-17
Minor improvements to Note

Waqar144:


pbek:
Thank you!

src/entities/note.cpp
```diff
@@ -1229,10 +1215,11 @@ bool Note::storeNoteTextFileToDisk() {
     }
 
     const bool fileExists = this->fileExists();
+    const QString newName = _name;
 
     // assign the tags to the new name if the name has changed
     if (oldName != newName) {
-        if (!noteFileWasRenamed && TrashItem::isLocalTrashEnabled()) {
```
pbek, line 1235: Why did you remove this check?

src/entities/note.cpp
```diff
@@ -1272,46 +1259,41 @@ bool Note::storeNoteTextFileToDisk() {
...
+    QFile oldFile(oldNoteFilePath);
+    const QFileInfo oldFileInfo(oldFile);
+    const QFile newFile(fullNoteFilePath());
+    const QFileInfo newFileInfo(newFile);
+
+    // in the end we want to remove the old note file if note was stored and
+    // filename has changed
+    // #1190: we also need to check if the files are the same even if the name
+    // is not the same for NTFS
+    if (noteStored && (fullNoteFilePath() != oldNoteFilePath) &&
```
pbek, line 1279: Does those changes still do the same? :grimacing:

src/entities/note.cpp
```diff
@@ -321,42 +317,42 @@ bool Note::copyToPath(const QString &destinationPath, QString noteFolderPath) {
...
         const bool isFileCopied = file.copy(destinationFileName);
+        if (!isFileCopied) {
+            return false;
+        }
 
-        if (isFileCopied) {
-            const QStringList mediaFileList = getMediaFileList();
+        const QStringList mediaFileList = getMediaFileList();
+        if (mediaFileList.empty()) {
+            return false;
```
pbek, line 326: I already fixed those bad returns.

src/entities/note.cpp
pbek: Please don't slip in such big rewrites with minor variable changes. These should better go into a different pull requests where you explain what you did. :sweat_smile: the same goes for the `copyToPath` rewrite, that caused breaking of coping and moving of notes (#1764)

src/entities/note.cpp
pbek: See #1764

src/entities/note.cpp
Waqar144: Its weird but i dont remember making this change

pbek: Now I'm even more scared! :laughing:

pbek: You removed the whole block where the variable was declared.

src/entities/note.cpp
Waqar144: I dont remember making these changes. Maybe this happened because of the renaming

src/entities/note.cpp
Waqar144: I think it happened due to the renaming. very specific things are moved.

src/entities/note.cpp
pbek: Large chunks of code were rewritten or restructured (not only that one line).

Waqar144: Its best that i restore these

src/entities/note.cpp
pbek: Renaming of variables? What did you use for refactoring that?

Waqar144: qt create :eyes:

pbek: Why would it move blocks of code around when you rename a variable? :grimacing:

pbek: Maybe that git did that all. e.g. I did https://github.com/pbek/QOwnNotes/commit/48806768f3d4d31fcae4df8fb01c0d6f4a56b85c#diff-69560e71be6fc4483e10036d29839d66R1235, maybe your PR ripped that all out again

src/entities/note.cpp
pbek: These changes were yours, @Waqar144 and not cause by git, right?

Issue #1746, opened 2020-05-14
[Suggestion] Script detachedProcess result


r00tr4v3n:
Scripts like the uml-renderer are great but unusable at the moment because the ui freezes for multiple seconds if you generate more than one image. (I had multiple graphs in one file and QON froze for a minute upon starting.) Caching like in the latex-math script helps but isn't good enough. The same goes for a refresh button which would just freeze as well.
The shortest solution might be a [workerscript](https://doc.qt.io/qt-5/qml-qtqml-workerscript-workerscript.html#sendMessage-method) but I was not able to get it to run with QON.

I'd like to add a script thread with a callback hook like for context menus to allow async/detached processes with results.
e. g. 
```cpp
startDetachedProcess(QString executablePath, QStringList parameters, QString callBackIdentifier);
    ………
    function onCallback(identifier, resultSet) {
       if(identifier == "genImages"){
           script.regenerateNotePreview();
       }
    }
```

pbek:
What will be started detached? An external command? Or a function from a qml? Who starts the thread? Can you please make an example of the flow.

r00tr4v3n:
I don't think a qml script itself would create a heavy load, rather external calls. Might be useful though for e. g. http fetcher/parser.

This was the entire flow on the script side.

```js
startDetachedProcess("/bin/latex", {"x^2", "/tmp/out.png"}, "callback-latex");
```
```cpp
// Qt
// instead of starting the process detached, create a new thread 
// and start it synchronously to retrieve the result
callCallbackInvokedForObject(scriptComponent.object, identifier);
```

```js
    function onCallback(identifier, resultSet) {
       if(identifier == "callback-latex"){
           script.regenerateNotePreview();
       }
    }
```

pbek:
So you want that `startDetachedProcess` uses `Utils::Misc::startSynchronousProcess` in a new thread and return the output to the scripting engine via `callBackIdentifier` if `callBackIdentifier` is set?

How would that callback with `callBackIdentifier` know which process ended if multiple of with the same `callBackIdentifier` were started? I think the pid wouldn't be very useful...

r00tr4v3n:
> So you want that `startDetachedProcess` uses `Utils::Misc::startSynchronousProcess` in a new thread and return the output to the scripting engine via `callBackIdentifier` if `callBackIdentifier` is set?

Yes.

[...]

PR #1750, opened 2020-05-17
Script multithreading

r00tr4v3n:
[https://github.com/pbek/QOwnNotes/issues/1746](https://github.com/pbek/QOwnNotes/issues/1746)

### DIFF
--- a/docs/content/scripting/examples/callback.qml
+++ b/docs/content/scripting/examples/callback.qml
@@ -31,7 +31,7 @@ QtObject {
     * @param {QVariantList} cmd - the entire command array [0-executablePath, 1-parameters, 2-exitCode]
     * @param {QVariantList} thread - the thread information array [0-passed index, 1-remaining threads for this identifier]
     */
-    function onCallback(identifier, resultSet, cmd, thread){
+    function onDetachedProcessCallback(identifier, resultSet, cmd, thread){
       if(identifier == "callback-example"){
         log(`#${thread[1]} i[${thread[0]}] t${cmd[1]}`);
       }
--- a/docs/content/scripting/examples/callback.qml
+++ b/docs/content/scripting/examples/callback.qml
@@ -9,15 +9,15 @@ import com.qownnotes.noteapi 1.0
 Script {
 
     function log(txt) {
-     script.log("[callback-example] "+txt);
+        script.log("[callback-example] " + txt);
     }
 
     function init() {
-      log("init")
-      for (var i = 0; i < 100; i++) {
-        var dur = Math.floor(Math.random() * 10) + 1;
-        script.startDetachedProcess("sleep", [`${dur}s`], "callback-example", i);
-      }
+        log("init")
+        for (var i = 0; i < 100; i++) {
+            var dur = Math.floor(Math.random() * 10) + 1;
+            script.startDetachedProcess("sleep", [`${dur}s`], "callback-example", i);
+        }
     }
 
     /**
--- a/docs/content/scripting/examples/callback.qml
+++ b/docs/content/scripting/examples/callback.qml
@@ -1,0 +1,39 @@
+import QtQml 2.0
+import com.qownnotes.noteapi 1.0
+
+/**
+ * This script is just an example of how to use scripts
+ * Visit http://docs.qownnotes.org/ for more information about scripting
+ */
+QtObject {
+    /**
+     * Just show some log entries when the script is initialized
+     */
+
+    function log(txt){
+     script.log("[callback-example] "+txt);
+    }
+
+    function init() {
+      log("init")
+      for (var i = 0; i < 100; i++) {
+        var dur = Math.floor(Math.random() * Math.floor(10))+1;
+        script.startDetachedProcess("sleep", [`${dur}s`], "callback-example", i);
+      }
+    }
+
+    /**
+    * This function is called when a script thread is done executing.
+    * Hint: thread[1]==0 helps to determine if a bulk of started processes for a certain identifier is done.
+    *
+    * @param {QString} identifier - the provided id when calling startDetachedProcess()
+    * @param {QString} resultSet - the result of the process
+    * @param {QVariantList} cmd - the entire command array [0-executablePath, 1-parameters, 2-exitCode]
+    * @param {QVariantList} thread - the thread information array [0-passed index, 1-remaining threads for this identifier]
+    */
+    function onCallback(identifier, resultSet, cmd, thread){
+      if(identifier == "callback-example"){
+        log(`#${thread[1]} i[${thread[0]}] t${cmd[1]}`);
+      }
+    }
+}
--- a/docs/content/scripting/examples/callback.qml
+++ b/docs/content/scripting/examples/callback.qml
@@ -2,22 +2,20 @@ import QtQml 2.0
 import com.qownnotes.noteapi 1.0
 
 /**
- * This script is just an example of how to use scripts
+ * This script is a short example of how to use detatched processes and callbacks
+ * to relieve the UI thread.
  * Visit http://docs.qownnotes.org/ for more information about scripting
  */
-QtObject {
-    /**
-     * Just show some log entries when the script is initialized
-     */
+Script {
 
-    function log(txt){
+    function log(txt) {
      script.log("[callback-example] "+txt);
     }
 
     function init() {
       log("init")
       for (var i = 0; i < 100; i++) {
-        var dur = Math.floor(Math.random() * Math.floor(10))+1;
+        var dur = Math.floor(Math.random() * 10) + 1;
         script.startDetachedProcess("sleep", [`${dur}s`], "callback-example", i);
       }
     }
@@ -31,8 +29,8 @@ QtObject {
     * @param {QVariantList} cmd - the entire command array [0-executablePath, 1-parameters, 2-exitCode]
     * @param {QVariantList} thread - the thread information array [0-passed callbackParameter, 1-remaining threads for this identifier]
     */
-    function onDetachedProcessCallback(identifier, resultSet, cmd, thread){
-      if(identifier == "callback-example"){
+    function onDetachedProcessCallback(identifier, resultSet, cmd, thread) {
+      if (identifier == "callback-example") {
         log(`#${thread[1]} i[${thread[0]}] t${cmd[1]}`);
       }
     }
--- a/docs/content/scripting/examples/callback.qml
+++ b/docs/content/scripting/examples/callback.qml
@@ -29,7 +29,7 @@ QtObject {
     * @param {QString} identifier - the provided id when calling startDetachedProcess()
     * @param {QString} resultSet - the result of the process
     * @param {QVariantList} cmd - the entire command array [0-executablePath, 1-parameters, 2-exitCode]
-    * @param {QVariantList} thread - the thread information array [0-passed index, 1-remaining threads for this identifier]
+    * @param {QVariantList} thread - the thread information array [0-passed callbackParameter, 1-remaining threads for this identifier]
     */
     function onDetachedProcessCallback(identifier, resultSet, cmd, thread){
       if(identifier == "callback-example"){
--- a/docs/content/scripting/examples/callback.qml
+++ b/docs/content/scripting/examples/callback.qml
@@ -24,13 +24,13 @@ Script {
     * This function is called when a script thread is done executing.
     * Hint: thread[1]==0 helps to determine if a bulk of started processes for a certain identifier is done.
     *
-    * @param {QString} identifier - the provided id when calling startDetachedProcess()
+    * @param {QString} callBackIdentifier - the provided id when calling startDetachedProcess()
     * @param {QString} resultSet - the result of the process
     * @param {QVariantList} cmd - the entire command array [0-executablePath, 1-parameters, 2-exitCode]
     * @param {QVariantList} thread - the thread information array [0-passed callbackParameter, 1-remaining threads for this identifier]
     */
-    function onDetachedProcessCallback(identifier, resultSet, cmd, thread) {
-      if (identifier == "callback-example") {
+    function onDetachedProcessCallback(callBackIdentifier, resultSet, cmd, thread) {
+      if (callBackIdentifier == "callback-example") {
         log(`#${thread[1]} i[${thread[0]}] t${cmd[1]}`);
       }
     }
--- a/docs/content/scripting/hooks.md
+++ b/docs/content/scripting/hooks.md
@@ -9,12 +9,12 @@ onCallback
      * This function is called when a script thread is done executing.
      * Hint: thread[1]==0 helps to determine if a bulk of started processes for a certain identifier is done.
      *
-     * @param {QString} identifier - the provided id when calling startDetachedProcess()
+     * @param {QString} callbackIdentifier - the provided id when calling startDetachedProcess()
      * @param {QString} resultSet - the result of the process
      * @param {QVariantList} cmd - the entire command array [0-executablePath, 1-parameters, 2-exitCode]
      * @param {QVariantList} thread - the thread information array [0-passed callbackParameter, 1-remaining threads for this identifier]
      */
-    function onCallback(identifier, resultSet, cmd, thread);
+    function onDetachedProcessCallback(callbackIdentifier, resultSet, cmd, thread);
     ```
     You may want to take a look at the example
     [callback-example.qml](https://github.com/pbek/QOwnNotes/blob/develop/docs/content/scripting/examples/callback.qml).
--- a/docs/content/scripting/hooks.md
+++ b/docs/content/scripting/hooks.md
@@ -1,5 +1,24 @@
 # Hooks
 
+onCallback
+------------
+
+!!! help "Method call and parameters"
+    ```js
+    /**
+     * This function is called when a script thread is done executing.
+     * Hint: thread[1]==0 helps to determine if a bulk of started processes for a certain identifier is done.
+     *
+     * @param {QString} identifier - the provided id when calling startDetachedProcess()
+     * @param {QString} resultSet - the result of the process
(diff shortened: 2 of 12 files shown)


src/utils/misc.h
```diff
@@ -19,6 +19,7 @@
 #include <QString>
 #include <QStringList>
 #include <QVector>
+#include <QProcess>
```
Waqar144, line 22: I don't think this is necessary

src/utils/misc.cpp
```diff
@@ -436,6 +436,41 @@ QByteArray Utils::Misc::startSynchronousProcess(const QString &executablePath,
...
+ * @param cmd the terminal command containing all input and output information
+ * @return true on success, false otherwise
+ */
+bool Utils::Misc::startSynchronousResultProcess(TerminalCmd &cmd) {
+    QProcess process;
+
+    // start executablePath synchronous with parameters
+#ifdef Q_OS_MAC
+    process.start("open", QStringList()
+                              << executablePath << "--args" << parameters);
```
Waqar144, line 451: this probably will fail to compile on mac

src/threads/scriptthread.cpp
```diff
@@ -0,0 +1,52 @@
...
+                           const QString &identifier,
+                           const int index) : QThread() {
+    this->cmd = cmd;
+    this->identifier = identifier;
+    this->index = index;
+    increaseThreadCounter();
+    ScriptThread::connect(this, SIGNAL(callBack(ScriptThread*)), ss, SLOT(onScriptThreadDone(ScriptThread*)));
+}
+
+void ScriptThread::increaseThreadCounter(){
```
pbek, line 31: please use a white-space before the {

src/threads/scriptthread.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+int ScriptThread::getThreadCounter(){
+    return threadCounter[identifier];
+}
+
+// We overrides the QThread's run() method here
+// run() will be called when a thread starts
+// the code will be shared by all threads
+void ScriptThread::run()
+{
```
pbek, line 47: please put that in the previous line

src/threads/scriptthread.cpp
```diff
@@ -0,0 +1,52 @@
...
+ * for more details.
+ *
+ */
+
+#include "scriptthread.h"
+#include <QDebug>
+
+QMap<QString, int> ScriptThread::threadCounter;
+
+ScriptThread::ScriptThread(ScriptingService *ss,
```
pbek, line 20: Can't you do that all with QFuture?

src/services/scriptingservice.h
```diff
@@ -47,7 +49,10 @@ class ScriptingService : public QObject {
     bool methodExists(const QString &methodName) const;
     static bool validateScript(const Script &script, QString &errorMessage);
     Q_INVOKABLE bool startDetachedProcess(const QString &executablePath,
-                                          const QStringList &parameters);
+                                          const QStringList &parameters,
+                                          const QString &identifier = QString(),
```
pbek, line 53: I guess `callbackIdentifier` would be a better name

src/services/scriptingservice.cpp
```diff
@@ -2129,3 +2149,29 @@ void ScriptingService::triggerMenuAction(const QString &objectName,
...
+    QMapIterator<int, ScriptComponent> i(_scriptComponents);
+    TerminalCmd* cmd = thread->getTerminalCmd();
+    while (i.hasNext()) {
+        i.next();
+        ScriptComponent scriptComponent = i.value();
+        QVariantList cmdList, threadList;
+        cmdList << cmd->executablePath << cmd->parameters << cmd->exitCode;
+        threadList << thread->getIndex() << thread ->getThreadCounter();
+        if (methodExistsForObject(
+                scriptComponent.object, QStringLiteral("onCallback(QVariant,QVariant,QVariant,QVariant)"))) {
```
pbek, line 2168: I guess `onDetachedProcessCallback` would be a better name

src/services/scriptingservice.h
```diff
@@ -47,7 +49,10 @@ class ScriptingService : public QObject {
...
+                                          const int index = 0,
```
pbek, line 54: Wouldn't it be better to use a `QVariant callbackParameter`, then a script could pass everything

src/QOwnNotes.pro
```diff
@@ -125,6 +125,7 @@ SOURCES += main.cpp\
     helpers/fakevimproxy.cpp \
     helpers/flowlayout.cpp \
     services/databaseservice.cpp \
+    threads/scriptthread.cpp \
```
pbek, line 128: this also needs to go into the cmakefile

src/QOwnNotes.pro
```diff
@@ -178,6 +179,7 @@ HEADERS  += mainwindow.h \
     helpers/LanguageCache.h \
     helpers/codetohtmlconverter.h \
     helpers/qownspellchecker.h \
+    threads/scriptthread.h \
```
pbek, line 182: this also needs to go into the cmakefile

src/QOwnNotes.pro
pbek: ...and to the test project file

src/QOwnNotes.pro
pbek: ...and to the test project file

src/utils/misc.cpp
r00tr4v3n: This was copied from `startSynchronousProcess`?

src/threads/scriptthread.cpp
r00tr4v3n: Done. Is there a specific reason for this or just a convention?

pbek: Convention, please generally try to mimic the current code style (as long as it does not contradict itself) :grin:

r00tr4v3n: Sure

src/threads/scriptthread.cpp
r00tr4v3n: Don't know, I'll have a look. As I warned, I'm new to Qt.

src/services/scriptingservice.h
r00tr4v3n: Could you please elaborate?

src/QOwnNotes.pro
r00tr4v3n: `cmakelist.txt`?

pbek: yes

src/services/scriptingservice.h
pbek: What are you using the index for? Is that something that is passed by the script?

src/threads/scriptthread.cpp
r00tr4v3n: Do you mean the entire class or just the thread counting?

pbek: Before your pull request I thought about doing everything with `QFutureWatcher` (like in the `src/widgets/navigationwidget.cpp`). Maybe then you don't even need the whole new class...

src/QOwnNotes.pro
r00tr4v3n: There are multiple `test.pro` files, but I don't think those are what you mean?

src/services/scriptingservice.h
r00tr4v3n: Yes, e. g. the index of a for-loop when starting a bunch of commands associated with one callbackIdentifier.

src/QOwnNotes.pro
pbek: `tests/unit_tests/testcases/app/app-tests.pri` and `src/CMakeLists.txt`

src/services/scriptingservice.h
pbek: if you use `QVariant` the script could also pass a string, float, bool or a stringlist... the variable isn't interpreted or modified in C++, right? just passed along...

r00tr4v3n: Ah you are talking about generalizing the index variable, good point!

src/threads/scriptthread.cpp
r00tr4v3n: QFutureWatcher looks interesting, I could try rewriting it but that might currently be beyond me.

src/services/scriptingservice.h
pbek: exactly

src/threads/scriptthread.cpp
pbek: Maybe I'll look at it later. It should only be a view lines of code (if I didn't thought wrong).

docs/content/scripting/examples/callback.qml
```diff
@@ -0,0 +1,39 @@
+import QtQml 2.0
+import com.qownnotes.noteapi 1.0
+
+/**
+ * This script is just an example of how to use scripts
```
pbek, line 5: Can you please be more descriptive of what the script is for. :grin:

docs/content/scripting/examples/callback.qml
```diff
@@ -0,0 +1,39 @@
...
+    /**
+    * This function is called when a script thread is done executing.
+    * Hint: thread[1]==0 helps to determine if a bulk of started processes for a certain identifier is done.
+    *
+    * @param {QString} identifier - the provided id when calling startDetachedProcess()
+    * @param {QString} resultSet - the result of the process
+    * @param {QVariantList} cmd - the entire command array [0-executablePath, 1-parameters, 2-exitCode]
+    * @param {QVariantList} thread - the thread information array [0-passed callbackParameter, 1-remaining threads for this identifier]
+    */
+    function onDetachedProcessCallback(identifier, resultSet, cmd, thread){
```
pbek, line 34: Please space before {

docs/content/scripting/examples/callback.qml
```diff
@@ -0,0 +1,39 @@
...
+      if(identifier == "callback-example"){
```
pbek, line 35: Please space after if and before {

docs/content/scripting/examples/callback.qml
```diff
@@ -0,0 +1,39 @@
...
+ * This script is just an example of how to use scripts
+ * Visit http://docs.qownnotes.org/ for more information about scripting
+ */
+QtObject {
+    /**
+     * Just show some log entries when the script is initialized
+     */
+
+    function log(txt){
+     script.log("[callback-example] "+txt);
```
pbek, line 14: Please space before and after `+`

docs/content/scripting/examples/callback.qml
```diff
@@ -0,0 +1,39 @@
+import QtQml 2.0
+import com.qownnotes.noteapi 1.0
+
+/**
+ * This script is just an example of how to use scripts
+ * Visit http://docs.qownnotes.org/ for more information about scripting
+ */
+QtObject {
```
pbek, line 8: Please use `Script` (like in the newer scripts)

docs/content/scripting/hooks.md
```diff
@@ -1,5 +1,24 @@
 # Hooks
 
+onCallback
```
pbek, line 3: Please update the documentation

docs/content/scripting/hooks.md
```diff
@@ -1,5 +1,24 @@
...
+    /**
+     * This function is called when a script thread is done executing.
+     * Hint: thread[1]==0 helps to determine if a bulk of started processes for a certain identifier is done.
+     *
+     * @param {QString} identifier - the provided id when calling startDetachedProcess()
+     * @param {QString} resultSet - the result of the process
+     * @param {QVariantList} cmd - the entire command array [0-executablePath, 1-parameters, 2-exitCode]
+     * @param {QVariantList} thread - the thread information array [0-passed callbackParameter, 1-remaining threads for this identifier]
+     */
+    function onCallback(identifier, resultSet, cmd, thread);
```
pbek, line 17: Please update the documentation (name)

docs/content/scripting/methods-and-objects.md
```diff
@@ -11,15 +11,29 @@ Starting an external program in the background
      *
      * @param executablePath the path of the executable
      * @param parameters a list of parameter strings
+     * @param identifier an id to be used in the onDetachedProcessCallback() function (optional)
+     * @param callbackParameter an additional parameter for loops or the like (optional)
      * @return true on success, false otherwise
      */
-    bool startDetachedProcess(QString executablePath, QStringList parameters);
+    bool startDetachedProcess(QString executablePath, QStringList parameters, QString identifier, QVariant callbackParameter);
```
pbek, line 18: callbackIdentifier

docs/content/scripting/examples/callback.qml
r00tr4v3n: I'm getting this for `Script`:
```
Script is not a type Script { ^)
```

pbek: Is this missing?

```
import QtQml 2.0
import QOwnNotesTypes 1.0
```

src/threads/scriptthread.cpp
Waqar144: Inheriting from `QThread` is generally not the best idea, it is usually for threads that maybe running for long periods of time. A worker thread is a better solution in many cases.

src/utils/misc.cpp
Waqar144: yes but the variables, `executablePath` and so on are not declared, so it will not compile

src/utils/misc.h
```diff
@@ -30,6 +30,14 @@ struct ExternalImageHashItem {
...
 };
 
+struct TerminalCmd {
+    QString executablePath;
+    QStringList parameters;
+    QByteArray data;
+    int exitCode;
+    QByteArray resultSet;
+};
+
```
Waqar144, line 40: I am not too sure about this approach. This object will be emitted in every translation unit `misc.h` gets included in. And `misc` is included in a *lot* of places.

docs/content/scripting/examples/callback.qml
r00tr4v3n: Yes, `QOwnNotesTypes 1.0` was missing. It's missing in `example-script` I used as base.
What is the advantage of Script compared to the QtObject parser?

src/threads/scriptthread.cpp
Waqar144: https://doc.qt.io/qt-5/threads-technologies.html a nice overview

docs/content/scripting/examples/callback.qml
pbek: Just to make sure everyone understands that the QML is a script (for QOwnNotes).

docs/content/scripting/examples/callback.qml
pbek: I can't see any spaces. :grin:

pbek: And please use 4 spaces for indenting all the time (not 4 and 2)

src/utils/misc.cpp
pbek: See: https://travis-ci.org/github/pbek/QOwnNotes/jobs/688374906#L808 :grin:

pbek: `cmd.executablePath` and `cmd.parameters` will work better

docs/content/scripting/examples/callback.qml
r00tr4v3n: Wrong file :see_no_evil:

src/threads/scriptthread.cpp
r00tr4v3n: I've ran across  that overview but I was missing some minimal working example.

src/utils/misc.h
r00tr4v3n: Isn't that just a declaration? Should the struct be a typedef?

src/threads/scriptthread.cpp
Waqar144: yes, that is missing indeed. But we have a small and simple implementation of `QtConcurrent` in `navigationWidget` so that can be used as reference.

src/utils/misc.h
Waqar144: Its both a declaration and a definition. One way to avoid this could be to just forward declare it here and then define it where it is actually being used.

r00tr4v3n: OK. Is typedef not used for structs in Qt/C++ anymore?

src/threads/scriptthread.cpp
r00tr4v3n: I'll give it a shot.

r00tr4v3n: It works fine for one thread but `QFutureWatcher` is only meant to watch a single `QFuture`. However, the function might be invoked multiple times, so the mechanism needs a vector/list that connects in addition all signal/slots accordingly.
I don't see a Qt multithreading technology for multiple single life thread with result values other than `QThread`. :confused:

src/utils/misc.h
Waqar144: It's not a performance issue at all. Just that it's not being used here, and we could get around it by just defining it in the source file or the place where it actually gets used.

src/threads/scriptthread.cpp
Waqar144: If it doesn't work then yeah `QThread` is the only choice of course.

src/utils/misc.h
r00tr4v3n: I see now what you mean, I'll think about an alternative.

docs/content/scripting/examples/callback.qml
```diff
@@ -0,0 +1,37 @@
+import QtQml 2.0
+import com.qownnotes.noteapi 1.0
```
pbek, line 2: Needs to be `import QOwnNotesTypes 1.0` to work...

PR #1760, opened 2020-05-20
Optimize queries to get only required data

Waqar144:
This will improve startup time a little bit and will improve `reloadTagTree` performance

### DIFF
--- a/src/entities/notesubfolder.cpp
+++ b/src/entities/notesubfolder.cpp
@@ -320,7 +320,7 @@ QVector<int> NoteSubFolder::fetchIdsRecursivelyByParentId(int parentId) {
     } else {
         for (int r = 0; query.next(); r++) {
             const int id = query.value(QStringLiteral("id")).toInt();
-            idList.append(id);
+            idList << fetchIdsRecursivelyByParentId(id);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/entities/notesubfolder.cpp
```diff
@@ -303,13 +304,24 @@ QVector<NoteSubFolder> NoteSubFolder::fetchAllByParentId(
...
+
     QVector<int> idList = QVector<int>{parentId};
-    const auto noteSubFolders = fetchAllByParentId(parentId);
-    idList.reserve(noteSubFolders.count());
 
-    for (const NoteSubFolder& noteSubFolder : noteSubFolders) {
-        const int id = noteSubFolder.getId();
-        idList << fetchIdsRecursivelyByParentId(id);
+    const QString& sql = QStringLiteral(
+        "SELECT id FROM noteSubFolder WHERE parent_id = "
```
pbek, line 313: Are you sure this does the same as `fetchIdsRecursivelyByParentId` on all of `fetchAllByParentId`? Your query doesn't get the ids recursively... It basically only substitutes `fetchAllByParentId` alone.

Waqar144: Yes, i will fix it.

PR #1767, opened 2020-05-21
Simplify CMakeLists.txt. Part 1

tarhan:
First of all Part 1 because of libraries are not wrapped into targets in its own `CMakeLists.txt`.

For now this pull request provide same building as before. So it will not compile under Windows OS for now.
In Linux I've tested it with Qt 5.9.0 and 5.14.2

Let's describe various modifications I've made:
1. replaced most global CMAKE variables to target based. This change is not mandatory but provide more clean global space. For `AUTOMOC`, `AUTOUIC`, `AUTORCC` is obvious.
`CMAKE_INCLUDE_CURRENT_DIR` appends current source and binary folder into include directories variable. So I've replaced that this `target_include_directories`.

2. Using multiple `find_package(Qt5<SomeModule>)` possible but incorrect.
This could be problem on user's work PCs. If user define, for example, Qt5Widgets_DIR pointing to Qt 5.7 and Qt5Network_DIR pointing to Qt 5.14 code compile but it will definitely crash.  
When you use multiple Qt5 modules it is recommended to use just Qt5 with specifying desired its components. Using `find_package(Qt5 COMPONENTS Core Widget)` provide you always same Qt 5 version for all Qt5 modules.  
Also you can use `find_package(Qt5` multiple times as I done for conditionally finding `X11Extras` module on UNIX platform

3. Using transitive dependencies of CMake.
Despite its name `target_link_libraries` not only **link** to given target but also pull from it PUBLIC include directories list, its own dependencies, compiler options, compiler definitions.  
So if library's CMake defines target you **should** use its targets not ${lib_INCLUDES}, ${lib_LIBRARIES} to minimize errors. 
Qt5 defines its targets within Qt5 namespace (using ALIAS). So for `Core` components there will `Qt5::Core` imported target. All you need is to include desired modules into `target_link_libraries` without `include_directories`, `set(CMAKE_CXX_FLAGS`, `add_definitions`.

4. Remove `qt5_wrap_ui`.
Since you define `AUTOUIC` option you no longer need to `qt5_wrap_ui` since their behavior almost identical.

5. Remove `qt5_add_resources`
Using `qt5_add_resources` and `AUTORCC` option is identical. And latter is more clean looking. 
Instead of preparing binary resources and then put them into sources of executable it is more understandable to put resources into SOURCES.
On contrast if you have large resource which you will want to dynamically load on demand then you must use `qt5_add_binary_resources`. But not in our case.

6. Split `add_executable` into calls of  empty `add_executable` and filled `target_sources`.
Such split was not possible before CMake 3.1. Usually for small projects such split not mandatory.
Putting `add_executable` with empty `""` sources list at top of file provide quick information about what targets will be in this file.  
As for `target_sources` it will add sources list into target's `SOURCES` property same as `add_executable`. But you can call it multiple times. For example, instead of conditionally filling `SOURCE_FILES` for `APPLE` platform you call `target_sources` within `if(APPLE)` block.
Also `target_sources` usable for modular defining sources for target. For example you can create `CMakeLists.txt` or '*.cmake' file within `src/widgets` folder and it will allows decrease length of main `CMakeLists.txt` file and selectively switch of some `feature` from main file without even knowing with files it consists of.  
For now I've used `target_sources` just to be able to move `add_executable` to top of file.

7. Replace generic `target_link_libraries` call with one using `PRIVATE` link type.
In case of executables it does not matter `PRIVATE` or `PUBLIC`.  
If you use form with specifying propagation type you can call `target_link_libraries` several times.
For example, in our case, I've conditionally link to `X11` related libraries only on `UNIX` platform.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ prime
 *.snap
 snap/.snapcraft/state
 src/languages/*.qm
+.vscode/
+CMakeLists.txt.user*
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,12 +1,6 @@
 cmake_minimum_required(VERSION 3.2.2)
 project(QOwnNotes)
 
-#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-set(CMAKE_AUTOMOC ON)
-set(CMAKE_AUTOUIC ON)
-#set(CMAKE_AUTORCC ON)
-set(CMAKE_INCLUDE_CURRENT_DIR ON)
-
 # include some Botan settings
 include(libraries/botan/CMakeLists.txt)
 #include sonnet core directory, required for building plugins
@@ -15,20 +9,40 @@ include_directories(libraries/sonnet/src/core)
 include(libraries/sonnet/src/plugins/hunspell/hunspell/CMakeLists.txt)
 
 
-find_package( Qt5LinguistTools )
-find_package( Qt5Core 5.3 REQUIRED )
-find_package( Qt5Widgets REQUIRED )
-find_package( Qt5Gui REQUIRED )
-find_package( Qt5Sql REQUIRED )
-find_package( Qt5Svg REQUIRED )
-find_package( Qt5Network REQUIRED )
-find_package( Qt5Xml REQUIRED )
-find_package( Qt5XmlPatterns REQUIRED )
-find_package( Qt5PrintSupport REQUIRED )
-find_package( Qt5WebSockets REQUIRED )
-find_package( Qt5Qml REQUIRED )
-find_package( Qt5X11Extras REQUIRED )
-find_package( X11 REQUIRED )
+add_executable(QOwnNotes "")
+
+set_target_properties(QOwnNotes PROPERTIES
+    CXX_STANDARD 11
+    CXX_STANDARD_REQUIRED ON
+    AUTOMOC ON
+    AUTOUIC ON
+    AUTORCC ON
+)
+
+target_include_directories(QOwnNotes PRIVATE
+    ${CMAKE_CURRENT_LIST_DIR}
+    ${CMAKE_CURRENT_BINARY_DIR}
+)
+
+find_package(Qt5 COMPONENTS REQUIRED
+    LinguistTools
+    Core
+    Widgets
+    Gui
+    Sql
+    Svg
+    Network
+    Xml
+    XmlPatterns
+    PrintSupport
+    WebSockets
+    Qml
+)
+if(UNIX)
+    find_package(Qt5 COMPONENTS REQUIRED X11Extras)
+    find_package( X11 REQUIRED )
+endif()
+
 find_program(CCACHE_FOUND ccache)
 
 if(CCACHE_FOUND)
@@ -37,49 +51,18 @@ if(CCACHE_FOUND)
     set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
 endif(CCACHE_FOUND)
 
-qt5_wrap_ui(dialogs/ui_settingsdialog.h settingsdialog.ui)
-qt5_wrap_ui(dialogs/ui_tododialog.h tododialog.ui)
-qt5_wrap_ui(dialogs/ui_aboutdialog.h aboutdialog.ui)
-qt5_wrap_ui(widgets/ui_logwidget.h logwidget.ui)
-qt5_wrap_ui(dialogs/ui_trashdialog.h trashdialog.ui)
-qt5_wrap_ui(dialogs/ui_localtrashdialog.h localtrashdialog.ui)
-qt5_wrap_ui(dialogs/ui_linkdialog.h linkdialog.ui)
-qt5_wrap_ui(dialogs/ui_updatedialog.h updatedialog.ui)
-qt5_wrap_ui(dialogs/ui_notediffdialog.h notediffdialog.ui)
-qt5_wrap_ui(dialogs/ui_versiondialog.h versiondialog.ui)
-qt5_wrap_ui(dialogs/ui_welcomedialog.h welcomedialog.ui)
-qt5_wrap_ui(dialogs/ui_issueassistantdialog.h issueassistantdialog.ui)
-qt5_wrap_ui(dialogs/ui_tagadddialog.h tagadddialog.ui)
-qt5_wrap_ui(widgets/ui_fontcolorwidget.h fontcolorwidget.ui)
-qt5_wrap_ui(widgets/ui_evernoteimportdialog.h ui_evernoteimportdialog.ui)
-qt5_wrap_ui(widgets/ui_serverbookmarksimportdialog.h ui_serverbookmarksimportdialog.ui)
-qt5_wrap_ui(widgets/ui_layoutwidget.ui ui_layoutwidget.ui)
-qt5_wrap_ui(dialogs/ui_orphanedimagesdialog.h orphanedimagesdialog.ui)
-qt5_wrap_ui(dialogs/ui_orphanedattachmentsdialog.h orphanedattachmentsdialog.ui)
-qt5_wrap_ui(dialogs/ui_actiondialog.h actiondialog.ui)
-qt5_wrap_ui(dialogs/ui_tabledialog.h tabledialog.ui)
-qt5_wrap_ui(dialogs/ui_notedialog.h notedialog.ui)
-qt5_wrap_ui(dialogs/ui_websockettokendialog.h websockettokendialog.ui)
-qt5_wrap_ui(dialogs/ui_imagedialog.h imagedialog.ui)
-qt5_wrap_ui(libraries/qmarkdowntextedit/ui_qplaintexteditsearchwidget.h
-        libraries/qmarkdowntextedit/qplaintexteditsearchwidget.ui)
-qt5_wrap_ui(libraries/qttoolbareditor/src/toolbar_editor.hpp
-        libraries/qttoolbareditor/src/toolbar_editor.ui)
-
 set(RESOURCE_FILES
-        breeze-qownnotes.qrc
-        breeze-dark-qownnotes.qrc
-        qownnotes.qrc
-        demonotes.qrc
-        images.qrc
-        texts.qrc
-        configurations.qrc
-        libraries/qmarkdowntextedit/media.qrc
-        libraries/qdarkstyle/style.qrc
-        libraries/qkeysequencewidget/qkeysequencewidget/qkeysequencewidget.qrc
-        )
-
-qt5_add_resources(RESOURCE_ADDED ${RESOURCE_FILES})
+    breeze-qownnotes.qrc
+    breeze-dark-qownnotes.qrc
+    qownnotes.qrc
+    demonotes.qrc
+    images.qrc
+    texts.qrc
+    configurations.qrc
+    libraries/qmarkdowntextedit/media.qrc
+    libraries/qdarkstyle/style.qrc
+    libraries/qkeysequencewidget/qkeysequencewidget/qkeysequencewidget.qrc
+)
 
 set(SOURCE_FILES
     dialogs/welcomedialog.cpp
@@ -255,7 +238,7 @@ set(SOURCE_FILES
     libraries/piwiktracker/piwiktracker.h
     libraries/piwiktracker/piwiktracker.cpp
     libraries/botan/botan.h
-    libraries/botan/botan_internal.h    
+    libraries/botan/botan_internal.h
     libraries/botan/botan.cpp
     libraries/botan/botanwrapper.h
     libraries/botan/botanwrapper.cpp
@@ -291,8 +274,8 @@ set(SOURCE_FILES
     libraries/qhotkey/QHotkey/qhotkey_p.h
     libraries/qhotkey/QHotkey/qhotkey_x11.cpp
 
-	threads/scriptthread.cpp
-	threads/scriptthread.h
+    threads/scriptthread.cpp
+    threads/scriptthread.h
 
     widgets/qownnotesmarkdowntextedit.cpp
     widgets/qownnotesmarkdowntextedit.h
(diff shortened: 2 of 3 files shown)


pbek:
Thank you, it looks very tidy. I'll test it when I'm at my development rig.

tarhan:
I've added second commit to support compiling under Windows using MinGW toolchain.

Changes consist of two parts.

I've isolated **Botan** library in its own library as CMake target.  
Within its `CMakeLists.txt` I've copied defines from `src/libraries/botan/botan.pri`.
Some defines are not used at all so I removed them. Some other used only internally by **Botan** sources so I've put them under `PRIVATE` scope.
Since it is recommended by CMake not to pollute global namespace. I've create `ALIAS` target `Botan::Botan` to be used outside.

**QHotKey** library already had well defined target (almost). So I've just linked to it.  
Following lines in main `CMakeLists.txt`
```cmake
# Disable building tests inside QHotKey library
set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
set(QHOTKEY_EXAMPLES OFF)
```
need to disable compiling test application of **QHotKey** without any warning.
Because [qownnotes/QHotkey](https://github.com/qownnotes/QHotkey) does not include [83a0066](https://github.com/Skycoder42/QHotkey/commit/83a0066760140e676fc12278be6f1aedde3a5987) commit of its upstream test application of **QHotKey** will not able to compile.

tarhan:
I've added commit for compiling using Visual Studio toolchain.

Small fix made in `src/libraries/botan/CMakeLists.txt` to suppress warning as well as error starting from VS 2017 15.8.

Library **FakeVim** almost ready to be used as CMake target `fakevim`. Only couple compiler defines required in main `CMakeLists.txt` until [pbek/FakeVim](https://github.com/pbek/FakeVim) repository will be patched for MSVC.

As for now in addition to MinGW I've tested compilation `CMakeLists.txt` using MSVC in following configurations:
- Qt Creactor with MSVC 2017 x64 toolchain
- Visual Studio 2017 using Visual Studio 2017 x64 CMake Generator (via cmake-gui)
- Visual Studio 2019 using Visual Studio 2019 x64 CMake Generator (via cmake-gui)

pbek:
Looking great, thank you!

src/CMakeLists.txt
```diff
@@ -1,549 +1,534 @@
...
+    Qml
+)
+if(UNIX)
+    find_package(Qt5 COMPONENTS REQUIRED X11Extras)
+    find_package( X11 REQUIRED )
+endif()
+
+find_program(CCACHE_FOUND ccache)
+
+if(CCACHE_FOUND)
```
Waqar144, line 64: this time we use `ccache` in a better way. Instead of stuffing everything here, lets make a cmake folder and move stuff like this over there. This will include:
- ccache
- Warnings
- Sanitizers

see: https://github.com/lefticus/cpp_starter_project/

Its mostly just copy paste

src/CMakeLists.txt
```diff
@@ -1,549 +1,534 @@
...
+    breeze-qownnotes.qrc
+    breeze-dark-qownnotes.qrc
+    qownnotes.qrc
+    demonotes.qrc
+    images.qrc
+    texts.qrc
+    configurations.qrc
+    libraries/qmarkdowntextedit/media.qrc
+    libraries/qdarkstyle/style.qrc
+    libraries/qkeysequencewidget/qkeysequencewidget/qkeysequencewidget.qrc
```
Waqar144, line 80: can we not do this? `qkeysequencewidget` has a cmake file already, we can just use the variables from there or link to it instead

src/CMakeLists.txt
```diff
@@ -1,549 +1,534 @@
...
+    dialogs/notedialog.ui
+    dialogs/websockettokendialog.cpp
+    dialogs/websockettokendialog.h
+    dialogs/websockettokendialog.ui
+    dialogs/imagedialog.cpp
+    dialogs/imagedialog.h
+    dialogs/imagedialog.ui
+    dialogs/attachmentdialog.cpp
+    dialogs/attachmentdialog.h
+    dialogs/attachmentdialog.ui
```
Waqar144, line 163: no this is too long. If we are going to revise our cmake, lets use cmake `include` and file globbing to make our lives easier. We should do this for all main directories i.e., `dialogs`, `entities`, `helpers`, `api`, `widgets` and so on.

src/CMakeLists.txt
```diff
@@ -1,549 +1,534 @@
...
+    helpers/codetohtmlconverter.cpp
+    helpers/codetohtmlconverter.h
+    api/noteapi.cpp
+    api/noteapi.h
+    api/tagapi.cpp
+    api/tagapi.h
+    api/scriptapi.cpp
+    api/scriptapi.h
+    libraries/diff_match_patch/diff_match_patch.cpp
+    libraries/diff_match_patch/diff_match_patch.h
```
Waqar144, line 206: All the libraries would ideally be linked using `target_link_libraries`

src/CMakeLists.txt
```diff
@@ -1,549 +1,534 @@
...
+    mainwindow.ui
+    release.h
+    version.h
+    helpers/fakevimproxy.h
+    helpers/fakevimproxy.cpp
+    services/websocketserverservice.cpp
+    services/websocketserverservice.h
+    entities/bookmark.h entities/bookmark.cpp)
+
+if(NOT APPLE)
```
Waqar144, line 338: this is unnecessary now. All platforms use hunspell so you can remove this check

src/CMakeLists.txt
```diff
@@ -1,549 +1,534 @@
...
+
+if (APPLE)
+   list(APPEND SOURCE_FILES
+    libraries/sonnet/src/plugins/nsspellchecker/nsspellcheckerclient.h
+    libraries/sonnet/src/plugins/nsspellchecker/nsspellcheckerclient.mm
+    libraries/sonnet/src/plugins/nsspellchecker/nsspellcheckerdebug.cpp
+    libraries/sonnet/src/plugins/nsspellchecker/nsspellcheckerdebug.h
+    libraries/sonnet/src/plugins/nsspellchecker/nsspellcheckerdict.h
+    libraries/sonnet/src/plugins/nsspellchecker/nsspellcheckerdict.mm)
+endif()
```
Waqar144, line 400: `nsspellchecker` is disabled, this part can safely be removed.

src/CMakeLists.txt
```diff
@@ -1,549 +1,534 @@
...
+
+# Translation files
+SET(QON_TS_FILES
```
Waqar144, line 403: lets move this part to `languages/` instead and `include()` here.

.gitignore
```diff
@@ -15,3 +15,5 @@ prime
 *.snap
 snap/.snapcraft/state
 src/languages/*.qm
+.vscode/
+CMakeLists.txt.user*
```
Waqar144, line 19: instead of this can you just ignore `*.user` and remove this and `*.pro.user` at the top

src/CMakeLists.txt
```diff
@@ -1,549 +1,534 @@
...
-        ${Qt5Qml_LIBRARIES} ${Qt5X11Extras_LIBRARIES} ${X11_LIBRARIES})
-
-#qt5_use_modules(QOwnNotes Widgets Core Gui Sql Svg Network Script Xml PrintSupport)
-
-# Sonnet support if we can get it to run in the future
-#find_package(KF5Sonnet)
-#target_link_libraries(KF5::SonnetCore KF5::SonnetUi)
-
-add_compile_definitions(QAPPLICATION_CLASS=QApplication)
+cmake_minimum_required(VERSION 3.2.2)
```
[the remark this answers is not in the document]
tarhan: No. I think CMake minimum version should as minimum as possible. But at same time high enough not to pulling very outdated CMake techniques.

src/CMakeLists.txt
tarhan: Yes. But it is not good to make many steps in single commit. One by one upgrade.
And that starter project have several odds.
For example, instead of adding `cmake` folder as modules folder, author includes cmake files.
More correct way is to put following line at root `CMainLists.txt`:
```cmake
list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
```
After that in that example project instead of 
```cmake
include(cmake/CompilerWarnings.cmake)
```
you will write
```cmake
include(CompilerWarnings)
```
```

File: src/CMakeLists.txt (hunk 2 above)
Comment (tarhan) [reply src/CMakeLists.txt:80]: I could not do all changes in one evening.

File: src/CMakeLists.txt (hunk 3 above)
Comment (tarhan) [reply src/CMakeLists.txt:163]: File globbing is very bad thing from my point of view. You can use it only in initial phase of converting project from qmake or Autotools. I've saw using glob only in very bad written tutorials.  

But you right it is good idea to use `src/CMakeLists.txt` only for setup project and filling actual sources from subfolders.
I'm myself prefer using `add_subdirectory` for this instead of `include`. Compare:
```cmake
add_subdirectory(dialogs)
# vs
include(dialogs/some_cmake_name.cmake)
```

`target_sources` if it called after `add_executable` even from inner `CMakeLists.txt` could see that target.
So `dialogs/CMakeList.txt` will look like:
```cmake
target_sources(QOwnNotes PRIVATE
    dialogs/imagedialog.ui
    dialogs/attachmentdialog.cpp
    dialogs/attachmentdialog.h
    dialogs/attachmentdialog.ui
)
```
Here you can see "problem" with `target_sources`. Before CMake 3.13 `target_sources` does not support relative to `CMakeLists.txt` sources file names. Before 3.13 it think that file name should be relative to target they pointing.
So instead of require user to have CMake 3.13 there two methods to overcome this.
One is using `${CMAKE_CURRENT_LIST_DIR}`:
```cmake
set(SRC ${CMAKE_CURRENT_LIST_DIR})
target_sources(QOwnNotes PRIVATE
    ${SRC}/imagedialog.ui
    ${SRC}/attachmentdialog.cpp
    ${SRC}/attachmentdialog.h
    ${SRC}/attachmentdialog.ui
)
```
This allows you to move lines between folders without modifying folder name.
But I do not like how it looks like.
There another method. Is to use macro `target_sources_local` which will works like `target_sources` from CMake 3.13.
It will be placed into `cmake` folder and included either in main or root `CMakeLists.txt`.
Here my `TargetSourcesLocal.cmake` source:
```cmake
function(target_sources_local target)
  if(POLICY CMP0076)
    # New behavior is available, so just forward to it by ensuring
    # that we have the policy set to request the new behavior, but
    # don't change the policy setting for the calling scope
    cmake_policy(PUSH)
    cmake_policy(SET CMP0076 NEW)
    target_sources(${target} ${ARGN})
    cmake_policy(POP)
    return()
  endif()

  # Must be using CMake 3.12 or earlier, so simulate the new behavior
  unset(_srcList)
  get_target_property(_targetSourceDir ${target} SOURCE_DIR)

  foreach(src ${ARGN})
    if(NOT src STREQUAL "PRIVATE" AND
       NOT src STREQUAL "PUBLIC" AND
       NOT src STREQUAL "INTERFACE" AND
       NOT IS_ABSOLUTE "${src}")
      # Relative path to source, prepend relative to where target was defined
      file(RELATIVE_PATH src "${_targetSourceDir}" "${CMAKE_CURRENT_LIST_DIR}/${src}")
    endif()
    list(APPEND _srcList ${src})
  endforeach()
  target_sources(${target} ${_srcList})
endfunction()
```
In main or root `CMakeLists.txt` we will add it:
```cmake
include(TargetSourcesLocal)
```
After that `src/dialogs/CMakeLists.txt` will looks like:
```cmake
target_sources_local(QOwnNotes PRIVATE
    imagedialog.ui
    attachmentdialog.cpp
    attachmentdialog.h
    attachmentdialog.ui
)
```
Which is much cleaner

src/CMakeLists.txt
tarhan: Yes. In time.

src/CMakeLists.txt
tarhan: I don't have Mac so I do not know.

src/CMakeLists.txt
tarhan: So I can fully remove it? 
And only keep hunspell?

src/CMakeLists.txt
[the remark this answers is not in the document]
tarhan: @Waqar144 What command you want which requires CMake 3.13+ ?

src/CMakeLists.txt
Waqar144: yes,
but the correct way  to handle this would be to build `sonnet` as a library (hunspell will be a dependency there) and link that statically into QON

src/CMakeLists.txt
Waqar144: Haha, I dont have mac either. I wrote this back when i first integrated the spellchecker, didn't workout very well so we removed it. Seems like no one updated the CMake ;)

src/CMakeLists.txt
Waqar144: > File globbing is very bad thing from my point of view. 

Really? I am curious as to why its bad :thinking: Are there any negative points?

> I'm myself prefer using `add_subdirectory` for this instead of `include`. Compare:

I use `add_subdirectory` when i want to make the subdirectory a static library, otherwise I stick to `include()`. Is there a way I can use `add_subdirectory()` without having to link it to some target?


> One is using `${CMAKE_CURRENT_LIST_DIR}`:

Yes, I learned this the hard way :laughing: 

> Which is much cleaner

Agreed. However it makes things more difficult to understand for someone who is not very good at cmake.

Btw, I just create `.cmake` files for all the subdirectories. Basically its just like this:

```cmake
file(GLOB dialogs_sources ${CMAKE_CURRENT_LIST_DIR}/*.cpp)
file(GLOB dialogs_headers ${CMAKE_CURRENT_LIST_DIR}/*.h)
file(GLOB dialogs_ui ${CMAKE_CURRENT_LIST_DIR}/*.h)

set(Dialogs ${dialogs_sources} ${dialogs_headers} $[dialogs_ui})

# same for other dirs, like entities/ utils/ so on.

## In main CMakeLists.txt file

set(SOURCES ${Dialogs} ${Entities} ${Utils} ...)
```

What do you think of this approach? It seems very simple and straightforward to me.

tarhan: > Really? I am curious as to why its bad  Are there any negative points?

I don't remember link to detailed description of problems. Of course most of these problems related to good code style. 
One problem related to performance is that speed of `file(GLOB` is not fast as you may think. It is far more slower that system find utility.
Another problem is then you add or remove files CMake will **NOT** re-run (since `file(GLOB` is slow it runs only once).
This particularly problematic when you checkout between git branches where file list could be changed but CMake will **NOT** re-run. In CMake 3.12 `CONFIGURE_DEPENDS` option tried to solve this but it will slow down CMake running almost 10 times and **it not supported** by some code generators like MSVC or XCode.
Also file globbing could add files not intended for Linux platform for example. You will need to put Linux code into other folder just for one single file.
Also CMake official documentation also discourage to use `file(GLOB`:

> Note. We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.

So in bottom. Since it describe you this. Next time you heard that someone try to use `file(GLOB` for source files you `should hit` him into face :)  It is easy to learn bad things and it is hard to dislearn them.

Waqar144: This really good to know. I never realized it could result in so many problems.

src/CMakeLists.txt
Waqar144: I should mention here that KDE provides Cmake files for sonnet however they depend on ECM(Extra cmake modules) to work correctly. So thats not an option

src/CMakeLists.txt
tarhan: > I use `add_subdirectory` when i want to make the subdirectory a static library, otherwise I stick to `include()`. 

`add_subdirectory` and `include` almost same. They differs how scope of inner things.  
`include` is like `#include` in **C/C++* (remember min/max problem in MSVC).
`add_subdirectory` run *code* in isolation since it creates new scope. You use pass result variable to caller's scope using cache variables only or with `PARENT_SCOPE` in `set` command.
Some commands like `add_library`, `add_executable` make their targets visible outside their `add_subdirectory` call (they visible at any level **after** `add_subdirectory` call.

> Is there a way I can use `add_subdirectory()` without having to link it to some target?

Yes. `add_subdirectory` does not have any relation to target linking.

tarhan: > However it makes things more difficult to understand for someone who is not very good at cmake

What makes things difficult? Placing file list in its own folder?

tarhan: >Btw, I just create .cmake files for all the subdirectories. Basically its just like this:
```cmake
file(GLOB dialogs_sources ${CMAKE_CURRENT_LIST_DIR}/*.cpp)
file(GLOB dialogs_headers ${CMAKE_CURRENT_LIST_DIR}/*.h)
file(GLOB dialogs_ui ${CMAKE_CURRENT_LIST_DIR}/*.h)

set(Dialogs ${dialogs_sources} ${dialogs_headers} $[dialogs_ui})
```
`set` command without options create **local** variable. That is why you are using `include` because it does not create new scope.  
Problem can raise if you forget rename during copy-paste to another folder you will lost some sources. Since scope is same new assignment will override old one.
If you use `add_subdirectory` you need *send* variable to parent scope (without polluting cache variables only to one level up) like this:
```cmake
set(Dialogs ${dialogs_sources} ${dialogs_headers} $[dialogs_ui} PARENT_SCOPE)
```
> What do you think of this approach? It seems very simple and straightforward to me.

As I've described using `file(GLOB` is bad. So I do not like it. We use CMake at work at daily basis for many utility tools. Using `file(GLOB` and also `${PROJECT_NAME}` for target name (also common practice in awful tutorials) always create more problems than solves.
You as developer should not be lazy in some things. DRY (Don't-Repeat-Yourself) does not mean that you can you `*` for everything.

tarhan: > This really good to know. I never realized it could result in so many problems.

I do not know if book ads allowed here. But I've increased CMake understanding radically when I bought e-book **Professional CMake**. It written in simple English which good for me not English native speaker.

src/CMakeLists.txt
tarhan: > yes,
> but the correct way to handle this would be to build `sonnet` as a library (hunspell will be a dependency there) and link that statically into QON

Ok. In time.

src/CMakeLists.txt
Waqar144: None of us(the devs here) are native English speakers here. I use that book for reference sometimes, but yeah I am lazy and stop as soon as something starts working :laughing:

tarhan: You are always welcome

PR #1769, opened 2020-05-22
forExport parameter for (pre/)noteToMarkdownHtmlHook

r00tr4v3n:
This comes in handy when using dark mode. 
e. g. generated formulas don't look so great white on white

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # QOwnNotes Changelog
 
+## 20.5.12
+- fixed copying and moving of notes into sub-folders and note-folders that was
+  broken by [#1749](https://github.com/pbek/QOwnNotes/pull/1749) 
+  (for [#1764](https://github.com/pbek/QOwnNotes/issues/1764))
+
 ## 20.5.11
 - some startup performance improvements were made (for [#1755](https://github.com/pbek/QOwnNotes/pull/1755), thank you @Waqar144)
 - markdown checkbox list items now have a square as list-style in the preview
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 - fixed copying and moving of notes into sub-folders and note-folders that was
   broken by [#1749](https://github.com/pbek/QOwnNotes/pull/1749) 
   (for [#1764](https://github.com/pbek/QOwnNotes/issues/1764))
+- the scripting hook `noteToMarkdownHtmlHook` will now be called after modifications
+  to images in the note are made (for [#1766](https://github.com/pbek/QOwnNotes/issues/1766))
 
 ## 20.5.11
 - some startup performance improvements were made (for [#1755](https://github.com/pbek/QOwnNotes/pull/1755), thank you @Waqar144)
--- a/docs/content/scripting/examples/preview-styling.qml
+++ b/docs/content/scripting/examples/preview-styling.qml
@@ -7,7 +7,7 @@ QtObject {
     /**
      * This function is called before the markdown html of a note is generated
      *
-     * It allows you to modify what is passed to the markdown to html converter 
+     * It allows you to modify what is passed to the markdown to html converter
      *
      * The method can for example be used in multiple scripts to render code (like LaTeX math or mermaid)
      * to its graphical representation for the preview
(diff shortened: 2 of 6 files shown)


pbek:
I'll take a look once you make it build :grin:

pbek:
Sadly this will break all scripts that are currently using `preNoteToMarkdownHtmlHook` because they have only implemented two parameters. 😬

r00tr4v3n:
> Sadly this will break all scripts that are currently using `preNoteToMarkdownHtmlHook` and `noteToMarkdownHtmlHook` because they have only implemented two parameters. grimacing

Is there no polymorphic function overloading in QML? :sob: 
Using two function might be a workaround till the legacy function is deprecated and removed.

pbek:
Adding this to `noteToMarkdownHtmlHook` is not necessary because there is already set a css class in the body tag for this.

If you really insist changing `preNoteToMarkdownHtmlHook` please create an issue where you explain why you are doing it, notify the one(s) that originally wanted to have this hook, update all scripts in this repo that are using this hook and check if there are any scripts on https://github.com/qownnotes/scripts that are using this hook.

pbek:
> Is there no polymorphic function overloading in QML? 😭

3 parameters added means 3 parameters are looked for

> Using two function might be a workaround till the legacy function is deprecated and removed.

I don't want to double the amount of hook checks at that critical place. 99% of people don't use this hook. And there will only be one person that is using the new parameter for a long time.

pbek:
`methodExistsForObject` seems to be fast enough, you can additionally add code to `ScriptingService::callPreNoteToMarkdownHtmlHook` that checks for both variants.

r00tr4v3n:
> As I said, we don't need a `noteToMarkdownHtmlHook(QVariant,QVariant,QVariant)` with `forExport`.

Why do you prefer noteTo over pre?

r00tr4v3n:
> Adding this to `noteToMarkdownHtmlHook` is not necessary because there is already set a css class in the body tag for this.

I cannot change image sizes through CSS though. The sizes vary depending on the preview window or PDF orientation.

pbek:
> I cannot change image sizes through CSS though. The sizes vary depending on the preview window or PDF orientation.

If you also want to break noteToMarkdownHtmlHook then you need do everything I said in https://github.com/pbek/QOwnNotes/pull/1769#issuecomment-632893717 for this hook too (including updating all scripts in the script repository, don't forget the bump the version number of each script)

r00tr4v3n:
If there is no performance penalty, the extra method check might be preferable for now since users might update QON without immediately the script as well?

This PR isn't urgent but long term I'd like to render documents in preview and PDF neatly without changing setting manually for the export.

r00tr4v3n:
>     1. please also mention the script name in the log, should be something like `scriptComponent.script.getName()`
> 
>     2. test if that log really works properly by using the deprecated function in a script grin

Way ahead of you after seeing the warning for a dozen scripts without knowing which ones :grin:

```shell
[23:05:32] [warning] "Warning: preNoteToMarkdownHtmlHook(note, markdown) is deprecated, please use preNoteToMarkdownHtmlHook(note, markdown, forExport) in image-toolkit"
[23:05:32] [warning] "Warning: noteToMarkdownHtmlHook(note, html) is deprecated, please use noteToMarkdownHtmlHook(note, html, forExport) in yaml-nested-tags"
[23:05:32] [warning] "Warning: noteToMarkdownHtmlHook(note, html) is deprecated, please use noteToMarkdownHtmlHook(note, html, forExport) in latex-math"
```

pbek:
> If there is no performance penalty,

Oh, I suppose there is a penalty. I just hope it isn't high. 😬

pbek:
> Way ahead of you after seeing the warning for a dozen scripts without knowing which ones 😁

very well 😁

r00tr4v3n:
> > If there is no performance penalty,
> 
> Oh, I suppose there is a penalty. I just hope it isn't high. grimacing

It is measurable! 
```cpp
    long start = QDateTime::currentMSecsSinceEpoch();
    long count = 0;
    int iter = 1000000;
    for (int i=0; i<iter; i++) {
        if (methodExistsForObject(
                scriptComponent->object,
                QStringLiteral("noteToMarkdownHtmlHook(QVariant,QVariant)"))) {
            count++;
        }
    }
    double dur = QDateTime::currentMSecsSinceEpoch()-start;
    double avg = dur / iter;
```

avg: 0.7 μs for unknown methods
avg: 0.5 μs for known methods

I guess we can deprecate a thousand methods more before we have to worry about it :grin:

pbek:
Thank you for measuring it, I merely looked at the source code. Good to have that affirmed!

pbek:
Seems like as soon you create the issue, ping the people of issues where the hooks where talked about and refactor all existing scripts we are good to go. :grin:

r00tr4v3n:
I thought the low penalty would be a reason to leave both methods for now?
Maybe an additional hint like "will be removed in v22.0".

pbek:
Are you still working on this? Because I still need a PR for the script updates before I can merge this. :grin:

r00tr4v3n:
That was my question. The benchmark showed it's fine to leave both for now and remove the support later on. What do you think about the deprecation sentence?

pbek:
I don't want the unnecessary code duplication, that's why the old calls need to get deprecated.

pbek:
Great. :grin:

pbek:
## 20.6.0
- the scripting commands `preNoteToMarkdownHtmlHook` and `preNoteToMarkdownHtmlHook`
  now have a new parameter `forExport` (for [#1769](https://github.com/pbek/QOwnNotes/pull/1769), thank you @r00tr4v3n)
    - please take a look at the documentation for
      [preNoteToMarkdownHtmlHook](https://docs.qownnotes.org/en/latest/scripting/hooks/#prenotetomarkdownhtmlhook)
      and [noteToMarkdownHtmlHook](https://docs.qownnotes.org/en/latest/scripting/hooks/#notetomarkdownhtmlhook)
      for more information

pbek:

r00tr4v3n:
Works as expected :+1:

pbek:
Great, thank you for testing!

PR #1770, opened 2020-05-22
Modernize CMake. Part 2. Isolate some libs.

tarhan:
I've replaced usage of **Botan** as it should: with angle brackes.

Also I've added option to use system installed Botan using CMake cache variable `BUILD_WITH_SYSTEM_BOTAN` to match `botan.pri` behavior. 
For now this option works only on Unix platforms where pkg-config exists (like in `botan.pri`).

Usage of `botan.h` replaced with individual headers in case of using system **Botan** library

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,4 +2,6 @@ cmake_minimum_required(VERSION 3.2.2)
 
 project(QOwnNotes)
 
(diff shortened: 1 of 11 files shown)


pbek:
Looks nice, I'll test it when I'm at my dev rig.

tarhan:
I've isolated diff_match_path library

pbek:
Thank you!

Issue #1761, opened 2020-05-20
List checkboxes do not replace bullet sign


r00tr4v3n:
A list with checkboxes keeps an additional bullet sign.

#### Expected behaviour
- [x] task

#### Actual behaviour
- - [x] task

#### Suggestion
`style="list-style-type:none;` in the `<ul>` tag should do the trick if there is a way to detect task lists beforehand.
~Unfortunately adding the style in the `<li>` tag destroys the checkbox link/href.~

pbek:
> `style="list-style-type:none;` in the `<ul>` tag should do the trick if there is a way to detect task lists beforehand.

you can mix normal list items with "todo-list" items

r00tr4v3n:
OK, so the `<li>` tag should be the way  to go? Turns out I missed a `"`...

pbek:
Very tricky

pbek:
I rewrote the parts of the todo-list parser to do the extra styling... But I wasn't able to get the indent right so far when normal list items, checked checkboxes and unchecked checkboxes are mixed. It just looks a bit ugly.

![image](https://user-images.githubusercontent.com/1798101/82496853-aa1a5300-9aed-11ea-8121-25488ff84787.png)

```
- [ ] test
    - normal text
    - [ ] test
    - [x] test
        - [ ] test
    - normal text

- [] asasddasd
- [ ] asdas
```

Mainly also because the checked and unchecked checkboxes may have different sizes.

r00tr4v3n:
> Very tricky

As cosmetic improvements usually are :smile: 

> Mainly also because the checked and unchecked checkboxes may have different sizes.

How about same-size checkbox images as bullets instead of invisible bullets and the unicode character? [fiddle](http://jsfiddle.net/cxoahgn4/)

pbek:
1) I wasn't able to use css classes on the `li` because they affected the sub-list items too in the preview 😬  I needed to add a `style` attribute to every `li` with a checkbox
2) Images would of course be a less performant in long lists
3) There would be different images needed for dark mode
4) If you want to showcase something please do it directly in the QOwnNotes preview, since the `QTextBrowser` is only able to show a subset of CSS and HTML (and is very quirky)

pbek:
[...]

PR #1771, opened 2020-05-23
Task list item checkbox fix

r00tr4v3n:
https://github.com/pbek/QOwnNotes/issues/1761

### DIFF
--- a/src/entities/note.cpp
+++ b/src/entities/note.cpp
@@ -2278,7 +2278,8 @@ QString Note::textToMarkdownHtml(QString str, const QString &notesPath,
                 "\\1file://" + windowsSlash + notePath + "/\\2\"");
     */
     // css classes applicable for export and preview
-    const QString additionalClasses = ".task-list-item-checkbox {list-style: none; text-decoration: none;  margin-left: -14;}";
+    QString additionalClasses = ".task-list-item-checkbox {list-style: none; margin-left: -14px; text-decoration: none;} "
+                                ".normal-list-item {list-style: disc; margin-left: 0px;} ";
 
     const QString fontString =
         settings.value(QStringLiteral("MainWindow/noteTextView.code.font"))
--- a/src/entities/note.cpp
+++ b/src/entities/note.cpp
@@ -2278,13 +2278,11 @@ QString Note::textToMarkdownHtml(QString str, const QString &notesPath,
                 "\\1file://" + windowsSlash + notePath + "/\\2\"");
     */
     const QString previewFontString =
-        settings.value(QStringLiteral("MainWindow/noteTextEdit.font"))//MainWindow/noteTextEdit.font MainWindow/noteTextView.font
+        settings.value(QStringLiteral("MainWindow/noteTextEdit.font"))
             .toString();
-    int previewFontSize = 22;
     QString margin_listitem_str = "";
     QString margin_checkbox_str = "";
-    if(!previewFontString.isEmpty()){
-        previewFontSize = previewFontString.split(",")[1].toInt(); // e.g. Andale Mono,12,-1,5,50,0,0,0,0,0,Regular
+    if (!previewFontString.isEmpty()) {
         QFont previewFont;
         previewFont.fromString(previewFontString);
         const int width_checkbox = QFontMetrics(previewFont).horizontalAdvance("☑"); // &#9744; and &#9745;
@@ -2294,7 +2292,6 @@ QString Note::textToMarkdownHtml(QString str, const QString &notesPath,
         const int margin_checkbox = -width_checkbox - width_bullet / 2 - 1 + margin_listitem;
         margin_listitem_str = QStringLiteral(" margin-left: %1px; ").arg(margin_listitem);
         margin_checkbox_str = QStringLiteral(" margin-left: %1px; ").arg(margin_checkbox);
-        qDebug() << margin_listitem << "c " << margin_checkbox;
     }
     // css classes applicable for export and preview
     QString listCSS = QStringLiteral(".list-item-normal {list-style: disc;  %1 } "
--- a/src/entities/note.cpp
+++ b/src/entities/note.cpp
@@ -321,9 +321,10 @@ bool Note::copyToPath(const QString &destinationPath, QString noteFolderPath) {
             return false;
         }
 
+        // check if there are media files in the note
         const QStringList mediaFileList = getMediaFileList();
         if (mediaFileList.empty()) {
-            return false;
+            return true;
         }
 
         if (noteFolderPath.isEmpty()) {
@@ -331,15 +332,16 @@ bool Note::copyToPath(const QString &destinationPath, QString noteFolderPath) {
         }
 
         if (!NoteFolder::isPathNoteFolder(noteFolderPath)) {
-            return false;
+            return true;
         }
+
         const QDir mediaDir(noteFolderPath + QDir::separator() +
                             QStringLiteral("media"));
 
         // created the media folder if it doesn't exist
         if (!mediaDir.exists()) {
             if (!mediaDir.mkpath(mediaDir.path())) {
-                return false;
+                return true;
             }
         }
 
--- a/src/entities/note.cpp
+++ b/src/entities/note.cpp
@@ -2278,8 +2278,8 @@ QString Note::textToMarkdownHtml(QString str, const QString &notesPath,
                 "\\1file://" + windowsSlash + notePath + "/\\2\"");
     */
     // css classes applicable for export and preview
-    QString listCSS = QStringLiteral(".task-list-item-checkbox {list-style: none; margin-left: -14px; text-decoration: none;} "
-                                     ".normal-list-item {list-style: disc; margin-left: 0px;} ");
+    QString listCSS = QStringLiteral(".list-item-task {list-style: none; margin-left: -14px; text-decoration: none;} "
+                                     ".list-item-normal {list-style: disc; margin-left: 0px;} ");
 
     const QString fontString =
         settings.value(QStringLiteral("MainWindow/noteTextView.code.font"))
--- a/src/entities/note.cpp
+++ b/src/entities/note.cpp
@@ -2401,14 +2401,6 @@ QString Note::textToMarkdownHtml(QString str, const QString &notesPath,
         result.replace(QStringLiteral("\n</code>"), QStringLiteral("</code>"));
     }
 
-    // check if there is a script that wants to modify the content
-    const QString scriptResult =
-        ScriptingService::instance()->callNoteToMarkdownHtmlHook(this, result);
-
-    if (!scriptResult.isEmpty()) {
-        result = scriptResult;
-    }
-
     // check if width of embedded local images is too high
     const QRegularExpression re(
         QStringLiteral("<img src=\"(file:\\/\\/[^\"]+)\""));
@@ -2467,6 +2459,14 @@ QString Note::textToMarkdownHtml(QString str, const QString &notesPath,
                 QStringLiteral(R"(<img\1src="data:%1;base64,%2")")
                     .arg(type.name(), QString(ba.toBase64())));
         }
+    }
+
+    // check if there is a script that wants to modify the content
+    const QString scriptResult =
+        ScriptingService::instance()->callNoteToMarkdownHtmlHook(this, result);
+
+    if (!scriptResult.isEmpty()) {
+        result = scriptResult;
     }
 
     //    qDebug() << __func__ << " - 'result': " << result;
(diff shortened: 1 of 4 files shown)


r00tr4v3n:
How can I check the reason for the fail? It worked fine for me. Did it just fail on Windows?

pbek:
All builds failed (just click on "Details"). In the case of AppVeyor github wasn't reachable. I restarted all builds to make sure.

r00tr4v3n:
> All builds failed (just click on "Details").

I couldn't find any error message/stack trace... I guess the console output was too long :grimacing: 
So it's not the test case?
```shell
testcases/app/test_utilsmisc.cpp(44) : failure location
PASS   : TestUtilsMisc::testShorten()
PASS   : TestUtilsMisc::testCycleTextCase()
PASS   : TestUtilsMisc::testHtmlToMarkdown()
FAIL!  : TestUtilsMisc::testParseTaskList() 'r1 == expec' returned FALSE. ()
testcases/app/test_utilsmisc.cpp(127) : failure location
PASS   : TestUtilsMisc::testUnescapeHtml()
PASS   : TestUtilsMisc::testHtmlSpecialChars()
PASS   : TestUtilsMisc::testToHumanReadableByteSize()
PASS   : TestUtilsMisc::cleanupTestCase()
Totals: 12 passed, 1 failed, 1 skipped, 0 blacklisted, 2ms
********* Finished testing of TestUtilsMisc *********
[Tests FAIL]
Command exited with code 1
```

pbek:
> FAIL!  : TestUtilsMisc::testParseTaskList() 'r1 == expec' returned FALSE. ()

that's the one you are looking for

src/utils/misc.cpp
```diff
@@ -734,47 +734,54 @@ QString Utils::Misc::parseTaskList(const QString &html, bool clickable) {
...
+                QStringLiteral("\">&#9745;</a>"));
 
-    // TODO
-    // to ensure the clicking behavior of checkboxes,
-    // line numbers of checkboxes in the original markdown text
-    // should be provided by the markdown parser
+        int count = 0;
+        int pos = 0;
+        while (true) {
+            pos = text.indexOf(checkboxStart % QStringLiteral("\""), pos);
```
Waqar144, line 771: can you lift `checkboxStart % QStringLiteral("\"")` out of the loop?

src/entities/note.cpp
```diff
@@ -2277,6 +2277,9 @@ QString Note::textToMarkdownHtml(QString str, const QString &notesPath,
        src=\")((?!file:\/\/).+)\")"),
                 "\\1file://" + windowsSlash + notePath + "/\\2\"");
     */
+    // css classes applicable for export and preview
+    QString additionalClasses = ".task-list-item-checkbox {list-style: none; margin-left: -14px; text-decoration: none;} "
```
Waqar144, line 2281: `additionalClasses` might not be the best name here. How about `listClasses / listCSS` or something along those lines.

Waqar144, line 2281: Additionally, the string literal should be wrapped in `QStringLiteral`

Waqar144:
https://github.com/pbek/QOwnNotes/blob/d8d2669d66ba153635cbe03783fe44120e660396/tests/unit_tests/testcases/app/test_utilsmisc.cpp#L120-L145

This is where its failing. You will need to update the test since `parseTaskList()` has changed now.

src/utils/misc.cpp
```diff
@@ -734,47 +734,54 @@ QString Utils::Misc::parseTaskList(const QString &html, bool clickable) {
 
     // set a list item style for todo items
     // using a css class didn't work because the styling seems to affects the sub-items too
-    const auto listTag = QStringLiteral("<li style=\"list-style-type:square\">");
+    const auto taskListTag = QStringLiteral("<li class=\"task-list-item-checkbox\">");
```
r00tr4v3n, line 737: Isn't the `-checkbox` a little much now?

Waqar144: Could be. @pbek what do you think?

tests/unit_tests/testcases/app/test_utilsmisc.cpp
```diff
@@ -118,29 +118,30 @@ void TestUtilsMisc::testHtmlToMarkdown() {
 }
 
 void TestUtilsMisc::testParseTaskList() {
-    const auto listTag = QStringLiteral("<li style=\"list-style-type:square\">");
+    const auto taskListTag = QStringLiteral("<li class=\"task-list-item-checkbox\">");
+    const auto normalListTag = QStringLiteral("<li class=\"normal-list-item\">");
```
Waqar144, line 122: I wonder what this is doing here if its not going to be used :grin:

r00tr4v3n: Waiting for its moment to shine! Oh well...

src/utils/misc.cpp
pbek: The smaller (yet readable) the better

r00tr4v3n: I'd also turn it around in case we'll think of more ;)
`list-item-task`

pbek:
Any news here? :grin:

r00tr4v3n:
It builds fine now?

pbek:
But it still doesn't work the same for all editor fonts and sizes, right?

pbek:
image

r00tr4v3n:
I cannot reproduce the differently coloured checkboxes.
The font size is tricky:
- the `list-item-checkbox` class seems to overwrite the system setting with a default, so at least this looks fine everywhere
- if somebody changes this class to increase the font though, I have no way of knowing in Qt

So the user would have to change `margin` as well if changing font-size.

pbek:
> I cannot reproduce the differently coloured checkboxes.

That's just the editor schema `Dark` in dark-mode.

> So the user would have to change margin as well if changing font-size.

Who wants to do that? :sweat_smile:

r00tr4v3n:
> That's just the editor schema `Dark` in dark-mode.

Yes, I tried it with several themes:
image

pbek:
> Yes, I tried it with several themes:

Do you mean the checkbox color? I guess that's from `Noto Sans`.

> I will play with the font-size/margin ratio later, maybe this yields some constant results.

You should be able to get the character width of a certain character with something like `const int charWidth = QFontMetrics(EDITOR(font())).width(' ');` where `' '` is the character you want to measure (in our case the checkbox characters).

r00tr4v3n:
Good hint!
Shouldn't this change the default text size in the preview?

image

r00tr4v3n:
The `MainWindow/noteTextEdit.font` also changes the preview font? It looks OK with it `margin_checkbox = width_checkbox+width_bullet/2+1`:
| | |
|---|---|
|image|image|

pbek:
>  Shouldn't this change the default text size in the preview? It does not change.

I've to look into that.

pbek:
I guess you have that setting on:

image

pbek:
For me the text after checkboxes and bullets doesn't align yet:

image

And RTL (right to left) is completely broken:

image

src/entities/note.cpp
```diff
@@ -2277,17 +2277,36 @@ QString Note::textToMarkdownHtml(QString str, const QString &notesPath,
...
-        settings.value(QStringLiteral("MainWindow/noteTextView.code.font"))
+    const QString previewFontString =
+        settings.value(QStringLiteral("MainWindow/noteTextEdit.font"))
             .toString();
+    QString margin_listitem_str = "";
+    QString margin_checkbox_str = "";
+    if (!previewFontString.isEmpty()) {
+        QFont previewFont;
+        previewFont.fromString(previewFontString);
+        const int width_checkbox = QFontMetrics(previewFont).horizontalAdvance("☑"); // &#9744; and &#9745;
```
pbek, line 2288: You also need to take the size of the unchecked box, since the width can be different.

src/entities/note.cpp
```diff
@@ -2277,17 +2277,36 @@ QString Note::textToMarkdownHtml(QString str, const QString &notesPath,
...
+        const int width_bullet = QFontMetrics(previewFont).horizontalAdvance("•"); // &#8226;
```
pbek, line 2289: Are you sure that this bullet is the bullet used in lists?

src/entities/note.cpp
pbek: And can you please use the same variable style as in the rest of the code (camelCase).

src/entities/note.cpp
```diff
@@ -2277,17 +2277,36 @@ QString Note::textToMarkdownHtml(QString str, const QString &notesPath,
...
+        const int width_space = QFontMetrics(previewFont).horizontalAdvance(' ');
+        const int margin_listitem = 2 * width_space;
+        const int margin_checkbox = -width_checkbox - width_bullet / 2 - 1 + margin_listitem;
+        margin_listitem_str = QStringLiteral(" margin-left: %1px; ").arg(margin_listitem);
```
pbek, line 2293: the setting `MainWindow/noteTextView.rtl` will need different handling

src/entities/note.cpp
```diff
@@ -2277,17 +2277,36 @@ QString Note::textToMarkdownHtml(QString str, const QString &notesPath,
        src=\")((?!file:\/\/).+)\")"),
                 "\\1file://" + windowsSlash + notePath + "/\\2\"");
     */
-
-    const QString fontString =
-        settings.value(QStringLiteral("MainWindow/noteTextView.code.font"))
+    const QString previewFontString =
+        settings.value(QStringLiteral("MainWindow/noteTextEdit.font"))
```
pbek, line 2281: If the `MainWindow/noteTextView.useEditorStyles` is false then `MainWindow/noteTextView.font` needs to be used!
And if a user overrides styles with hooks when this also will break. I don't know how to get around that.

src/entities/note.cpp
r00tr4v3n: > I guess you have that setting on:
> 


Yes, that was on! But it does change the font...
image

src/entities/note.cpp
pbek: In 20.6.2 there will be a `Utils::Misc::previewFontString()` that should be used to get the font string.

> And if a user overrides styles with hooks then this also will break. I don't know how to get around that.

But this problem still remains.

pbek: > In 20.6.2 there will be...

You will need to merge the current `develop` branch into your branch to see the changes.

src/entities/note.cpp
pbek: > Yes, that was on! But it does change the font...

see https://github.com/pbek/QOwnNotes/pull/1771/files#r436251486

src/entities/note.cpp
r00tr4v3n: > And RTL (right to left) is completely broken:

How is it supposed to look like? I don't see the appeal unless the bullets would be on the right hand side.
image

pbek: > How is it supposed to look like? I don't see the appeal unless the bullets would be on the right hand side.

Forget what I said about RTL, it looks almost the same in the `develop` branch.

src/entities/note.cpp
r00tr4v3n: Are you sure? That would mean checked/unchecked need different style classes.

src/entities/note.cpp
r00tr4v3n: What do you mean specifically?

r00tr4v3n: > Are you sure that this bullet is the bullet used in lists?

I hope so:
https://www.fileformat.info/info/unicode/char/2022/index.htm

src/entities/note.cpp
pbek: My screenshot says so. They are two different characters. Every character can have different widths.

src/entities/note.cpp
pbek: > What do you mean specifically?

Try https://github.com/pbek/QOwnNotes/pull/1771/files#r436251486 again, you have wrongly "resolved the conversation" (because you smuggled your RTL comments into that conversation), and because of that the html anchor didn't work any more because GitHub hid the conversation.

pbek: For example your `Yes, that was on! But it does change the font...` also doesn't belong to this conversation. So we need to be careful of what we "resolve". :grin:

pbek:
I'll close this until this actually works...

r00tr4v3n:
Sorry, I can't get the alignment right. And even if a formula would work on all fonts and sizes, it would not be very pretty ...

pbek:
It was an up-hill battle 😬

Issue #1794, opened 2020-06-18
unwanted links inside of code blocks


ryandbair:
If you are inside a code block and use syntax that looks like a markdown link, QOwnNotes will turn it into a link which is not expected behavior. 

For instance: 
```
def foo[A](a: A)
```
Turns "A" into a link to "a: A" in the current directory.

pbek:
Confirmed, @Waqar144 anything we can do about that?

Waqar144:
> Confirmed, @Waqar144 anything we can do about that?

It didn't turn into a link for me :thinking:

pbek:
I got `def foo[A](file:///home/omega/ownCloud/Notes.bak/a:%20A)` in the preview...

Waqar144:
> I got `def foo[A](file:///home/omega/ownCloud/Notes.bak/a:%20A)` in the preview...

Oh yes, I didnt look in the preview :laughing:

pbek:
Inline code blocks have the same issue with `def foo[A](a: A)`. :grimacing:

Waqar144:
> Inline code blocks have the same issue with `def foo[A](a: A)`. grimacing

putting a space between `[A]` and `(a: A)` is a temporary solution ;)

pbek:
## 20.6.9
- fixed unwanted `def foo[A](a: A)` note link generation in code blocks in the preview
  (for [#1794](https://github.com/pbek/QOwnNotes/issues/1794), thank you @Waqar144)

pbek:
There now is a new release, could you please test it and report if it works for you?

ryandbair:
Looks great. Thanks!

pbek:
Great, thank you for testing!

PR #1797, opened 2020-06-19
Fix expression in brackets being converted to link in preview

Waqar144:
#1794

### DIFF
--- a/src/helpers/codetohtmlconverter.cpp
+++ b/src/helpers/codetohtmlconverter.cpp
@@ -213,7 +213,7 @@ QString CodeToHtmlConverter::process() const {
                 i--;
             }
         } else {
-            output += _input.at(i);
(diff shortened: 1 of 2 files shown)


pbek:
Tests are failing on Windows. 😁

pbek:
Inline code blocks still have the same issue with `def foo[A](a: A)`. :grimacing:

Waqar144:
> Inline code blocks still have the same issue with `def foo[A](a: A)`. grimacing

Really? I just checked and the issue is fixed on my side :thinking: Did you try it with some other code sample or the same one?

pbek:
image

Waqar144:
hmm those are different things.

The second one is not so easy I am afraid haha. The conversion is being handled by `MD4C` for that.

PR #1805, opened 2020-06-25
Design changes to highlighter and spellchecker class for efficiency

Waqar144:
- Removed some dead code like `setCommentHighlighting`
- Use the spellchecker object directly, no need to make it member variable when we have it available globally
- Removed redundant initialization of spellchecker properties from it's constructor. It now gets initialized from the `QOwnNotesMarkdownEditor::updateSettings()`
- `wordTokenizer` and `languageFilter` are now a part of spellchecker. This gives us one point to control everything related to spellchecking.
- `_currentNote` object is now a ptr.
- The overall size of `QOwnNotesMarkdownHighlighter` class is now 64 bytes from 196 previously. It can be reduced further by making some changes to the base class

### DIFF
--- a/src/dialogs/issueassistantdialog.cpp
+++ b/src/dialogs/issueassistantdialog.cpp
@@ -24,12 +24,6 @@ IssueAssistantDialog::IssueAssistantDialog(QWidget *parent)
     on_issueTypeComboBox_currentIndexChanged(ProblemIssueType);
     ui->logOutputPlainTextEdit->setHighlightingEnabled(false);
 
-    // enable spellchecking
-    ui->expectedBehaviourPlainTextEdit->enableSpellChecker();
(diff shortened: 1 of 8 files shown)


pbek:
Will this PR break highlighting on any platform? :sweat_smile:

Waqar144:
> Will this PR break highlighting on any platform? 😅

I guess we'll find out soon enough 😂

pbek:
Looking good on macOS and Linux. 😁

pbek:
> _currentNote object is now a ptr.

It took me two hours to iron out the crashes this introduced... :sweat_smile: (https://github.com/pbek/QOwnNotes/commit/748608cd01347ce5966c8787183e5a7c94809d56)

PR #1822, opened 2020-07-13
Fix null pointer dereference

shlyakpavel:
I am not sure if checking `!Note::allowDifferentFileName()` before `item->data(0, Qt::UserRole + 1) == FolderType` is a good idea, but it seems to be alright

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -10446,7 +10446,7 @@ void MainWindow::openNotesContextMenu(const QPoint globalPos,
  */
 void MainWindow::on_noteTreeWidget_itemChanged(QTreeWidgetItem *item,
                                                int column) {
-    if (item == nullptr || !Note::allowDifferentFileName()) {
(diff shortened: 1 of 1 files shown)


pbek:
In your PR `on_noteSubFolderTreeWidget_itemChanged(item, column);` will not be called for folder items in the note tree in `Note::allowDifferentFileName()` is `false`. The `Note::allowDifferentFileName()` part needs to be extracted and stay at the old place.

shlyakpavel:
Good. I have changed this. Don't forget to squash commits during merging :)

pbek:
Thank you. 

Code Styöe: Can you please add newlines before each if statement and after the last if statement like it was before. :wink:

shlyakpavel:
I'm starting to hate GitHub web editor
I hope it is fine now :)

pbek:
Thank you!
Yeah, I think I should enforce a linter check as github action. :grin:

PR #1836, opened 2020-07-29
Cache folder

r00tr4v3n:
Added an API call to get the system's cache folder as discussed in https://github.com/qownnotes/scripts/issues/98


pbek:
- Where does `QStandardPaths::CacheLocation` point to in case of QOwnNotes on Linux, macOS and Windows?
- What mechanism invalidates (aka. removes) cached files?
- Is there a reason that a lot of libraries are also touched by this pull request?

@Waqar144, what's your opinion on this scripting command (to give a script a path to write too)?

r00tr4v3n:
It is the same path you are using for qmlcache: 
Linux: ~/.cache/PBE/QOwnNotes/
Mac: "~/Library/Caches/QOwnNotes", "/Library/Caches/QOwnNotes" 
Win: "C:/Users/<USER>/AppData/Local/QOwnNotes/cache"

The script would be responsible for keeping it neat. But since cache files should be generated, it could also be cleaned regularly by QON. I don't see the benefit though.
The reason for the changes in the libraries is probably my new system setup and my lack of understanding for git :disappointed: 
I thought `git fetch upstream && git merge upstream/develop` syncs the repos accordingly before braching. 

The scripts could write anywhere but this way it would be at least bundled in one place.

Waqar144:
I think you forgot to do git submodule update before branching

Waqar144:
> @Waqar144, what's your opinion on this scripting command (to give a script a path to write too)?

I think it's okay, but it's doing two things, (creating a Dir in case it doesn't exist). This should be clarified in the doc.

pbek:
> The script would be responsible for keeping it neat. But since cache files should be generated, it could also be cleaned regularly by QON. I don't see the benefit though.

If scripts are supposed to write to that path there should at least be a script method to clear the cache directory used

> The reason for the changes in the libraries is probably my new system setup and my lack of understanding for git disappointed

Yeah, as @Waqar144 said... It guess it would be safest to create a new pull requests (or try to update your sub-modules and commit the changes if that works)

PR #1849, opened 2020-08-08
Add option to disable note count with tags

Waqar144:
Add option to disable counting notes for every tag. This increases performance by quite a margin when reloading the tag tree.
 - Count will still be visible for "All Notes" and "Untagged Notes"

#943

### DIFF
--- a/src/dialogs/settingsdialog.ui
+++ b/src/dialogs/settingsdialog.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>800</width>
-    <height>600</height>
+    <width>1078</width>
+    <height>1194</height>
    </rect>
   </property>
   <property name="sizePolicy">
@@ -397,7 +397,7 @@
          <item>
           <widget class="QStackedWidget" name="settingsStackedWidget">
            <property name="currentIndex">
-            <number>1</number>
+            <number>16</number>
            </property>
            <widget class="QWidget" name="noteFoldersPage">
             <layout class="QGridLayout" name="gridLayout_19">
@@ -979,6 +979,7 @@
                    <font>
                     <weight>50</weight>
                     <bold>false</bold>
+                    <kerning>true</kerning>
                    </font>
                   </property>
                   <property name="styleSheet">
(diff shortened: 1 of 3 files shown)


src/dialogs/settingsdialog.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>800</width>
-    <height>600</height>
+    <width>1078</width>
```
pbek, line 9: Can yo please revert those size changes

src/dialogs/settingsdialog.ui
```diff
@@ -979,6 +979,7 @@
                    <font>
                     <weight>50</weight>
                     <bold>false</bold>
+                    <kerning>true</kerning>
```
pbek, line 982: Did you add this on purpose?

src/dialogs/settingsdialog.ui
```diff
@@ -3468,8 +3469,8 @@
                           <rect>
                            <x>0</x>
                            <y>0</y>
-                           <width>430</width>
-                           <height>538</height>
+                           <width>310</width>
```
pbek, line 3472: Can yo please revert those size changes

src/dialogs/settingsdialog.ui
```diff
@@ -4628,7 +4629,7 @@ Just test yourself if you get sync conflicts and set a higher value if so.</stri
                 <string notr="true">&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
 &lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
 p, li { white-space: pre-wrap; }
-&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'Noto Sans'; font-size:14pt; font-weight:400; font-style:normal;&quot;&gt;
+&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'Noto Sans'; font-size:10pt; font-weight:400; font-style:normal;&quot;&gt;
```
pbek, line 4632: Can you please revert this size change

src/dialogs/settingsdialog.ui
Waqar144: No, I think it is because my system has these things as defaults? Or some other reason maybe.

pbek: Then please remove it as well. 😁

Issue #1875, opened 2020-09-19
Tab key generates tab character in Vim mode even if the "Use tab character" option is disabled


johnliu55tw:
#### Expected behaviour
In Vim mode, with the option "Use tab character instead of spaces" disabled, when I press tab key, it should generate spaces instead of tab character.

#### Actual behaviour
Pressing tab key always generates tab character in Vim mode, even when "Use tab character instead of spaces" option is disabled.

#### Steps to reproduce
1. Enable Vim mode
2. Make sure "Use tab character instead of spaces" is disabled
3. Go to editor
4. Press tab key, and it generates tab character

Thanks for the great app!

pbek:
> Thanks for the great app!

Thank you for your kind words.

> In Vim mode, with the option "Use tab character instead of spaces" disabled, when I press tab key, it should generate spaces instead of tab character.

@Waqar144, are you using the vim mode?

Waqar144:
> > Thanks for the great app!
> 
> Thank you for your kind words.
> 
> > In Vim mode, with the option "Use tab character instead of spaces" disabled, when I press tab key, it should generate spaces instead of tab character.
> 
> @Waqar144, are you using the vim mode?

Yes, I use vim mode. I think we need to set "expandTab" setting somehow.

pbek:
## 20.9.9
- tab stop width and tab versus space settings are now also respected in fakevim mode
  (for [#1875](https://github.com/pbek/QOwnNotes/issues/1875), thank you @Waqar144)

pbek:
There now is a new release, could you please test it and report if it works for you?

johnliu55tw:
Thanks for the quick release! However the issue still exists.

BTW I'm using AUR built version, not sure if that's related:
![image](https://user-images.githubusercontent.com/5047724/93956397-af7df780-fd84-11ea-8221-4214fce053b0.png)

Please let me know if I need to provide more information.

Waqar144:
> Thanks for the quick release! However the issue still exists.

Yes, that's on me. It will work if you enable "Use Tabs" :D.
Will fix in the next release.

pbek:
## 20.9.10
- fixed the incorrect setExpandTab setting for fakeVim (for [#1875](https://github.com/pbek/QOwnNotes/issues/1875), thank you @Waqar144)

pbek:
There now is a new release, could you please test it and report if it works for you?

johnliu55tw:
[...]

PR #1876, opened 2020-09-20
Respect tab stop width and tab vs space in fakevim mode

Waqar144:
https://github.com/pbek/QOwnNotes/issues/1875

I have tried very hard to fix "Ctrl + V", it deactivates in fakeVim correctly but for some reason the text never gets pasted. It is probably because of some shortcut conflict or some other thing. Not sure. Any ideas?

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -604,12 +604,8 @@ void MainWindow::initFakeVim(QOwnNotesMarkdownTextEdit *noteTextEdit) {
     auto proxy = new FakeVimProxy(noteTextEdit, this, handler);
 
     QSettings settings;
-    if (settings.value(QStringLiteral("Editor/useTabIndent")).toBool()) {
-        FakeVim::Internal::theFakeVimSettings()->item("et")->setValue(true);
-    } else {
-        FakeVim::Internal::theFakeVimSettings()->item("et")->setValue(true);
-    }
-
+    bool setExpandTab = settings.value(QStringLiteral("Editor/useTabIndent")).toBool();
+    FakeVim::Internal::theFakeVimSettings()->item("et")->setValue(setExpandTab);
     auto width = settings.value("Editor/indentSize", 4).toInt();
     FakeVim::Internal::theFakeVimSettings()->item("ts")->setValue(width);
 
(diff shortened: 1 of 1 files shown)


src/mainwindow.cpp
```diff
@@ -602,6 +603,16 @@ void MainWindow::initFakeVim(QOwnNotesMarkdownTextEdit *noteTextEdit) {
 
     auto proxy = new FakeVimProxy(noteTextEdit, this, handler);
 
+    QSettings settings;
+    if (settings.value(QStringLiteral("Editor/useTabIndent")).toBool()) {
+        FakeVim::Internal::theFakeVimSettings()->item("et")->setValue(true);
```
pbek, line 608: Can't you use `FakeVim::Internal::theFakeVimSettings()->item("et")->setValue(VARIABLE_HERE);` to get around code duplication?

pbek:
> I have tried very hard to fix "Ctrl + V", it deactivates in fakeVim correctly but for some reason the text never gets pasted. 

Maybe https://github.com/pbek/qmarkdowntextedit/blob/dbccd4a2afda9f12a463269a8c6768076834d254/qmarkdowntextedit.cpp#L252-L261 ?

src/mainwindow.cpp
pbek: Wait, what are those two lines of `FakeVim::Internal::theFakeVimSettings()->item("et")->setValue(true);` doing? 😆

pbek: Both times true?

Waqar144: > Can't you use `FakeVim::Internal::theFakeVimSettings()->item("et")->setValue(VARIABLE_HERE);` to get around code duplication?

Yes, I guess

Waqar144: > Wait, what are those two lines of `FakeVim::Internal::theFakeVimSettings()->item("et")->setValue(true);` doing? laughing

AH yes, one needs to be false.

Waqar144:
> > I have tried very hard to fix "Ctrl + V", it deactivates in fakeVim correctly but for some reason the text never gets pasted.
> 
> Maybe https://github.com/pbek/qmarkdowntextedit/blob/dbccd4a2afda9f12a463269a8c6768076834d254/qmarkdowntextedit.cpp#L252-L261 ?

Tested that already, the problem seems to be elsewhere :thinking:

PR #1881, opened 2020-09-23
Fix query in fetchAllIdsByParentId

Waqar144:


pbek:
There is a merge conflict with my commit. You can use your version without the parameter (I also fixed fetchAllByParentId)

PR #1882, opened 2020-09-23
Optimize Notesubfolder::relativepath for the common case

Waqar144:


src/entities/notesubfolder.cpp
```diff
@@ -60,6 +60,7 @@ NoteSubFolder NoteSubFolder::fetch(int id) {
     return NoteSubFolder();
 }
 
+
```
pbek, line 63: Do you need that newline?

Waqar144: Not really :D
artifacts from refactoring

PR #1886, opened 2020-09-24
Add fetchNoteIdByName to fetch only note Id

Waqar144:
#943

With this commit almost all of the lag with around 3000 tags is gone. (There will still be lag if multiple folders are selected, but that is *rather difficult* to get around because we have a large amount of data to work with)

### DIFF
--- a/src/entities/tag.cpp
+++ b/src/entities/tag.cpp
@@ -725,6 +725,7 @@ QVector<int> Tag::fetchAllLinkedNoteIds(int tagId, const bool fromAllSubfolders,
             int noteId = Note::fetchNoteIdByName(name, noteSubFolderId);
             noteIdList.append(noteId);
         }
+        return noteIdList;
     }
 
     DatabaseService::closeDatabaseConnection(db, query);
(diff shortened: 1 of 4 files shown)


src/entities/tag.cpp
```diff
@@ -712,23 +711,26 @@ QVector<int> Tag::fetchAllLinkedNoteIds(int tagId, const bool fromAllSubfolders,
...
-                                                 QStringLiteral("/"));
-
-            noteIdList.append(note.getId());
+            int noteSubFolderId =
+                NoteSubFolder::fetchByPathData(noteSubFolderPathData, QStringLiteral("/"))
+                    .getId();
+            int noteId = Note::fetchNoteIdByName(name, noteSubFolderId);
+            noteIdList.append(noteId);
         }
+        return noteIdList;
```
pbek, line 728: If you return here then `DatabaseService::closeDatabaseConnection` will not be called!

pbek: I'll fix that.

Waqar144: @pbek does it really matter? The db is local and is always accessed from a single thread

pbek: It "should" not matter, but it did on Windows. 😬

pbek: see: `DatabaseService::closeDatabaseConnection`

PR #1909, opened 2020-10-19
Fix 'All Notes' count showing count for previous selected item

Waqar144:
#1903

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -7857,10 +7857,14 @@ void MainWindow::reloadTagTree() {
 
     QVector<int> noteSubFolderIds;
 
-    const auto noteSubFolderWidgetItems =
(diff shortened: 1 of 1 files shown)


src/mainwindow.cpp
```diff
@@ -7857,10 +7857,14 @@ void MainWindow::reloadTagTree() {
 
     QVector<int> noteSubFolderIds;
 
-    const auto noteSubFolderWidgetItems =
-        ui->noteSubFolderTreeWidget->selectedItems();
-    noteSubFolderIds.reserve(noteSubFolderWidgetItems.count());
+    auto noteSubFolderWidgetItems = ui->noteSubFolderTreeWidget->selectedItems();
+    //if only one item is selected, then take current Item otherwise we will get the item that
```
pbek, line 7861: Can you please use the same comment style like in the other lines? 😁

PR #1910, opened 2020-10-19
Dont reloadTagTree multiple times during startup

Waqar144:
`reloadTagTree()` is called **3** times during startup.

# 1
```bt
1 MainWindow::reloadTagTree                           mainwindow.cpp 7858 0x55555562ba97 
2 MainWindow::setupTags                               mainwindow.cpp 8364 0x55555562f8a4 
3 MainWindow::loadNoteDirectoryList                   mainwindow.cpp 2148 0x555555604fbc 
4 MainWindow::buildNotesIndexAndLoadNoteDirectoryList mainwindow.cpp 1031 0x5555555fd0e3 
5 MainWindow::MainWindow                              mainwindow.cpp 290  0x5555555f7111 
6 main                                                main.cpp       616  0x5555555df24a 
```

# 2

```bt
1 MainWindow::reloadTagTree             mainwindow.cpp 7858  0x55555562ba97 
2 MainWindow::setupTags                 mainwindow.cpp 8364  0x55555562f8a4 
3 MainWindow::loadNoteDirectoryList     mainwindow.cpp 2148  0x555555604fbc 
4 MainWindow::updateNotesPanelSortOrder mainwindow.cpp 11962 0x555555647330 
5 MainWindow::updatePanelsSortOrder     mainwindow.cpp 11944 0x5555556470ae 
6 MainWindow::MainWindow                mainwindow.cpp 399   0x5555555f7e4e 
7 main                                  main.cpp       616   0x5555555df24a 
```

# 3

```bt
1 MainWindow::reloadTagTree             mainwindow.cpp 7858  0x55555562ba97 
5 MainWindow::updatePanelsSortOrder     mainwindow.cpp 11944 0x5555556470ae 
6 MainWindow::MainWindow                mainwindow.cpp 399   0x5555555f7e4e 
7 main                                  main.cpp       616   0x5555555df24a 
```

With this commit we are down to two.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -11943,6 +11943,9 @@ Qt::SortOrder MainWindow::toQtOrder(int order) {
 void MainWindow::updatePanelsSortOrder() {
     updateNotesPanelSortOrder();
     reloadNoteSubFolderTree();
+    // do not reload it again, it has already been reloaded when
+    // updateNotesPanelSortOrder() was called
+    //reloadTagTree();
 }
 
 void MainWindow::updateNotesPanelSortOrder() {
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -11943,7 +11943,6 @@ Qt::SortOrder MainWindow::toQtOrder(int order) {
 void MainWindow::updatePanelsSortOrder() {
     updateNotesPanelSortOrder();
     reloadNoteSubFolderTree();
-    reloadTagTree();
 }
 
 void MainWindow::updateNotesPanelSortOrder() {


src/mainwindow.cpp
```diff
@@ -11943,7 +11943,6 @@ Qt::SortOrder MainWindow::toQtOrder(int order) {
 void MainWindow::updatePanelsSortOrder() {
     updateNotesPanelSortOrder();
     reloadNoteSubFolderTree();
-    reloadTagTree();
```
pbek, line 11946: And how how will the tag tree be reloaded after updating the panel sort order?

Waqar144: tag tree is already reloaded once after panel sort order is updated, isn't it?

pbek: You mean actually in that method? I don't think so. Is it?

Waqar144: In `updateNotesPanelSortOrder()`, `loadNoteDirectoryList()` gets called, which calls `setupTags` and tag tree is reloaded. See Call stack 2 in OP.

Waqar144: **1** already does almost everything. **2** seems redundant except for the function `updateNotesPanelSortOrder` which basically updates some checkboxes I think. (Unless I am missing something here)

pbek: Looks ok. thank you. But can you please instead of removing the method call comment it out and add a comment explaining that you just said just make clear what we need to achieve updatePanelsSortOrder().

Waqar144: Done

Issue #1888, opened 2020-09-26
[F] Enable drag and drop within Note Edit Panel


alexanderino:
#### Expected behaviour

With virtually all text editors and word processors, selecting text and dragging it to another location in the document moves it there.

#### Actual behaviour

With QOwnNotes [as of `20.9.11`], it is not possible to move selected text by dragging it to another location within the Note Edit Panel. (It is possible to drop it into another program like Notepad or LibreOffice Writer.)

The workaround is to cut and paste, but this clears the clipboard memory, which may not be always desirable.

#### Steps to reproduce

- select text with mouse (or equivalent device)
- attempt to drag and drop it to another part of the note within the Node Edit Panel — this fails
    - it is worth noting that my mouse's cursor turns into a circle with a slash, indicating that it won't work

This behaviour is consistent on Windows 10 and 8.1 (both Pro, 64-bit).

pbek:
Thank you for your suggestion, I'll look into it on how much effort that would cost. 😉  (by the way, I use clipboard managers everywhere 😁 )

alexanderino:
Thank you, Patrizio :slightly_smiling_face: I appreciate you investigating this. I'll also take the opportunity to thank you for creating this wonderful program.

Warm regards,

Jason

pbek:
Thank you for your kind words! 😃

pbek:
## 20.10.7
- it is now also possible to move text in the note text edit via drag and drop
  (for [#1888](https://github.com/pbek/QOwnNotes/issues/1888), thank you @fnkabit)

pbek:
There now is a new release, could you please test it and report if it works for you?

alexanderino:
> There now is a new release, could you please test it and report if it works for you?

Hi, Patrizio. I've been dragging and dropping text in my existing MD files (including an encrypted note), and there has not been one problem or issue — it works like it should. This improves ease of use *tremendously* for me.

This issue can be considered closed from my end.

Thank you so much!

pbek:
Great, thank you for testing!

PR #1917, opened 2020-10-23
Allow text drag and drop in note edit panel

fnkabit:
Feature #1888 
Video demo: https://youtu.be/UHil5b90OWI

### DIFF
--- a/src/widgets/qownnotesmarkdowntextedit.cpp
+++ b/src/widgets/qownnotesmarkdowntextedit.cpp
@@ -408,6 +408,11 @@ void QOwnNotesMarkdownTextEdit::setMainWindow(MainWindow *mainWindow) {
     this->mainWindow = mainWindow;
 }
 
+bool QOwnNotesMarkdownTextEdit::canInsertFromMimeData(const QMimeData *source) const
+{
+    return (!source->hasUrls());
+}
+
 /**
  * Handles pasting from clipboard
  */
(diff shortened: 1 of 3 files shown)


src/widgets/qownnotesmarkdowntextedit.cpp
```diff
@@ -408,6 +408,11 @@ void QOwnNotesMarkdownTextEdit::setMainWindow(MainWindow *mainWindow) {
     this->mainWindow = mainWindow;
 }
 
+bool QOwnNotesMarkdownTextEdit::canInsertFromMimeData(const QMimeData *source) const
+{
+    return (!source->hasUrls());
```
fnkabit, line 413: text/uri-list mime-types will be handled in MainWindow

fnkabit:
@pbek Tested on Linux only; if you think it should be tested on Windows as well, I could set up an Windows VM tomorrow.

pbek:
Awesome, thank you very much!

> @pbek Tested on Linux only; if you think it should be tested on Windows as well, I could set up an Windows VM tomorrow.

It will not break very important stuff, we can fix that afterwards if there really is a problem. 😁

fnkabit:
That was fast :) Thank you for the review @pbek .
Have a great weekend !

pbek:
Thanks a lot for your contributions! (I'm still working on webpage automation 😁). Have a great weekend yourself!

alexanderino:
I can confirm that drag and drop works perfectly on Windows 8.1 Pro x64. This makes a huge difference to me, so thanks a lot, @fnkabit !

Issue #1698, opened 2020-04-19
Allow creation of note by clicking on non-legacy note-link if link is broken


pbek:
It would be nice if there was a way to allow the creation of the note if a non-legacy note link is broken (like it works with `note://note-name` links in #736).

Of course this is a lot more complex.

Links like `[Deep-Note](a%20sub%20folder%202%2Fdfolder%202%2FDeep-Note2.md)` are currently (rightfully) detected as external files if the linked note doesn't exist in the current note folder (detected in `Note::fileUrlIsNoteInCurrentNoteFolder`).

And beside the note name of course the correct sub-folder also needs to be detected (maybe even optionally created?). Extra warnings are needed if sub-folders are disabled for the current note-folder.

Waqar144:
Where will be the note created? you mean new note?

pbek:
Yes, the new note would (should) be created in the sub-folder the link points to.

pbek:
@fnkabit, that would also be a nice feature in case you are looking for more. 😁

fnkabit:
@pbek Thanks! Will look into this one next.

pbek:
Hehe, awesome! 🎉  I hoped you would "bite". 😀

pbek:
## 20.10.8
- you can now choose to create the note of broken note links when you <kbd>Ctrl</kbd> + Click
  on the note in the note text edit (for [#1698](https://github.com/pbek/QOwnNotes/issues/1698), thank you @fnkabit)

pbek:
## 20.10.9
- you can now also choose to create the note of broken note links when clicking
  on the note in the note preview (for [#1698](https://github.com/pbek/QOwnNotes/issues/1698), thank you @fnkabit)
    - if note subfolders are disabled for your note folder you will now also get
      a proper error message when trying to create notes in subfolders

PR #1918, opened 2020-10-25
Allow creating note by Ctrl clicking on non-legacy broken note link

fnkabit:
Feature #1698 
Video demo: https://www.youtube.com/watch?v=UFSM_hf7vh0

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -6133,7 +6133,7 @@ void MainWindow::on_noteTextView_anchorClicked(const QUrl &url) {
          scheme == QStringLiteral("task") ||
          scheme == QStringLiteral("checkbox")) ||
         (scheme == QStringLiteral("file") &&
-         Note::fileUrlIsNoteInCurrentNoteFolder(url))) {
+         Note::fileUrlIsExistingNoteInCurrentNoteFolder(url))) {
         openLocalUrl(url.toString());
     } else {
         ui->noteTextEdit->openUrl(url.toString());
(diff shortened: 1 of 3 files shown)


fnkabit:
Should also add handling of case where sub-folders are not allowed.
In this case, would a warning message be enough (and automatically aborting operation) ?

pbek:
Awesome, I'll test that today!

> Should also add handling of case where sub-folders are not allowed.
> In this case, would a warning message be enough (and automatically aborting operation) ?

Yes, that would be great. A warning message what note sub-folders are turned off for this note folder and aborting seems appropriate.

pbek:
I now tested your PR, most cases already work! Thank you very much.

One case that didn't work was when I have in a sub-folder and add a link like `[Broken Link](../Broken-Note.md)`. I get the question if I want to create the note in the root folder but when I say `Yes` the note is created in the sub-folder instead of the root.

And it would be great if your mechanism would also work when you click on the note link in the preview.

pbek:
I merged it now to let others test too. If you care, the clicking on the note link in the preview would also really be great.

fnkabit:
Thank you Patrizio. I will try to reproduce the scenario that you mentioned.
In the next PR I will add the warning for when the subfolders are not allowed and also implement this same logic for note links in the preview.

pbek:
Thank you very much.

> Thank you Patrizio. I will try to reproduce the scenario that you mentioned.

Like I said, I couldn't reproduce it any more. 😁

pbek:
Two regressions so far. Happy days. 😆

src/mainwindow.cpp
```diff
@@ -6223,6 +6224,7 @@ void MainWindow::openLocalUrl(QString urlString) {
         Note note;
 
         if (isNoteFileUrl) {
+            ui->noteSubFolderTreeWidget->reset();
```
pbek, line 6227: @fnkabit, do you by change remember why you added this line? It causes #2128.

fnkabit: @pbek Ah..sorry, don't remember.

pbek: No worries, I removed the line for now. 😁

Issue #1165, opened 2019-03-13
Feature Request: Import/Export Vivaldi .JSON notes


RealDrJester:
Hi! 
I would like to request the possibility of importing the .json notes from Vivaldi(browser), following the folder or hierarchy it has. For full disclosure, I've asked this request to Joplin, Zim, and Cherrytree about a week ago. 

I've tried the python script available in Vivaldi's forum, but it only exports it plainly to one single .txt file, which makes it unusable anywhere. Another issue is that it doesn't export the folders, or the hierarchy. 

Thank you, in advance, for any help or consideration.

pbek:
You should be able to do that with any scripting language, since QOwnNotes just uses simple markdown files for notes. Or maybe someone has done that before, https://duckduckgo.com/?q=vivaldi+json+to+markdown+files&t=ffab&ia=web

RealDrJester:
Thanks for another venue to explore. I've given a look on scripting, before posting this, but I'm afraid it is beyond my skills of barely newbie level in programming(hello world level here).

fnkabit:
Will add export ASAP.

fnkabit:
Created a PR for the note export functionality.

The notes export functionality will generate an export file that should replace the Vivaldi Notes file (on Linux located in **.config/vivaldi/Default/Notes**).
Please make sure to **close all instances of Vivaldi** and **backup your Vivaldi Notes file** first (just to make sure your notes are not lost, if sync is not enabled).

Normally, **if sync is enabled**, no notes are lost when replacing the Vivaldi Notes file with the file generated by this script. This is due to the fact that your Vivaldi Notes will be synced back from the cloud as soon as you start Vivaldi.

RealDrJester:
Just gave it a go again. The problem is still there. The notes are still being stored in one folder, and all the folders below the "B" on the screenshot were located in other places, or were subfolders from other folders. 
![image](https://user-images.githubusercontent.com/3292341/97820210-d7b61b80-1c8b-11eb-9845-ceca0d39390b.png)

fnkabit:
Hello @RealDrJester ,
Tried to reproduce the issue with the folder structure that you previously provided. Sadly, I wasn't able to reproduce the issue..
Video demo: https://youtu.be/XJLvYqv-Zo0

If you manage to find a simple scenario to reproduce this issue, please let me know and I will fix it ASAP.

RealDrJester:
[...]

PR #1925, opened 2020-10-28
Add ScriptingService method for creating subfolders

fnkabit:
I am currently working on #1165 and it would be required to create folders when importing notes.

### DIFF
--- a/src/services/scriptingservice.cpp
+++ b/src/services/scriptingservice.cpp
@@ -1501,6 +1501,27 @@ void ScriptingService::createNote(QString text) {
 }
 
 /**
+ * Creates a new subfolder
+ *
+ * @param folderName the subfolder name
+ */
+void ScriptingService::createSubFolder(QString folderName)
+{
+#ifndef INTEGRATION_TESTS
+    MainWindow *mainWindow = MainWindow::instance();
+
+    if (mainWindow != Q_NULLPTR) {
+        MetricsService::instance()->sendVisitIfEnabled(
+            QStringLiteral("scripting/") % QString(__func__));
+
+        mainWindow->createNewNoteSubFolder(std::move(folderName));
+    }
+#else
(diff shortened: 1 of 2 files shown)


fnkabit:
@pbek Is there something wrong with the **ubuntu-latest 5.7 env** ?
```
2020-10-28 07:34:01,046 - aqt - ERROR - Download error: Symlink loop from '/home/runner/work/QOwnNotes/Qt/5.7/gcc_64/lib/libQt53DCore.so.5.7.1'
2020-10-28 07:34:02,456 - aqt - ERROR - Download error: [Errno 40] Too many levels of symbolic link
```

pbek:
I restarted the jobs again.

pbek:
I just saw `createNewNoteSubFolder` in `mainwindow` is already `Q_INVOKABLE`, so there is no need for a new scripting method.
It's also documented in https://www.qownnotes.org/scripting/classes.html#properties-and-methods-3.

fnkabit:
Ah, sorry. Had the impression that all scripting methods should be available in `ScriptingService`. Will close PR.

pbek:
> Had the impression that all scripting methods should be available in ScriptingService

It would be better to not expose more private API, but after writing too many wrapper for mainwindow methods I started to expose them directly 😁 since you have to mark those methods with `Q_INVOKABLE` it is at least clear that they are exposed...

pbek:
If you want you can add a `::: tip` at https://www.qownnotes.org/scripting/methods-and-objects.html#jumping-to-a-note-subfolder telling people how to also create note subfolders.

fnkabit:
> If you want you can add a `::: tip` at https://www.qownnotes.org/scripting/methods-and-objects.html#jumping-to-a-note-subfolder telling people how to also create note subfolders.

Good idea ! :+1: Thanks.

PR #1929, opened 2020-10-29
Add additional ScriptingService methods

fnkabit:
These methods would be required in order to be able to export all notes and subfolders ( required for feature #1165 )

### DIFF
--- a/src/entities/notesubfolder.cpp
+++ b/src/entities/notesubfolder.cpp
@@ -113,31 +113,6 @@ QString NoteSubFolder::pathData() const {
 }
 
 /**
- * Fetches a note sub folder by its id
- */
-NoteSubFolder NoteSubFolder::fetchById(int id)
-{
-    const QSqlDatabase db = QSqlDatabase::database(QStringLiteral("memory"));
-    QSqlQuery query(db);
-
-    QVector<int> idList;
-    QString sql = QStringLiteral("SELECT * FROM noteSubFolder where id = :id");
-
-    query.prepare(sql);
-    query.bindValue(QStringLiteral(":id"), id);
-
-    if (!query.exec()) {
-        qWarning() << __func__ << ": " << query.lastError();
-    } else {
-        if (query.first()) {
-            return noteSubFolderFromQuery(query);
-        }
-    }
-
-    return NoteSubFolder();
-}
-
-/**
  * Fetches a note sub folder by its path data
  */
 NoteSubFolder NoteSubFolder::fetchByPathData(QString pathData,
--- a/src/entities/notesubfolder.cpp
+++ b/src/entities/notesubfolder.cpp
@@ -113,6 +113,31 @@ QString NoteSubFolder::pathData() const {
 }
 
 /**
+ * Fetches a note sub folder by its id
+ */
+NoteSubFolder NoteSubFolder::fetchById(int id)
+{
+    const QSqlDatabase db = QSqlDatabase::database(QStringLiteral("memory"));
+    QSqlQuery query(db);
+
+    QVector<int> idList;
+    QString sql = QStringLiteral("SELECT * FROM noteSubFolder where id = :id");
+
+    query.prepare(sql);
+    query.bindValue(QStringLiteral(":id"), id);
+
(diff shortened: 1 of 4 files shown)


src/entities/notesubfolder.cpp
```diff
@@ -112,6 +112,31 @@ QString NoteSubFolder::pathData() const {
                           : getParent().pathData() + QChar('\n') + _name;
 }
 
+/**
+ * Fetches a note sub folder by its id
+ */
+NoteSubFolder NoteSubFolder::fetchById(int id)
```
pbek, line 118: In what way does this method differ from `NoteSubFolder::fetch`?

src/services/scriptingservice.cpp
```diff
@@ -2012,6 +2012,56 @@ QVariant ScriptingService::getApplicationSettingsVariable(
...
+QList<int> ScriptingService::fetchSubFoldersByParentId(int parentId)
+{
+    MetricsService::instance()->sendVisitIfEnabled(
+        QStringLiteral("scripting/") % QString(__func__));
+
+    QList<int> subFolderIds;
+
+    const auto subFolders = NoteSubFolder::fetchAllByParentId(parentId);
+    for (const auto &subFolder : subFolders) {
+        subFolderIds.push_back(subFolder.getId());
```
pbek, line 2030: Please use `append`, `push_back` is only an alias

src/services/scriptingservice.h
```diff
@@ -172,6 +172,12 @@ class ScriptingService : public QObject {
     Q_INVOKABLE QVariant getApplicationSettingsVariable(
         const QString &key, const QVariant &defaultValue = QVariant());
 
+    Q_INVOKABLE QList<int> fetchSubFoldersByParentId(int parentId);
```
pbek, line 175: Please always use `NoteSubFolder` instead of `SubFolder` and we need documentation for those new methods.

src/entities/notesubfolder.cpp
fnkabit: @pbek Once again, I didn't see it :smile:

src/services/scriptingservice.h
```diff
@@ -172,6 +172,12 @@ class ScriptingService : public QObject {
...
+
+    Q_INVOKABLE QString getSubFoldersName(int id);
```
pbek, line 177: Why is this called `SubFolders`, instead of `SubFolder`?

fnkabit: Typo

src/services/scriptingservice.cpp
```diff
@@ -2012,6 +2012,56 @@ QVariant ScriptingService::getApplicationSettingsVariable(
     return settings.value(key, defaultValue);
 }
 
+/**
+ * Fetches all subfolder ids for subfolders that are children of parentId
+ *
+ * @param parentId {int} id of parent subfolder
+ * @return QList<int> list of subfolder ids
+ */
+QList<int> ScriptingService::fetchSubFoldersByParentId(int parentId)
```
pbek, line 2021: This actually doesn't return NoteSubFolders, but only ids, doesn't it?
I generally would prefer it more if there actually would be `NoteSubFolderApi` class (like for notes and tags) with proper methods that return actual objects.

fnkabit: Asked you regarding this on IRC, written to @pbek (probably not the correct username?) :)

fnkabit: Ok, @pbek so we should drop this review and implement the `NoteSubFolderApi` instead, is that right ?

fnkabit:
Updated code-review.

src/services/scriptingservice.cpp
pbek: Yes, please.

pbek: > Asked you regarding this on IRC, written to @pbek (probably not the correct username?) :)

It is, but I didn't get any notifications. 😬  In the `#qownnotes` channel your messages would have been bridged to Telegram.

PR #1936, opened 2020-11-04
Disable 'distraction free mode' when no note is selected

fnkabit:
Otherwise you would get a blank screen when activating 'distraction free mode' while no note is selected.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -3841,6 +3841,8 @@ void MainWindow::setCurrentNote(Note note, bool updateNoteText,
 
     // clear external image cache
     Note::externalImageHash()->clear();
+
+    ui->actionToggle_distraction_free_mode->setEnabled(true);
 }
 
 void MainWindow::updateCurrentTabData(const Note &note) const {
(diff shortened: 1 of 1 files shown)


src/mainwindow.cpp
```diff
@@ -3841,6 +3841,8 @@ void MainWindow::setCurrentNote(Note note, bool updateNoteText,
 
     // clear external image cache
     Note::externalImageHash()->clear();
+
+    ui->actionToggle_distraction_free_mode->setEnabled(true);
```
pbek, line 3845: Why do you want to turn on the distraction free mode every time the note changes (even if not in distraction free mode)?

src/mainwindow.cpp
```diff
@@ -4810,6 +4812,8 @@ void MainWindow::unsetCurrentNote() {
     ui->encryptedNoteTextEdit->hide();
     ui->encryptedNoteTextEdit->clear();
 
+    ui->actionToggle_distraction_free_mode->setEnabled(false);
```
pbek, line 4815: That looks reasonable, thank you.

src/mainwindow.cpp
pbek: Ah, sorry. you enabled the checkbox!

PR #1955, opened 2020-11-18
Do not display subfolder actions when right clicking empty space in note list panel

fnkabit:
Hides subfolder context menu actions in note list panel when subfolders are not used for the current folder.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -10360,12 +10360,17 @@ void MainWindow::openCurrentNoteInTab() {
 void MainWindow::on_noteTreeWidget_customContextMenuRequested(
     const QPoint pos) {
     auto *item = ui->noteTreeWidget->itemAt(pos);
+
+    // if the user clicks at empty space, this is null and if it isn't handled
+    // QON crashes
+    if (item == nullptr) {
+        return;
+    }
+
     const QPoint globalPos = ui->noteTreeWidget->mapToGlobal(pos);
     const int type = item == nullptr ?
          0 : item->data(0, Qt::UserRole + 1).toInt();
 
-    // if the user clicks at empty space, this is null and if it isn't handled
-    // QON crashes
     if (item == nullptr || type == FolderType) {
         openNoteSubFolderContextMenu(globalPos, ui->noteTreeWidget);
     } else if (type == NoteType) {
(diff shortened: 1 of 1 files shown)


pbek:
In what scenario would one see the NoteSubFolderContextMenu in the note list?

pbek:
If one uses the work-in-progress "show sub-folders in note list" setting?

fnkabit:
> In what scenario would one see the NoteSubFolderContextMenu in the note list?

If 'Use note subfolders' is checked for the current note.

pbek:
But `openNoteSubFolderContextMenu` is called for the note sub folder panel, not the note list.
Do you have a screenshot showing what we are talking about?

fnkabit:
image

pbek:
Oh, it seem the `openNoteSubFolderContextMenu` is opened in the note list if you click in an empty area! That looks like a regression from the "show sub-folders in note list" setting. 😬

fnkabit:
> Oh, it seem the `openNoteSubFolderContextMenu` is opened in the note list if you click in an empty area! That looks like a regression from the "show sub-folders in note list" setting.

I thought it was intended to be this way.

```
void MainWindow::on_noteTreeWidget_customContextMenuRequested(
    const QPoint pos) {
    auto *item = ui->noteTreeWidget->itemAt(pos);
    const QPoint globalPos = ui->noteTreeWidget->mapToGlobal(pos);
    const int type = item == nullptr ?
         0 : item->data(0, Qt::UserRole + 1).toInt();

    // if the user clicks at empty space, this is null and if it isn't handled
    // QON crashes
    if (item == nullptr || type == FolderType) {
        openNoteSubFolderContextMenu(globalPos, ui->noteTreeWidget);
    } else if (type == NoteType) {
        openNotesContextMenu(globalPos);
    }
}

What should the correct behaviour be, when clicking on an empty space in the note list panel ?
```

pbek:
> What should the correct behaviour be, when clicking on an empty space in the note list panel ?

I'm already thinking about those lines. 😸

pbek:
I vote for not showing any context menu at all if no note or subfolder is selected.

```cpp
void MainWindow::on_noteTreeWidget_customContextMenuRequested(
    const QPoint pos) {
    auto *item = ui->noteTreeWidget->itemAt(pos);

    // if the user clicks at empty space, this is null and if it isn't handled
    // QON crashes
    if (item == nullptr) {
        return;
    }

    const QPoint globalPos = ui->noteTreeWidget->mapToGlobal(pos);
    const int type = item->data(0, Qt::UserRole + 1).toInt();

    if (type == FolderType) {
        openNoteSubFolderContextMenu(globalPos, ui->noteTreeWidget);
    } else if (type == NoteType) {
        openNotesContextMenu(globalPos);
    }
}
```

pbek:
I now implemented that. Thank you for brining it to my attention!

PR #1957, opened 2020-11-19
Remove rename/remove subfolder context menu option if no active subfolder

fnkabit:
It is currently possible, from the subfolder pane, to select the current folder or 'All notes' and, from the context menu, to select 'Rename subfolder' or 'Remove selected folder'. 

Actions would fail, but they should not be visible in the context menu, in the first place.

image


pbek:
Great, thank you!

pbek:
## 20.11.9
- the rename/remove subfolder context menu option are now not shown if no subfolder
  was clicked in the subfolder panel (for [#1957](https://github.com/pbek/QOwnNotes/pull/1957), thank you @fnkabit)

PR #1960, opened 2020-11-24
Rework tags for performance

Waqar144:
- Tag tree reloads are much faster now (2x faster). There should be no lag even with huge trees 3000+ note tag links
- Tag tree is not reloaded if the tag panel is not visible
- Removed an unused method in mainwindow

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -7993,7 +7993,6 @@ void MainWindow::reloadTagTree() {
         noteIdList << Note::fetchAllIdsByNoteSubFolderId(noteSubFolderId);
     }
 
-
     // create an item to view all notes
     int linkCount =
         _showNotesFromAllNoteSubFolders ? Note::countAll() : noteIdList.count();
@@ -8015,7 +8014,6 @@ void MainWindow::reloadTagTree() {
 
     // this time, the tags come first
     buildTagTreeForParentItem();
-
     // and get sorted
     if (settings.value(QStringLiteral("tagsPanelSort")).toInt() ==
         SORT_ALPHABETICAL) {
@@ -8336,7 +8334,6 @@ QTreeWidgetItem *MainWindow::addTagToTagTreeWidget(QTreeWidgetItem *parent,
             }
         }
     }
-
 
     const int linkCount = count;
     const QString toolTip = tr("show all notes tagged with '%1' (%2)")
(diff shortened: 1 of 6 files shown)


pbek:
Thank you very much!

Waqar144:
Linking https://github.com/pbek/QOwnNotes/issues/943

It's not a big thing for newer CPUs but older / weaker processors will likely benefit a lot from this.

src/mainwindow.cpp
```diff
@@ -8320,29 +8320,22 @@ QTreeWidgetItem *MainWindow::addTagToTagTreeWidget(QTreeWidgetItem *parent,
...
+                    count = Tag::countLinkedNoteFileNamesForNoteSubFolder(
+                                tagIdToCount, folder,
+                                showNotesFromAllSubFolders, isShowNotesRecursively);
                 }
             }
         } else {
-            linkedNoteIds.reserve(tagIdListToCount.size());
             for (const int tagToCount : tagIdListToCount) {
-                linkedNoteIds << Tag::fetchAllLinkedNoteIds(
+                count =  Tag::countLinkedNoteFileNames(
```
pbek, line 8330: @Waqar144, this doesn't work if `tagIdListToCount` contain multiple tags (when `Tag::isTaggingShowNotesRecursively()` is true), because it will overwrite the note count all the time. This will cause #2819.

pbek:
Causes regression #2819.

src/mainwindow.cpp
Waqar144: indeed, fortunately easy to reverse this particular change.

pbek: I already tried in [qownnotes.patch.gz](https://github.com/pbek/QOwnNotes/files/12466079/qownnotes.patch.gz), but it isn't working properly yet...

Waqar144: Full revert will be difficult, it will be easier to just revert this portion `count = ...Tag::countAll..` to `linkedNoteIds << Tag:::countAll..`. It's too old to be fully reverted and we changed apis a bit

Waqar144: So basically it will be a fix, not really a revert.

pbek: Yes, that's what I tried in above patch, but it didn't work yet...

PR #1961, opened 2020-11-25
Dont refetech the note in Note::exists, just check the id

Waqar144:


Waqar144:
This is unnecessary. I realized we can just use `isFetched()` correctly in the places where two fetches are happening.

pbek:
Hm, there were cases where a note object or note id was stored somewhere, but the internal note database was already reloaded. Theses cases wouldn't be handled correctly any more, since the note id would be > 0, but the note wouldn't actually exist any more... Hm, but loading the note wouldn't actually help neither...

pbek:
Oh, you closed it. Sorry, I'm a bit slow. Had a very painful tooth extraction... Still very painful.

Waqar144:
> Oh, you closed it. Sorry, I'm a bit slow. Had a very painful tooth extraction... Still very painful.

Haha, its fine. I created another one with `isFetched()`

PR #1962, opened 2020-11-26
Use Note::isFetched instead of exists()

Waqar144:
It doesn't replace all usages of `exists()`, only those where a second fetch is pointless


pbek:
Sounds legit!

PR #1965, opened 2020-11-26
Remove the script from db if it wasn't found

Waqar144:


src/services/scriptingservice.cpp
```diff
@@ -288,12 +290,11 @@ void ScriptingService::initComponents() {
     clearCustomStyleSheets();
     _scriptComponents.clear();
     _settingsVariables.clear();
-    QList<Script> scripts = Script::fetchAll();
+    bool enabledOnly = true;
```
pbek, line 293: Do we need `enabledOnly`?

Waqar144: Not really, but we check for it in the loop anyways. So I thought to just fetch only the enabled scripts in the first place and remove the check inside loop

pbek: No, I meant "do we need the variable"? 😁

Waqar144: Hahah. No, but the reads like a story with it 🤣

pbek: Clion shows me the names of method parameters. 😆 
And YOU, eager to get the last bit of a byte optimized, introduce a variable that is only used once and never changes. 😆

Waqar144: It will be optimized by the compiler i am sure :wink: :laughing:

pbek: Seeing that variable makes my head spin. Can't you just write a comment over the function call? 😁

Waqar144: Okay done.

src/services/scriptingservice.cpp
```diff
@@ -288,12 +290,12 @@ void ScriptingService::initComponents() {
     clearCustomStyleSheets();
     _scriptComponents.clear();
     _settingsVariables.clear();
-    QList<Script> scripts = Script::fetchAll();
 
-    Q_FOREACH (Script script, scripts) {
-        if (script.isEnabled()) {
-            initComponent(script);
-        }
+    //fetch enabled only
```
pbek, line 294: ...with a space space after `//`, like everywhere else, pleeeease 😁

Waqar144: done

pbek: Thank you 😁

PR #1989, opened 2020-12-17
Corrected a typo from "node" to "mode"

sup3rhum4n:
The suggested change is under the Tip in Portable Mode for installing QOwnNotes on Windows.
pictureOfTips


pbek:
Thank you very much! 😁

Issue #2007, opened 2021-01-04
[F] Add option to automatically switch between workspace layouts depending on window size


fnkabit:
I would like to be able to specify that below a certain window size, use layout x, above that window size, use layout y.
For example, my use case:
* when I use QON windowed I like to use it in a **Minimal** layout, where I can only edit (others may prefer to see only the **Preview Pane** when windowed, though)
* when I maximize the window, I want to switch to **Full** layout, where I can also see the **Preview Pane**

pbek:
Sounds like you want to implement:

- a hook that is called when the windows size / state changes
- a scripting function to switch between workspaces

fnkabit:
> Sounds like you want to implement:
> 
> * a hook that is called when the windows size / state changes
> * a scripting function to switch between workspaces

Yes, I think so :)
Do you think it would be useful ? Can I work on it ?
Thanks !

pbek:
Sure and sure. 😁  It would be even better if you then also write a script that allows users to switch between two workspaces in maximized / non-maximized mode.

I guess it would be great if the hook would recognize all kind of layout changes. Moving, resizing, minimizing, maximizing, full-screen, ... And also coordinates and size as parameter (or a way to fetch them).

fnkabit:
@pbek Will start working on it ASAP.

pbek:
Most excellent! 👍🏻

fnkabit:
Created PR to implement a basic version of this..

fnkabit:
Will update docs once you agree on API changes. :-)

pbek:
I added my 2c. 😉

pbek:
Ok, all mopping up is done 😆:

## 21.1.3
- there now are new scripting commands `mainWindow.getWorkspaceUuidList()`,
  `mainWindow.getWorkspaceUuid()` and `mainWindow.setCurrentWorkspace()` to
  access workspaces and a new hook `windowStateChangedHook` to notice window
  state changes (for [#2007](https://github.com/pbek/QOwnNotes/issues/2007), thank you @fnkabit)
  - for more information please take a look at the
    [MainWindow scripting documentation](https://www.qownnotes.org/scripting/classes.html#mainwindow)
    and the [Hook scripting documentation](https://www.qownnotes.org/scripting/hooks.html#windowstatechangedhook)
  - there also is a new script `Dynamic Workspace` in the script repository that
    you can use to toggle workspaces when you maximize the window
    
[...]

PR #2009, opened 2021-01-10
Facilitate scriptable workspace updates

fnkabit:
API changes required in order to implement #2007 .
Script PR: https://github.com/qownnotes/scripts/pull/120

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1203,6 +1203,8 @@ void MainWindow::updateWorkspaceLists(bool rebuild) {
         initWorkspaceComboBox();
 
         ui->menuWorkspaces->clear();
+
+        _workspaceNameUuidMap.clear();
     }
 
     const QSignalBlocker blocker(_workspaceComboBox);
@@ -1227,6 +1229,8 @@ void MainWindow::updateWorkspaceLists(bool rebuild) {
                                         QStringLiteral("/name"))
                                  .toString();
         const QString objectName = QStringLiteral("restoreWorkspace-") + uuid;
+
+        _workspaceNameUuidMap.insert(name, uuid);
 
         _workspaceComboBox->addItem(name, uuid);
 
@@ -7753,6 +7757,11 @@ void MainWindow::resetBrokenTagNotesLinkFlag() {
     if (_brokenTagNoteLinksRemoved) _brokenTagNoteLinksRemoved = false;
 }
 
+QString MainWindow::getWorkspaceUuuid(const QString &workspaceName)
+{
+    return _workspaceNameUuidMap.value(workspaceName, "");
+}
+
 /**
  * Evaluates if file is a media file
  */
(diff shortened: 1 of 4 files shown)


src/mainwindow.cpp
```diff
@@ -8596,6 +8605,34 @@ void MainWindow::linkTagNameToCurrentNote(const QString &tagName,
...
+            break;
+        default:
+            windowStateString = "nostate";
+            break;
+        }
+
+        ScriptingService::instance()->callWindowStateChangeHook(windowStateString);
+    }
+
+    QWidget::changeEvent(event);
```
pbek, line 8633: Can you please use QMainWindow::changeEvent, just to make sure.

src/mainwindow.h
```diff
@@ -161,6 +161,8 @@ class MainWindow : public QMainWindow {
     void linkTagNameToCurrentNote(const QString &tagName,
                                   bool linkToSelectedNotes = false);
 
+    void changeEvent(QEvent *event) override;
```
pbek, line 164: shouldn't that be protected? https://doc.qt.io/qt-5/qwidget.html#changeEvent

src/services/scriptingservice.h
```diff
@@ -46,6 +46,7 @@ class ScriptingService : public QObject {
                                          const QString &action,
                                          const Tag &tag = Tag(),
                                          const QString &newTagName = QString());
+    void callWindowStateChangeHook(const QString &windowStateStr);
```
pbek, line 49: We'll also need documentation for that hook on the webpage.

src/mainwindow.h
```diff
@@ -188,8 +190,14 @@ class MainWindow : public QMainWindow {
 
     void resetBrokenTagNotesLinkFlag();
 
+    Q_INVOKABLE QString getWorkspaceUuuid(const QString &workspaceName);
```
pbek, line 193: We'll need documentation for all those on https://www.qownnotes.org/scripting/classes.html#mainwindow

src/services/scriptingservice.cpp
```diff
@@ -496,6 +496,26 @@ QVariant ScriptingService::callNoteTaggingByObjectHook(
     return QVariant();
 }
 
+void ScriptingService::callWindowStateChangeHook(const QString &windowStateStr)
```
pbek, line 499: @fnkabit, @Waqar144 Do you think there ever will be a "StateChangeHook" for another windows than the MainWindow? Would that make sense?

Waqar144: I think no. I like the idea that we are single window only. A single window is giving you all the abilities. But it also depends on what the user has in mind.

pbek: Ok, thanks.

src/mainwindow.h
pbek: see: https://github.com/pbek/QOwnNotes/tree/develop/webpage/src/scripting

fnkabit: Updated docs.

pbek: Where?

src/services/scriptingservice.h
pbek: Did you push your documentation commit? I can't see any. 😁

src/mainwindow.h
pbek: Did you push your documentation commit? I can't see any. 😁

fnkabit: Something must have happened with that change. Will redo it :D

fnkabit: Forgot to create the PR :)

pbek: `getWorkspaceUuuid` has one `u` to many, hasn't it?

pbek: I'll refactor that.

PR #2010, opened 2021-01-12
windowStateChangedHook documentation

fnkabit:
Added documentation for windowStateChangedHook.


pbek:
We also need an example. I'll create one.

PR #2018, opened 2021-01-22
Replace FindAction with a more slick command bar with improved filtering

Waqar144:
This change replaces the "find action" dialog with a more (imo) slick command bar. It has the option to fuzzy filter and navigate through submenus. 

Screenshot:

image

I haven't removed the code for "find action" yet so that if this gets tested and accepted we can take care of it later. For now it is just disabled. Most of the code is vanilla Qt with a couple of things taken from outside (removal of accelerator markers is taken from KDE and fuzzy match library is taken from Forrest Woods's lib_fts).

### DIFF
--- a/src/QOwnNotes.pro
+++ b/src/QOwnNotes.pro
@@ -173,7 +173,9 @@ SOURCES += main.cpp\
     widgets/layoutwidget.cpp \
     dialogs/serverbookmarksimportdialog.cpp \
     dialogs/websockettokendialog.cpp \
-    dialogs/imagedialog.cpp
+    dialogs/imagedialog.cpp \
+    dialogs/commandbar.cpp \
+    models/commandmodel.cpp
 
 HEADERS  += mainwindow.h \
     build_number.h \
@@ -268,7 +270,9 @@ HEADERS  += mainwindow.h \
     widgets/layoutwidget.h \
     dialogs/serverbookmarksimportdialog.h \
     dialogs/websockettokendialog.h \
-    dialogs/imagedialog.h
+    dialogs/imagedialog.h \
+    dialogs/commandbar.h \
+    models/commandmodel.h
 
 FORMS    += mainwindow.ui \
     dialogs/attachmentdialog.ui \
--- a/src/dialogs/commandbar.cpp
+++ b/src/dialogs/commandbar.cpp
@@ -1,0 +1,383 @@
+#include "commandbar.h"
+
+#include <QAction>
+#include <QPainter>
+#include <QTreeView>
+#include <QLineEdit>
+#include <QKeyEvent>
+#include <QStyledItemDelegate>
+#include <QSortFilterProxyModel>
+#include <QTextDocument>
+#include <QVBoxLayout>
+#include <QCoreApplication>
+#include <QMenu>
+
+#include "models/commandmodel.h"
+#include "utils/misc.h"
+#include "libraries/fuzzy/fts_fuzzy_match.h"
+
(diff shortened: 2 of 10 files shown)


pbek:
cmake support is missing...

Waqar144:
> cmake support is missing...

Ah, yes. Will add that. Its two files just

pbek:
I'll add it

Waqar144:
> I'll add it

Awesome :D

pbek:
Dark mode needs more work, you cannot read the text. 😁 

image

Waqar144:
> Dark mode needs more work, you cannot read the text. 😁
> 
> image

Hmm, weird. Because the screenshot I shared was using dark mode. 🤔

Also can you check if it resizes / positions itself correctly on larger screens?

pbek:
And what are all those empty lines?

image

The new dialogue can't search for note folders.

pbek:
> Also can you check if it resizes / positions itself correctly on larger screens?

LGTM

Waqar144:
> And what are all those empty lines?
> 
> image
> 
> The new dialogue can't search for note folders.

Those would be the actions that don't have display text set. Can you try triggering them to see what they do and we can add display texts accordingly

pbek:
> Those would be the actions that don't have display text set. Can you try triggering them to see what they do and we can add display texts accordingly

I can't see that they are doing anything when I click them.

All sub-menus are missing too in the new list.

Waqar144:
> > Those would be the actions that don't have display text set. Can you try triggering them to see what they do and we can add display texts accordingly
> 
> I can't see that they are doing anything when I click them.
> 
> All sub-menus are missing too in the new list.

Submenus will be loaded if you press enter on a submenu, (the model will reload with submenu items)

pbek:
> Submenus will be loaded if you press enter on a submenu, (the model will reload with submenu items)

But how should a user know that the item he/she is looking for is in a certain sub-menu? The old dialog lists everything there is in a tree and you are able to search for items in that tree.

pbek:
> It has the option to fuzzy filter 

I wonder... In what way is it more fuzzy than the old dialog (substring search of multiple search terms)?

Waqar144:
> > It has the option to fuzzy filter
> 
> I wonder... In what way is it more fuzzy than the old dialog (substring search of multiple search terms)?

Type "nf" and it will match Note Folder, type "nolder" and it will still match and give you results. Basically, it does fuzzy matching inside the string and then ranks the results based on the quality of the match.

An idea, we can have a "Go To Note" Dialog, which will let you quickly jump to a note using a quick-open dialog.

Waqar144:
> > Submenus will be loaded if you press enter on a submenu, (the model will reload with submenu items)
> 
> But how should a user know that the item he/she is looking for is in a certain sub-menu? The old dialog lists everything there is in a tree and you are able to search for items in that tree.

Hmm, yes. There is not much hint there about that except you have the Menu "prefix" which kind of tells you where you are, I guess they will find out by experimenting? lol

We can load the menu actions directly too, of course.

pbek:
> Type "nf" and it will match Note Folder, type "nolder" and it will still match and give you results. Basically, it does fuzzy matching inside the string and then ranks the results based on the quality of the match.

nice

> An idea, we can have a "Go To Note" Dialog, which will let you quickly jump to a note using a quick-open dialog.

nice too :)

> We can load the menu actions directly too, of course.

we should :)

Waqar144:
I am still in favour of recursive menus ;)

PR #2019, opened 2021-01-22
Force usage of palette to ensure color is correct

Waqar144:
This should fix the colors hopefully. Not sure why they were appearing like that because I am also using dark theme.

### DIFF
--- a/src/dialogs/commandbar.cpp
+++ b/src/dialogs/commandbar.cpp
@@ -11,6 +11,7 @@
 #include <QVBoxLayout>
 #include <QCoreApplication>
 #include <QMenu>
+#include <QAbstractTextDocumentLayout>
 
 #include "models/commandmodel.h"
 #include "utils/misc.h"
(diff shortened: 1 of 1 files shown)


Waqar144:
This will address the empty actions and colors issues, hopefully. The empty items are probably the initial menu items I think.

pbek:
> because I am also using dark theme

are you sure? or are you rather using KDE with a dark theme and that is inherited by QON?

Waqar144:
> > because I am also using dark theme
> 
> are you sure? or are you rather using KDE with a dark theme and that is inherited by QON?

Hmm yes, I think that might be the reason.

Is the color okay with this change?

Waqar144:
Also we can add more things to this, such as custom actions or actions that invoke a script or context menu actions / toolbar actions and so on.

pbek:
> Is the color okay with this change?

It looks ok on my dark mode neon, need to test on my light mode neon

> Also we can add more things to this, such as custom actions or actions that invoke a script or context menu actions / toolbar actions and so on.

you mean like in the old dialog? yes, we need all the sub-items too again before we can roll out the new dialog! 😆

Waqar144:
> you mean like in the old dialog? yes, we need all the sub-items too again before we can roll out the new dialog! laughing

Only the toolbar items are missing I think. Will add that.

pbek:
> It looks ok on my dark mode neon, need to test on my light mode neon

looking good in dark mode on light mode kde neon!

PR #2020, opened 2021-01-22
Load all actions in command bar

Waqar144:
All actions will be visible now.

It has the same amount of actions as the previous widget as far as I can see. Toolbar actions are redundant so loading them will only lead to duplication. Actions for invoking scripts / context menu actions / any other can be added as an addition

### DIFF
--- a/src/dialogs/commandbar.cpp
+++ b/src/dialogs/commandbar.cpp
@@ -272,16 +272,9 @@ CommandBar::CommandBar(QWidget *parent)
     setHidden(true);
 }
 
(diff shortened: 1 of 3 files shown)


pbek:
Everytime I switch note folders the note folders are duplicated in the dialogue.

pbek:
Btw. you don't really need to ask for all submenus in a recursion...
Something like...

```cpp
    QList<QMenu *> menuList =
        _menuBar->findChildren<QMenu *>(QString(), Qt::FindDirectChildrenOnly);
```

...also does the trick. 😉

pbek:
> Everytime I switch note folders the note folders are duplicated in the dialogue.

I fixed that now.

PR #2023, opened 2021-01-24
Update config.js

Fabijenna:
I've added Hungarian to the config.js


pbek:
This breaks the config file, but I will take the block you added when hungarian really is added as translation to the website.

PR #2031, opened 2021-01-30
Update overview.md

Fabijenna:
Corrected markdown error / typo


pbek:
Thank you, but this (Spanish) file will be overwritten the next time I fetch the translations from Crowdin. The error needs to corrected on Crowdin.

PR #2036, opened 2021-02-04
Improve find actions dialog

Waqar144:
- Fixed a use after free
- Moved to the better variant of fuzzy matching from KCoreAddons::KFuzzyMatcher which has much better performance


pbek:
Build failed. Those old Qt versions... 😁

pbek:
Not yet. 😁

Waqar144:
Should build now ;)

PR #2056, opened 2021-02-17
Restore workspace after window resize

fnkabit:
This is my QON workflow:
* I use it in a small window, alongside other applications, for quickly looking through my notes; I do this in a 'Preview only' workspace because I don't do any editing at this point
* When I want to edit something, I maximize the window, and this will switch to the 'Edit' workspace (using the 'Dynamic workspace' plugin).

The problem is that the window state is applied before the resize, so after the resize the docks will not have the sizes that I previously set:
https://www.youtube.com/watch?v=0y25bzaL2nw

In order to preserve the dock sizes, the window state must be applied after the resize:
https://www.youtube.com/watch?v=_bVFMDKUvkE

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -11533,7 +11533,7 @@ void MainWindow::setCurrentWorkspace(const QString &uuid) {
     settings.setValue(QStringLiteral("currentWorkspace"), uuid);
 
     // restore the new workspace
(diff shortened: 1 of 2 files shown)


pbek:
Thank you very much!

pbek:
I just found out this breaks https://github.com/pbek/QOwnNotes/issues/2180 😅

PR #2096, opened 2021-04-11
qt6: use QStringRef / QStringView as required

Waqar144:
This ports all of our own stuff to be more qt6 compatible. 

Usage inside all other libraries other than sonnet still needs fixing.


pbek:
Awesome! Thanks a lot! 👍🏻

PR #2097, opened 2021-04-11
Disable xmlpatterns and co. for Qt-6

Waqar144:
Make QON compile on Qt-6

Changes done:

- We are not getting xmlpatterns. So, we need to use something else. For now xml-pattern code is disabled for Qt-6

### DIFF
--- a/src/dialogs/attachmentdialog.cpp
+++ b/src/dialogs/attachmentdialog.cpp
@@ -88,7 +88,7 @@ void AttachmentDialog::on_downloadButton_clicked() {
     QUrl url(ui->fileEdit->text());
     QNetworkRequest networkRequest(url);
 
-#if (QT_VERSION >= QT_VERSION_CHECK(5, 6, 0))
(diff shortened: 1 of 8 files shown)


pbek:
Thank you! How do we remember to implement the functionality? 😅

Waqar144:
I didn't intend for this to be merged just yet ;p

> Thank you! How do we remember to implement the functionality? 😅

First we just make it compile, we need to access how much work needs to be done for Qt6 support.

Then, we write our own abstraction for XmlQuery and anything else that is unavailable. 

Till we are able to get everything working for Qt6, we keep it internal. We have a lot of time to do this. 

So, I will open another PR and we'll try to make Qt6 compile there.

Waqar144:
The biggest pain point right now is `int` vs `qsizetype`. 

That needs to be carefully handled otherwise it can bite us at some unknown time :/

pbek:
> I didn't intend for this to be merged just yet ;p

Oh, I thought that's what "pull requests" are for, asking for something to be merged. 😅 😁 

> The biggest pain point right now is int vs qsizetype.

yeah 😬

PR #2116, opened 2021-04-26
Create QON-PR-by-Amy-2

Fabijenna:
Added CSS


pbek:
Adding css to a html file doesn't make it markdown. 😅

PR #2127, opened 2021-05-05
QRegExp -> QRegularExpression for Qt6

Waqar144:
Hopefully this will not break anything..

Fakevim needs to be updated, that is still using QRegExp.

Signed-off-by: Waqar Ahmed <waqar.17a@gmail.com>


pbek:
Isn't the syntax different at times from QRegExp to QRegularExpression?

Waqar144:
> Isn't the syntax different at times from QRegExp to QRegularExpression?

Yes, but Qregularexpression is fully pcre compliant while Qregexp isn't I think, and they have api differences. I am not sure about the regex syntax differences between them.

PR #2140, opened 2021-05-16
Stop installing app icon in pixmaps location

pinotree:
The `/usr/share/pixmaps` location is considered a legacy location for application icons; since the application icons are already installed in the global XDG hicolor theme, then simply stop installing the 128px one in the legacy pixmaps location.


pbek:
Thank you!

PR #2155, opened 2021-06-06
Fix highlighting with python triple quote strings

Waqar144:
Fixes the highlighting for the preview

#2147


pbek:
Ah, nice. Do you have an example what this fix fixes? I never could reproduce #2147 (or even understand what the issue was supposed to be)

Waqar144:
> Ah, nice. Do you have an example what this fix fixes? I never could reproduce #2147 (or even understand what the issue was supposed to be)

You couldn't reproduce 2147 because your "code" color is similar to "string" color I guess ;)

PR #2168, opened 2021-06-27
Fix fakevim build for ancient Qt version

Waqar144:
The code is more crappy now but it should now build at least.

This is getting unreasonable and a time waste and I think will make things more and more hard to maintain in the future.

### DIFF
--- a/src/libraries/fakevim/fakevim/fakevimhandler.cpp
+++ b/src/libraries/fakevim/fakevim/fakevimhandler.cpp
@@ -104,6 +104,16 @@
 #   define UNDO_DEBUG(s)
 #endif
 
+// Added by waqar144
+// Purpose: We support Qt <= 5.8
+#ifndef Q_FALLTHROUGH
+#  if (defined(Q_CC_GNU) && Q_CC_GNU >= 700) && !defined(Q_CC_INTEL)
+#    define Q_FALLTHROUGH() __attribute__((fallthrough))
+#  else
+#    define Q_FALLTHROUGH() (void)0
+#  endif
+#endif
+
 namespace FakeVim {
 namespace Internal {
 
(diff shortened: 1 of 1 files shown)


pbek:
Thanks a lot for the effort! Maybe we should raise the bar to Qt 5.5 (Ubuntu Xenial has Qt 5.5.1). That would make live easier. Or just degradate functionality gracefully for older Qt versions (instead of re-implementing stuff that didn't exist back then).

Waqar144:
I think we should at least target Qt 5.9 as a minimum. We can't keep supporting old / dead operating systems. People need to move on.

pbek:
But wasn't it always a cure feature of QOwnNotes that it also runs on older operating systems...

pbek:
> no one else cares about such old versions.

the old versions of Qt were dropped because of their proprietary license enforcement strategy...

pbek:
I can see run-times down to Qt 5.5.1 this month in Matomo. This year even down to Qt 5.3.2.

pbek:
I more tend into the direction of "degradate functionality gracefully for older Qt versions"

Waqar144:
Ubuntu Xenial is EOL already. 🤷🏻‍♂️ And Qt 5.5 was released 6 years ago and died 3 years ago, not sure I want to support that anymore.

QON can still support old operating systems, *to an extent*, and the rest of the users who want to stay on old distributions can get an AppImage.

Also, I am unsure how degrading functionality will work. Take this PR, you can't degrade functionality here because that will result in a broken software, you can just altogether remove fakevim support for Qt X.X, but I fail to see the point of it.

Because we have to support such old versions, we can't just directly update our libraries because they use stuff that isn't available, and we have to implement or work around or revert to older versions.

pbek:
We could always just disable a whole feature for things < 1% of users use for older Qt versions. 😁

Waqar144:
> We could always just disable a whole feature for things < 1% of users use for older Qt versions. 😁

No we can't if they are core features or without crippling the app. And tbh, I don't understand why we would want to do something like that. And just thinking about that, I think it will be *much* more work.

PR #2173, opened 2021-06-28
Update get-involved.md

Fabijenna:
The YouTube channel  QOwnNotes in a nutshell (https://www.youtube.com/channel/UC6Xpk_B1MFfvhBCsH_MrOEw) will stream positive feedback from users in a series of "Feedback" (apart from tips, features and knowhows)


pbek:
> Publish your **experience** with QOwnNotes **on Youtube**! [Send audio/video/text to](mailto:amydoralang@aol.de)

1) Do we want to encourage users to publish YouTube videos (like you suggest in the text) or do we rather want them to send you an email (like suggested in the link)? We need to decide on one or mention that both options are ok.
2) The link text `Send audio/video/text to` (without a name or something) doesn't make a lot of sense. Maybe something like `Send audio/video/text to Amy` would be better.

PR #2178, opened 2021-07-03
Update get-involved.md

Fabijenna:
rewrote line - 18 to a shorter sentence
added - 19 for everyone to participate in a survey I wrote (might be expanded in future)


Fabijenna:
> added - 19 for everyone to participate in a survey I wrote (might be expanded in future)

You want me to work more on the Survey, and is that's why you didn't include point 19?

Fabijenna:
If yes, please list the things you'd rather have me updated. Thanks.

pbek:
> Fill out this very short new [**survey**](https://freeonlinesurveys.com/s/nA8t17k8)!

The text was added the way you put it in and I can see it on the bottom of https://www.qownnotes.org/contributing/get-involved.html.

pbek:
But now that I look at it I'll remove the word "new", because the survey will not be new in one month or one year when nobody is removing the word by then. 😁

Fabijenna:
Sure. I wouldn’t have put it this way an other time.

PR #2184, opened 2021-07-11
Fix typo

sk22:
Fix minor typo: "loosing" should be "losing" in "without loosing your current settings".


pbek:
Thank you very much!

sk22:
No problem!

PR #2250, opened 2021-10-05
Use static QregularExpressions to improve performance

Waqar144:
Constructing `QRegularExpression` is very expensive.

The change doesn't fix all the usages(since there are too many). I have fixed the ones that are more commonly used.

### DIFF
--- a/src/entities/note.cpp
+++ b/src/entities/note.cpp
@@ -395,7 +395,7 @@ QStringList Note::getMediaFileList() {
 
     // match image links like ![media-qV920](file://media/608766373.gif)
     // or  ![media-qV920](media/608766373.gif)
-    QRegularExpression re(QStringLiteral(R"(!\[.*?\]\(.*media/(.+?)\))"));
+    static const QRegularExpression re(QStringLiteral(R"(!\[.*?\]\(.*media/(.+?)\))"));
     QRegularExpressionMatchIterator i = re.globalMatch(_noteText);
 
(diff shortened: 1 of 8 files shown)


data-man:
Just migrate to [CTRE](https://github.com/hanickadot/compile-time-regular-expressions).

Waqar144:
> Just migrate to [CTRE](https://github.com/hanickadot/compile-time-regular-expressions).

That's.. pointless? and probably not even possible for us to completely migrate to it even if we wanted to, which again, is completely pointless for us.

pbek:
LGTM, thank you very much. 😁

data-man:
Yes, CTRE is only for powerful skills :rofl:

Waqar144:
Not sure if I understand you, If we only support one extension in this
case, then we don't even need to check for the default extension, right?

pbek:
## 21.10.9
- there now will be a warning that a restart of the application is needed when
  changing the default note file extension (for [#2250](https://github.com/pbek/QOwnNotes/pull/2250))

Waqar144:
https://github.com/pbek/QOwnNotes/pull/2267 created for the above.

pbek:
I liked your static variable. I didn't want to bother implementing that all with new variable for a use-case that happens once in a thousand years (someone changing the default note file extension) 🤣 

> #2267 created for the above.

...just needs some fixes 😉

Issue #2165, opened 2021-06-24
code block syntax highlighting in preview panel only working after editing note.


martinbra:
Thanks for solving the last issue so quickly.
The following issue was tested with 21.6.3 and 21.6.4 and I had the sames results.

> The syntax highlighting in the preview panel only works after I edit the note. If I am just browsing around different notes the markdown is parsed correctly, but the highlighting only works after I edit the file (add a space anywhere, for instance). I tried on the previous version too, 21.6.3, and the bug was already there.

_Originally posted by @martinbra in https://github.com/pbek/QOwnNotes/issues/2161#issuecomment-865178387_

#### Expected behaviour

When clicking to view a note, the preview panel should display the completed evaluated note markdown:
![image](https://user-images.githubusercontent.com/29928669/123273522-19285f00-d4d9-11eb-9a24-491057ca71db.png)

#### Actual behaviour

After **reopening** qOwnNotes, When clicking to view a note with code block in tripple ticks, the preview panel shows the code block, but doesn't highlight the syntax as specified in the code block (as if there is no specification).
![image](https://user-images.githubusercontent.com/29928669/123273851-5f7dbe00-d4d9-11eb-91df-0381f74ffce2.png)

#### Steps to reproduce

**(1)** Create Note with a code block:

    MyNote 
    ==============
    
    ```python
    import this
    song = "Mary had a little lamb."
    year = 1000 + 2*450 + 30 + 1
    ```

**(2)** **Restart qOwnNotes**

**(3)** View MyNote -> syntax highlighting is only working inside the edit area, not on preview panel.
![image](https://user-images.githubusercontent.com/29928669/123275157-796bd080-d4da-11eb-8eb7-585abdb94e31.png)

**(4)** View another note and come back -> still not working.
**(5)** Add any character anywhere inside the note. Preview is updated and highlights it correctly.
![image](https://user-images.githubusercontent.com/29928669/123275357-a28c6100-d4da-11eb-9e95-98eb1612bc4f.png)

**(6)** View another note and come back -> Still working.

pbek:
I tried that on KDE Neon and had full syntax highlighting directly after a restart.

Can you please post the output from the debug settings that you can copy when you get into the settings dialog in QOwnNotes and head over to the *Debug* section of it.

Do you have the same issue when running QOwnNotes in a new session (see [cli-parameters](https://www.qownnotes.org/getting-started/cli-parameters.html)).

pbek:
[...]

PR #2251, opened 2021-10-06
Add debug log for #2165

Waqar144:
Added a couple of debug statements to see if I can catch the issue.

### DIFF
--- a/src/helpers/codetohtmlconverter.cpp
+++ b/src/helpers/codetohtmlconverter.cpp
@@ -2,6 +2,7 @@
 
 #include <QStringBuilder>
 #include <QtCore/QRegularExpression>
+#include <QDebug>
 
 #include "libraries/qmarkdowntextedit/qownlanguagedata.h"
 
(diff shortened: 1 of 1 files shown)


src/helpers/codetohtmlconverter.cpp
```diff
@@ -51,6 +53,7 @@ void CodeToHtmlConverter::initCodeLangs() Q_DECL_NOTHROW {
 }
 
 QString CodeToHtmlConverter::process(const QString &input) const {
+    qDebug() << "Going to highlight input: " << input.mid(0, 12) << ", with lang: " << _currentLang;
```
pbek, line 56: You don't need leading and trailing spaces, qDebug seems to add them on its own.

Waqar144: done

its only temporary though. Will remove these once the issue is identified.

pbek: Ah, OK. Because the whitespaces are also in above debug log. 😁

Waqar144: yeah, but does it really matter? (as said its temporary, just need it to confirm a few assumptions)

Waqar144:
Lets hope this brings out something useful

PR #2257, opened 2021-10-11
Make the build faster

Waqar144:
- Use pch to improve the build time => +15 seconds improvement
- Some header cleanup

Also, use `-j 8` on github for linux builds. I can already see that the build finished in 2m instead of the previous 5+, which is an improvement.


pbek:
> Also, use -j 8 on github for linux builds.

Very cool. I didn't know how many cores can be used in the github ci!

Waqar144:
> > Also, use -j 8 on github for linux builds.
> 
> Very cool. I didn't know how many cores can be used in the github ci!

Well let's start with 8, and make github work harder 🤣

We can experiment with more later, perhaps there will be an improvement, perhaps not. Also, not sure how to do it for windows but I am sure there will be a way.

pbek:
Huge difference in Linux and macOS, not much in Windows.

PR #2284, opened 2021-10-31
Adds security token instructions & some pg re-org

alison985:
The Security Token component of installation was confusing to me, so I am submitting this update to the documentation.

The re-organization at the top of the page is an attempt to both make the steps easy and short to follow and to allow for continued expansion of the page's content.


pbek:
Thank you very much. It's much more clear now.

PR #2303, opened 2021-11-20
Try to improve startup times a bit more

Waqar144:
Delay loading things initially and allow the mainWindow to come up quickly. We can probably do a bit more, such as create notes list after creating the mainWindow and so on.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -292,6 +292,7 @@ MainWindow::MainWindow(QWidget *parent)
     initShowHidden();
 
     createSystemTrayIcon();
+
     buildNotesIndexAndLoadNoteDirectoryList(false, false, false);
 
     // setup the update available button
@@ -402,7 +403,9 @@ MainWindow::MainWindow(QWidget *parent)
     this->loadNoteFolderListMenu();
 
     // update panels sort and order
-    updatePanelsSortOrder();
+    QTimer::singleShot(10, this, [this]{
+        updatePanelsSortOrder();
+    });
 
     this->updateService = new UpdateService(this);
     this->updateService->checkForUpdates(this, UpdateService::AppStart);
@@ -567,8 +570,6 @@ MainWindow::MainWindow(QWidget *parent)
             &MainWindow::on_action_Quit_triggered);
 
     automaticScriptUpdateCheck();
-
-    _commandBar = new CommandBar(this);
 
     // trigger cli parameter menu action if there was any set
     triggerStartupMenuAction();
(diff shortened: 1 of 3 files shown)


src/mainwindow.cpp
```diff
@@ -1665,8 +1665,10 @@ void MainWindow::initStyling() {
                          "{background-color: %1;}")
                          .arg(noteTagFrameColorName);
 
-    qApp->setStyleSheet(appStyleSheet);
-    Utils::Gui::updateInterfaceFontSize();
+//     qApp->setStyleSheet(appStyleSheet);
```
pbek, line 1668: Why can or should this be disabled?

Waqar144: This happens twice, "setStyleSheet", once here once inside the updateInterfaceFontSize() function, So maybe we can just do it once? But maybe we will now need to pass in the stylesheet to the function..

pbek: updateInterfaceFontSize uses the stylesheets that were set by the line you disabled and updateInterfaceFontSize will be executed later too... if you want to remove that line you need to make sure the styles will be set in updateInterfaceFontSize and none of the styles from appStyleSheet are supposed to be updated later (because then those styles would get overwritten by updateInterfaceFontSize again and again).

"setStyleSheet" happens in some other places too.

pbek: If you want to play safe, just leave the line. 😁

Waqar144: But why does it need to be reset again? Why not just once?

pbek: > But why does it need to be reset again?

Who "it"?

Waqar144: The stylesheet of course, and the subsequent calls to updateInterfaceFontSize. Why not just once? The function sounds to me like something we would only do once per start.

pbek: on_overrideInterfaceFontSizeGroupBox_toggled and on_interfaceFontSizeSpinBox_valueChanged in the settings dialog

pbek: I've added the command back in until there is a better solution. 😉

pbek:
And the old tune again, @Waqar144 😅  The builds systems broke for older Qt versions...
https://build.opensuse.org/package/live_build_log/home:pbek:QOwnNotes/desktop/Debian_8.0/x86_64

pbek:
> And the old tune again, @Waqar144 sweat_smile The builds systems broke for older Qt versions... https://build.opensuse.org/package/live_build_log/home:pbek:QOwnNotes/desktop/Debian_8.0/x86_64

Should be fixed now.

PR #2307, opened 2021-11-22
Updated two paragraphs for readability

Fabijenna:

PR #2308, opened 2021-11-22
Update translation.md

Fabijenna:
Updated according to agreed content.


pbek:
What's wrong with this PR? Don't you want to add this one over #2307?

Fabijenna:
I couldn't see what you did and I thought I needed to update "having" myself so I wrote a whole other PR with also the previous updates.

I couldn't see what else I could do.

Fabijenna:
The only new is the sentence with the "having" error, I updated.

pbek:
Then I'll merge #2307, so you can edit it...

Issue #2309, opened 2021-11-23
v21.11.9 - Segfault on first launch when double-clicking highlighted file in note list


sahensley:
#### Expected behaviour

The content is opened in a new tab.

#### Actual behaviour

The application crashes.  When launching from the CLI - the error which goes to the console is a segfault.

#### Steps to reproduce

This issue happens in very specific circumstances.  I swear I'm not trying to model my workflow after this [appropriate XKCD](https://xkcd.com/1172/) 😄 

For reference, this screenshot has been taken immediately after launching 21.11.8. The last note I used - "_scratchpad" - loads into the main editor panel.
![QOwnNotes - 21 11 8](https://user-images.githubusercontent.com/6700334/142961826-c7ed2535-cc17-4b7c-8ab1-f0af040aa668.png)

This is a screenshot immediately after launching 21.11.9.  Notice how the main editor panel is blank.  I can trigger a crash (no error dialog) by double-clicking on the note which is highlighted by default in the Note List panel, which is the _scratchpad note in this instance.  I can completely avoid a crash by clicking on any other note, which then puts text into the editor panel.  Once the panel has content in it, I can go back to the initial note of the session without issue.

![QOwnNotes - version 21 11 9 - build 916](https://user-images.githubusercontent.com/6700334/142961833-4e0e0071-0485-4e06-b9d0-36a314c15d8d.png)

If QON has the option to "Restore open notes.." and tabs were open when the application was closed, QON will not crash when double-clicking the note highlighted upon launch.

So to summarize the _very_ specific circumstances:
- QON v21.11.9
- "Restore open note tabs" is unchecked OR previous session had _all_ tabs closed
- The default note which is highlighted upon launch is double-clicked to open as the first tab upon launching QON

I verified this happens on both the Linux AppImage and Windows install of 21.11.9.

Let me know if you run into any issues replicating the crash, I can record a video of some sort.  Thanks again! 😄

pbek:
I can't reproduce a crash with 21.11.9 so far on Linux or macOS. So the application doesn't crash for you in 21.11.8? It could have something to do with #2303 (@Waqar144?).

Can you please post the output from the debug settings that you can copy when you get into the settings dialog in QOwnNotes and head over to the *Debug* section of it.

Do you have the same issue when running QOwnNotes in a new session (see [cli-parameters](https://www.qownnotes.org/getting-started/cli-parameters.html))?

Waqar144:
[...]

PR #2310, opened 2021-11-24
Don't check non note files for tags

Waqar144:
For me it ended up checking images for tags resulting in storing lots of
non-sense in the database. This then slows everything down as we can potentially have tons of tags in the database

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -8948,9 +8948,15 @@ void MainWindow::handleScriptingNotesTagUpdating() {
 
     // workaround when signal blocking doesn't work correctly
     directoryWatcherWorkaround(true, true);
-
-    const QVector<Note> &notes = Note::fetchAll();
+    const QString defExt = Note::defaultNoteFileExtension();
+    const QStringList exts = Note::customNoteFileExtensionList(QStringLiteral("."));
+    const QVector<Note> notes = Note::fetchAll();
     for (const Note &note : notes) {
+        // ignore non-note files
+        const auto suffix = note.fileNameSuffix();
+        if (defExt != suffix || !exts.contains(suffix)) {
+            continue;
+        }
         QSet<int> tagIdList;
         const QStringList tagNameList =
             ScriptingService::instance()
--- a/src/entities/note.cpp
+++ b/src/entities/note.cpp
@@ -2221,9 +2221,7 @@ bool Note::refetch() {
  * Returns the suffix of the note file name
  */
 QString Note::fileNameSuffix() const {
-    QFileInfo fileInfo;
-    fileInfo.setFile(_fileName);
-    return fileInfo.suffix();
+    return QFileInfo(_fileName).suffix();
 }
 
 /**


Waqar144:
For the note-tag-script. I think if we just check whether `noteText.indexOf(tagMarker)` early on, we can get a lot of speedup as many notes may not have it and we can just return early.

pbek:
I was just about to push my fix for #2309. Is this pr about something different or also #2309?

Waqar144:
> I was just about to push my fix for #2309. Is this pr about something different or also #2309?

It's just an issue I faced when test 2309. As soon as I enabled the script, QON started taking 30+ seconds to startup.

But its different, not related to the crash.

pbek:
> But its different, not related to the crash.

The crash should already be fixed with 21.11.10!

src/mainwindow.cpp
```diff
@@ -8948,9 +8948,13 @@ void MainWindow::handleScriptingNotesTagUpdating() {
 
     // workaround when signal blocking doesn't work correctly
     directoryWatcherWorkaround(true, true);
-
-    const QVector<Note> &notes = Note::fetchAll();
+    const QString defaultNoteFileExtension = Note::defaultNoteFileExtension();
+    const QVector<Note> notes = Note::fetchAll();
     for (const Note &note : notes) {
+        // ignore non-note files
+        if (note.fileNameSuffix() != defaultNoteFileExtension) {
```
pbek, line 8955: You can have an unlimited amount of note file extensions, not just the `defaultNoteFileExtension`!

pbek: `Note::customNoteFileExtensionList` has that list.

Waqar144: done.

Waqar144: I will always wonder what is the use of such a feature

src/mainwindow.cpp
```diff
@@ -8948,9 +8948,15 @@ void MainWindow::handleScriptingNotesTagUpdating() {
...
     directoryWatcherWorkaround(true, true);
-
-    const QVector<Note> &notes = Note::fetchAll();
+    const QString defExt = Note::defaultNoteFileExtension();
+    const QStringList exts = Note::customNoteFileExtensionList(QStringLiteral("."));
+    const QVector<Note> notes = Note::fetchAll();
     for (const Note &note : notes) {
+        // ignore non-note files
+        const auto suffix = note.fileNameSuffix();
+        if (defExt != suffix || !exts.contains(note.fileNameSuffix())) {
```
pbek, line 8957: Is there a reason you are using `note.fileNameSuffix();` twice?

pbek: And doesn't `Note::customNoteFileExtensionList` already contain `Note::defaultNoteFileExtension()`?

Waqar144: changed.

No idea about whether its included or not.

pbek: > No idea about whether its included or not.

Don't you want to test that before we include broken code? 😆

Waqar144: checked, both are different for me. `custom...()` is empty, the other one isn't.

src/mainwindow.cpp
```diff
@@ -8948,9 +8948,15 @@ void MainWindow::handleScriptingNotesTagUpdating() {
...
     directoryWatcherWorkaround(true, true);
-
-    const QVector<Note> &notes = Note::fetchAll();
+    const QString defExt = Note::defaultNoteFileExtension();
+    const QStringList exts = Note::customNoteFileExtensionList(QStringLiteral("."));
+    const QVector<Note> notes = Note::fetchAll();
     for (const Note &note : notes) {
+        // ignore non-note files
+        const auto suffix = note.fileNameSuffix();
+        if (defExt != suffix || !exts.contains(suffix)) {
```
pbek, line 8957: > No idea about whether its included or not.

The answer is "yes", it's included. I just checked.

But what are you trying to achieve with this check? Why should there be non-note files in `Note::fetchAll()`?

src/mainwindow.cpp
pbek: Now that's weird... It should be included (and was for me)...

src/mainwindow.cpp
Waqar144: See my previous comment. For me the custom one is empty.

src/mainwindow.cpp
pbek: But still: https://github.com/pbek/QOwnNotes/pull/2310#discussion_r756094020

src/mainwindow.cpp
Waqar144: > But what are you trying to achieve with this check? Why should there be non-note files in Note::fetchAll()?

Good question. Don't know the answer though, but I guess the code can tell.

src/mainwindow.cpp
Waqar144: putting both in a `qWarning()<<..` gives:

```
Warning: Extensions "md" () ((null):0, (null)
```
notice the empty ()

src/mainwindow.cpp
pbek: > I guess the code can tell

what code? did you find none-notes in the note list? what files?

Waqar144: JPEG files

Waqar144: referenced by a note.

src/mainwindow.cpp
pbek: How did you manage that? 😆 

https://github.com/pbek/QOwnNotes/blob/bacd3b8e80d608c9dfab0c8dfb7160343d196bef/src/services/databaseservice.cpp#L872-L883

...should add at least md and txt.

pbek: And you shouldn't be able to remove all extensions in the settings.

src/mainwindow.cpp
pbek: Whaaaat? 😆

pbek: maybe because your Note::customNoteFileExtensionList is empty, can you please try to add some extension in the general settings.

pbek: in addition I'll add a fix for you in customNoteFileExtensionList:

```cpp
    if (list.isEmpty()) {
        list << defaultNoteFileExtension();
    }
```

pbek: > referenced by a note.

but why was it in the note list?

pbek: Can you please check if the latest development branch still has JPEG files in the note list?

Waqar144: just a note. not note list.

can check later.

pbek: Did you? 😁

Waqar144: can't reproduce anymore. I guess there wasn't a bug or some other misconfig.

pbek: Great

PR #2315, opened 2021-11-30
Update metrics.md

Fabijenna:
Changed a sentence because of easier readability and word repetition


Fabijenna:
But can you delete the "s" in "languages", please, that's wrong! It must be "language"!

PR #2317, opened 2021-11-30
Update metrics.md

Fabijenna:


pbek:
Can you please consolidate all your changes to the same file in *one* PR...

Fabijenna:
For me, they are on different pages. I will check how to find the whole file

pbek:
All your PRs were changing the same file. And even if they weren't for the same file, you can also change several files in the same commit or in the same PR (for the latter for example see https://webapps.stackexchange.com/questions/95940/how-can-i-change-multiple-files-and-submit-them-in-one-pull-request-on-github).

PR #2345, opened 2021-12-31
Use QLiteHtml for preview (experimental)

Waqar144:
This change introduces QLiteHtml for previewing. QLiteHtml brings with it full Html/Css support so it is able to render the preview a lot better than what we currently use.

Right now it is only availble if you build QOwnNotes yourself with following enabled:

### using cmake
```cmake
cmake -DUSE_QLITEHTML=ON ..
```
### using qmake
```
qmake CONFIG+=USE_QLITEHTML ../
```

I have tested things and I havent' found anything broken really. The selection is not too good, but useable. There is no regexp search, but that can be implemented I think. Rest everything seems to work as expected.

Also, I think it will be hard to port to older Qt versions and by default support Qt >= 5.11. Thus keeping both solutions is the only way if we want to keep old Qts.


src/CMakeLists.txt
```diff
@@ -40,6 +40,12 @@ endif()
 add_subdirectory(libraries/fakevim)
 add_subdirectory(libraries/diff_match_patch)
 
+option(USE_QLITE_HTML "Build using system installed Botan 2 crypto library" OFF)
```
pbek, line 43: What has USE_QLITE_HTML to do with Botan?

Waqar144: Copy paste ;)

pbek: I'll change that to "QLiteHtml" then... 😉

pbek:
So far I just had some styling issues with this markdown:

```markdown
*Italic*
_italic_

**bold**

__underline__
```

pbek:
Underline seems to have more priority what we give it in our note edit:

image

pbek:
Other things I found that didn't work:

- checkbox lists
- context menu
- yes, selections still feel strange 😁 

image

pbek:
But the Qt 5.9 build is broken! https://github.com/pbek/QOwnNotes/runs/4676252035?check_suite_focus=true

pbek:
Your singleshot: https://github.com/pbek/QOwnNotes/runs/4676252035?check_suite_focus=true#step:14:1575

Waqar144:
Context menu support will need to be added manually, but that should be fairly easy to add

Waqar144:
Proper checkbox support also need to be added, but the one we were using previously works if I recall correctly.

Waqar144:
> Underline seems to have more priority what we give it in our note edit:
> 
> image

This seems to work as expected. Am I missing something? The italic in the editor maybe a style just, and in fact shouldn't be italic + underline. The preview is correct in rendering it just as italic.

pbek:
> This seems to work as expected. Am I missing something? 

Ah, the old preview styles it the same way. My bad...

pbek:
Ok, I'm building a new release (without flags! 😁). Let's see if anything breaks for older versions of Qt. 😅

Waqar144:
I am interested in whether the fancy markdown styling css stylesheets will work correctly with this. Qlitehtml boasts that it can render bootstrap correctly, so seems like it should work (assuming the actual painting is being done correctly).

pbek:
...all the possibilities! 😁

pbek:
No building issues so far with QLiteHtml turned off...

Waqar144:
> No building issues so far with QLiteHtml turned off...

Great!

PR #2349, opened 2022-01-01
Qlitehtml changes

Waqar144:
Add context menu to new html preview widget
Add checkbox rendering
Add zooming

### DIFF
--- a/src/entities/note.cpp
+++ b/src/entities/note.cpp
@@ -2453,7 +2453,11 @@ QString Note::textToMarkdownHtml(QString str, const QString &notesPath,
     unsigned flags = MD_DIALECT_GITHUB | MD_FLAG_WIKILINKS |
                      MD_FLAG_LATEXMATHSPANS | MD_FLAG_UNDERLINE;
     // we parse the task lists ourselves
+
+    // we render checkboxes when using qlitehtml
+#ifndef USE_QLITEHTML
     flags &= ~MD_FLAG_TASKLISTS;
+#endif
 
     const QSettings settings;
     if (!settings
--- a/src/libraries/qlitehtml/src/CMakeLists.txt
+++ b/src/libraries/qlitehtml/src/CMakeLists.txt
@@ -42,6 +42,8 @@ endif()
 
 if(TARGET litehtml)
   set(PUBLIC_HEADERS
+      element_checkbox.h
+      element_checkbox.cpp
       container_qpainter.h
       container_qpainter_p.h
       qlitehtml_global.h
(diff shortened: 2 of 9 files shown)


pbek:
By default (reset zoom) the preview is far too small for me now:

image

pbek:
Are those checkboxes the way QLiteHtml renders them or can we use the old symbols for the checkmarks?

image

It would be great if there were clickable again too. 😁

Waqar144:
> Are those checkboxes the way QLiteHtml renders them or can we use the old symbols for the checkmarks?

It's how I did it for now. The colors are hard-coded right now but should be changed to use the palette instead. 

The old symbols might work, haven't tested it. 


> image
> 
> It would be great if there were clickable again too. 😁

Sure, but I chose not to add that because the click is just visual and otherwise useless(in qlitehtml). Making it useful is a *lot* of work that I am not interested in doing atm. And the preview is static, it changes based on the content of the editor and not the other way around.

Waqar144:
> By default (reset zoom) the preview is far too small for me now:
> 
> image

Yes, that needs fixing I guess.

pbek:
> Sure, but I chose not to add that because the click is just visual and otherwise useless(in qlitehtml). Making it useful is a lot of work that I am not interested in doing atm. And the preview is static, it changes based on the content of the editor and not the other way around.

Ah, I see. I very much enjoyed clicking on the preview. 😁

Waqar144:
> 
> Ah, I see. I very much enjoyed clicking on the preview. 😁

"checking" the checkbox on click is easy, but that "checked" checkbox isn't very useful if you can't signal some other class that this checkbox got checked. This is very hard to implement for us , as there can be many checkboxes and you not only need to signal the listeners but you also need a way to tell them "which one" of the checkboxes got clicked.

In a WebKit/browser world I guess one would use Javascript to communicate this info to C++ maybe, however it's much harder with litehtml currently. As soon as you make it dynamic, things get 100x difficult.

pbek:
But links are clickable... aren't our current checkboxes also links (I can't remember any more)...

Waqar144:
> But links are clickable... aren't our current checkboxes also links (I can't remember any more)...

Not sure about the current one, but this is `<intput type=checkbox>`, not sure if a link can be added to this checkbox.

pbek:
Something like this:

```cpp
    const QString checkboxStart = QStringLiteral(
        R"(<a class="task-list-item-checkbox" href="checkbox://_)");
    text.replace(re1,
        listTag % QStringLiteral("\\1") % checkboxStart % QStringLiteral("\">&#9744;</a>"));
    text.replace(re2,
        listTag % QStringLiteral("\\1") % checkboxStart % QStringLiteral("\">&#9745;</a>"));
```

Waqar144:
> 

This is what we already have, right? If so, I think there is a link clicked signal in qlitehtmlwidget which can used to handle the link and do whatever we want.

PR #2350, opened 2022-01-01
Use headerView auto resize to contents

Waqar144:
no need to do it manually


pbek:
Looks like it would be working... 😅

PR #2351, opened 2022-01-02
Move notesubfoldertreewidget to its own class

Waqar144:


pbek:
Seems to break some build processes...

Waqar144:
Can you merge this please, if its acceptable? It blocks other things that I want to do.

src/mainwindow.ui
```diff
@@ -770,7 +770,7 @@
              <rect>
               <x>0</x>
               <y>0</y>
-              <width>293</width>
+              <width>300</width>
```
pbek, line 773: is this needed?

src/mainwindow.ui
```diff
@@ -859,7 +859,7 @@ p, li { white-space: pre-wrap; }
      <x>0</x>
      <y>0</y>
      <width>1383</width>
-     <height>34</height>
+     <height>27</height>
```
pbek, line 862: is this needed?

src/mainwindow.ui
Waqar144: ask qtcreator ¯\_(ツ)_/¯

src/mainwindow.ui
Waqar144: ¯\_(ツ)_/¯

src/mainwindow.ui
pbek: I'll like to keep such changes out... often enough they are messing with something on low resolutions...

pbek:
Other than above... LGTM (without knowing what regressions it will cause 🙈 )

src/mainwindow.ui
Waqar144: They are not. And I doubt you even specified them by hand yourself.

pbek: Every time I don't discard the changes QtCreator makes to the resolution of the settings dialog it is too big for small screens by default...

Waqar144: done.

Waqar144:
> Other than above... LGTM (without knowing what....

you don't have to merge, its just a proposal.

pbek:
> its just a proposal.

is it? 🤣 

> Can you merge this please, if its acceptable? It blocks other things that I want to do.

😁 

Although it would be great if you would not force push to existing PRs... It's impossible to keep track of changes and review them and I don't even know that there was a change. 😅

Waqar144:
The idea i am trying to implement is to separate the logic from the UI where possible. But it can't be done in one go, that'd be too error prone. So, I am trying to do something like this:

- separate out the individual components into their own classes
- make the components talk directly in cases where that's the origial intention instead of going via mainWindow
- Move out all the "globals/singletons" to a "kernel or engine" class

There is probably a lot more to this but often things become obvious once basics are done.

PR #2352, opened 2022-01-02
More litehtml fixes

Waqar144:


pbek:
Zoom is much better now.

pbek:
Checkboxes look cooler now too 😀

PR #2362, opened 2022-01-03
Fix defaultNoteFileExtension handling in highlighter

Waqar144:
We weren't ever setting _defaultNoteFileExtension which could lead to
bugs in some cases.

Some other related fixes:
- Note::defaultNoteFileExtension() is very expensive here. Used cached value.
- Introduce settingsChanged signal in mainWindow which is triggered
whenever settings change and can be used throughout the app to update
things dynamically
- remove setting mainWindow as a QProperty. instead just store a static
reference.
- Don't store the settings dialog. Create when needed.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -5197,7 +5197,7 @@ void MainWindow::updateCurrentFolderTooltip() {
  * Opens the settings dialog
  */
 void MainWindow::openSettingsDialog(int page, bool openScriptRepository) {
-    SettingsDialog *settingsDialog = new SettingsDialog(page, this);
+    QPointer<SettingsDialog> settingsDialog = new SettingsDialog(page, this);
     settingsDialog->readSettings();
 
     if (openScriptRepository) {
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -121,8 +121,14 @@
 #include "widgets/qownnotesmarkdowntextedit.h"
 #include "widgets/htmlpreviewwidget.h"
 
+static MainWindow* s_self = nullptr;
+
 MainWindow::MainWindow(QWidget *parent)
     : QMainWindow(parent), ui(new Ui::MainWindow) {
+
+    // static reference to us
+    s_self = this;
+
     // handle logging as signal/slot to even more prevent crashes when
     // writing to the log-widget while the app is shutting down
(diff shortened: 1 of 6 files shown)


src/mainwindow.cpp
```diff
@@ -208,8 +213,6 @@ MainWindow::MainWindow(QWidget *parent)
                          QStringLiteral(VERSION) + QStringLiteral(" - build ") +
                          QString::number(BUILD));
 
-    qApp->setProperty("mainWindow", QVariant::fromValue<MainWindow *>(this));
```
pbek, line 211: Why should this be removed? This is used quite regularly in QON!

pbek:
This PR smells like regressions. 😅  I don't know how I should test it...

src/mainwindow.cpp
Waqar144: There's a replacement. See instance()

Waqar144:
> This PR smells like regressions. 😅 I don't know how I should test it...

Why so? The changes are very simple I think.

pbek:
> remove setting mainWindow as a QProperty. instead just store a static
reference.

breaks everything that depends on it

> Don't store the settings dialog. Create when needed.

there was a reason for it and I think it was something more than just "it takes a lot of time to open the dialog the first time and it makes a lot of requests"...

src/mainwindow.cpp
pbek: And you did use the replacement on all occasions that currently depend on the property? 😅

Waqar144:
> breaks everything that depends on it

What depends on it? You mean there are things that use qApp->property to access mainWindow instead of instance() function?

Waqar144:
> there was a reason for it and I think it was something more than just "it takes a lot of time to open the dialog the first time and it makes a lot of requests"...

Seems like something that should be fixed rather than just never deleting the dialog

src/mainwindow.cpp
Waqar144: Why does anything depend on the property directly? We provide an interface to access mainWindow, no?

Waqar144: Found one direct usage of `qApp->property("mainWindow")` and replaced it.

Waqar144:
Added some more checks to ensure we don't crash. But if you can still crash it, let me know and we can fix that too.

Waqar144:
Add a fix for a crash in litehtml when you click ok in settings

PR #2376, opened 2022-01-12
Added information about installation via Scoop

altwarg:


pbek:
Who is maintaining https://github.com/ScoopInstaller/Extras/blob/master/bucket/qownnotes.json?

pbek:
I added some more information to your entry. Thank you for the addition.

Issue #2136, opened 2021-05-10
[F] Highlight items in the navigation panel according to the position of the cursor


fabianski7:
#### Expected behaviour
Navigation panel items could be highlighted depending on the location of the mouse cursor.

<details><summary>Screenshot</summary>

![ksnip_tmp_xEasJs](https://user-images.githubusercontent.com/47755037/117728936-a50c5300-b1d9-11eb-9bc4-038d0691423c.png)

</details>

#### Actual behaviour
Currently they are already highlighted with a nice effect, but this only happens when items in the panel are clicked, the editor is ignored.

<details><summary>Screenshot</summary>

![ksnip_tmp_fswFwT](https://user-images.githubusercontent.com/47755037/117728996-b81f2300-b1d9-11eb-8328-110a397a16ec.png)

</details>

#### Additional information
I saw this behavior in the new version of the ghostwriter, and it really helps navigating large files. Since QON is an AIO solution for markdown, it would be great to have that here too 😅

https://user-images.githubusercontent.com/47755037/117729702-c1f55600-b1da-11eb-9e88-dc5828ec7dc9.mp4

#### Output from the debug section in the settings dialog

<details><summary>Expand</summary>

```
**Version**: `21.4.6`
**Build date**: `Apr 26 2021`
**Build number**: `832`
**Platform**: `linux`
**Operating System**: `Arch Linux`
**Build architecture**: `x86_64`
**Current architecture**: `x86_64`
**Release**: `AUR`
**Qt Version (build)**: `5.15.2`
**Qt Version (runtime)**: `5.15.2`
**Portable mode**: `no`
```
</details>

fnkhz:
@pbek Do you think it would be a good idea to implement this as a script ?
I would like to start working on extending the scripting framework for this.

pbek:
## 22.1.8
- items in the navigation panel are now highlighted according to the position of
  the cursor in the note edit panel (for [#2136](https://github.com/pbek/QOwnNotes/issues/2136), thank you @fnkhz)

pbek:
There now is a new release, could you please test it and report if it works for you?

fabianski7:
I have version 21.1.8 with the auto select option enabled but it is not working.
Is it necessary to do anything else besides that?

Here is the QON debug info
https://bin.snopyta.org/?74f8eb548a1c2ddd#B75LQuga5b9kqEUXWTB9o8s2WtxEUTXVf37SGQ4wg6sr

[...]

PR #2383, opened 2022-01-15
Implement auto-selecting current item in navigation panel when current line changes

fnkhz:
Implements: #2136 
Requires PR: https://github.com/pbek/qmarkdowntextedit/pull/152

Can be implemented:
* as a script (providing a hook for current line changed in the markdowntextedit and a setter for setting the currently selected item in the navigation pane)
* as provided in this PR (maybe as an option)

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -3671,6 +3671,14 @@ bool MainWindow::jumpToNoteSubFolder(int noteSubFolderId) {
     return false;
 }
 
+bool MainWindow::selectNavigationItemAtPosition(int position)
+{
+    if (ui->navigationWidget->isVisible()) {
+        return ui->navigationWidget->selectItemAtPosition(position);
+    }
+    return false;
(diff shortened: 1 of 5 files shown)


pbek:
@Waqar144, any words of wisdom?

Waqar144:
Hi,

I think we should first decide "when" should we navigate. Option 1, navigate if you are on a header. Option 2, navigate always to the last found header from the cursor position. The second one is more useful in my opinion.

Waqar144:
Regarding the implementation, maybe just iterating through the tree to find the block or the closest block will suffice? We have QtreewidgetitemIterator and the position data seems to already be in the item.

fnkhz:
Indeed, if we want to do the scrolling to the heading when any content after the heading is selected, a binary search on the vector that we already have would be the best option; will make another PR today.

pbek:
Do we still need this PR?

PR #2384, opened 2022-01-16
Implement auto-selecting current item in navigation panel when current line changes

fnkhz:
Implements: https://github.com/pbek/QOwnNotes/issues/2136

This will highlight the navigation item corresponding to the current text cursor position, as opposed to PR https://github.com/pbek/QOwnNotes/pull/2383 , which requires that the text cursor is located on a heading line in order to select the corresponding navigation item.

Again, this functionality could be available as an option.

### DIFF
--- a/src/dialogs/settingsdialog.cpp
+++ b/src/dialogs/settingsdialog.cpp
@@ -1055,6 +1055,9 @@ void SettingsDialog::storePanelSettings() {
     settings.setValue(QStringLiteral("navigationPanelHideSearch"),
                       ui->navigationPanelHideSearchCheckBox->isChecked());
 
+    settings.setValue(QStringLiteral("navigationPanelAutoSelect"),
+                      ui->navigationPanelAutoSelectCheckBox->isChecked());
+
     settings.setValue(QStringLiteral("enableNoteTree"),
                       ui->enableNoteTreeCheckBox->isChecked());
 }
@@ -1649,6 +1652,9 @@ void SettingsDialog::readPanelSettings() {
     // Navigation Panel Options
     ui->navigationPanelHideSearchCheckBox->setChecked(
         settings.value(QStringLiteral("navigationPanelHideSearch")).toBool());
+
+    ui->navigationPanelAutoSelectCheckBox->setChecked(
+        settings.value(QStringLiteral("navigationPanelAutoSelect")).toBool());
 
     ui->enableNoteTreeCheckBox->setChecked(Utils::Misc::isEnableNoteTree());
 }
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -3671,6 +3671,13 @@ bool MainWindow::jumpToNoteSubFolder(int noteSubFolderId) {
     return false;
 }
 
+void MainWindow::selectNavigationItemAtPosition(int position)
+{
+    if (ui->navigationWidget->isVisible()) {
+        ui->navigationWidget->selectItemForCursorPosition(position);
+    }
+}
+
 QString MainWindow::selectOwnCloudNotesFolder() {
     QString path = this->notesPath;
 
@@ -4502,7 +4509,8 @@ void MainWindow::setNoteTextFromNote(Note *note, bool updateNoteTextViewOnly,
  */
 void MainWindow::startNavigationParser() {
     if (ui->navigationWidget->isVisible())
-        ui->navigationWidget->parse(activeNoteTextEdit()->document());
+        ui->navigationWidget->parse(activeNoteTextEdit()->document(),
+                                    activeNoteTextEdit()->textCursor().position());
 }
 
 /**
--- a/src/widgets/navigationwidget.cpp
+++ b/src/widgets/navigationwidget.cpp
@@ -84,7 +84,7 @@ void NavigationWidget::parse(const QTextDocument *document, int textCursorPositi
     Q_UNUSED(blocker)
 
     setDocument(document);
-    _latestTextCursorPosition = textCursorPosition;
+    _cursorPosition = textCursorPosition;
 
     const QFuture<QVector<Node>> future =
         QtConcurrent::run(&NavigationWidget::parseDocument, document);
@@ -118,7 +118,7 @@ QVector<Node> NavigationWidget::parseDocument(
 void NavigationWidget::selectItemForCursorPosition(int position)
 {
     if (_parseFutureWatcher->isRunning()) {
-        _latestTextCursorPosition = position;
+        _cursorPosition = position;
         return;
     }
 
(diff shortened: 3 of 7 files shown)


pbek:
@Waqar144, any thoughts?

Hm, builds are breaking. Are you basing the PR on another PR?

> Again, this functionality could be available as an option.

yes, this shouldn't be the default

src/widgets/navigationwidget.h
```diff
@@ -55,6 +58,8 @@ class NavigationWidget : public QTreeWidget {
     QHash<int, QTreeWidgetItem *> _lastHeadingItemList;
     QFutureWatcher<QVector<Node>> *_parseFutureWatcher;
     QVector<Node> _navigationTreeNodes;
+    int _latestTextCursorPosition;
```
Waqar144, line 61: nit: maybe just `_cursorPosition`?

src/widgets/qownnotesmarkdowntextedit.cpp
```diff
@@ -72,6 +72,13 @@ QOwnNotesMarkdownTextEdit::QOwnNotesMarkdownTextEdit(QWidget *parent)
     connect(this, &QOwnNotesMarkdownTextEdit::zoomOut, this, [this](){
         onZoom(/*in=*/ false);
     });
+    connect(this, &QOwnNotesMarkdownTextEdit::newLinePosition, this, [this](int position){
```
Waqar144, line 75: since this signal is QON specific, maybe we can just emit it in this class instead of in the library?

src/mainwindow.cpp
```diff
@@ -3671,6 +3671,13 @@ bool MainWindow::jumpToNoteSubFolder(int noteSubFolderId) {
     return false;
 }
 
+void MainWindow::selectNavigationItemAtPosition(int position)
```
Waqar144, line 3674: Would have been better if this could live in NavigationWidget, but I guess as things currently are, it will do.

src/widgets/navigationwidget.cpp
```diff
@@ -149,6 +198,7 @@ void NavigationWidget::onParseCompleted() {
         _lastHeadingItemList[elementType] = item;
     }
     expandAll();
```
Waqar144, line 200: maybe the expandAll can be dropped now since we auto move to the current node?

fnkhz:
Thank you for the review, will get back to this in the evening when I get some time. 👍

src/widgets/navigationwidget.cpp
```diff
@@ -114,9 +115,57 @@ QVector<Node> NavigationWidget::parseDocument(
...
+        return;
+    }
+
+    int itemIndex = findItemIndexforCursorPosition(position);
+
+    QTreeWidgetItem *itemToSelect{nullptr};
+    if (itemIndex >= 0) {
+        QTreeWidgetItemIterator it(this);
+        it += itemIndex;
+        itemToSelect = *it;
```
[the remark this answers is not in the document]
fnkhz: Setting the `currentItem` to `nullptr` would just deselect.

src/widgets/navigationwidget.cpp
fnkhz: Not sure about that; auto selection would only expand the current node hierarchy, not all the nodes.

fnkhz:
> Hm, builds are breaking. Are you basing the PR on another PR?

They should be fine now, rebased on develop.

src/widgets/navigationwidget.h
```diff
@@ -15,6 +15,7 @@
 
 #include <QFutureWatcher>
 #include <QTreeWidget>
+#include <QMap>
```
Waqar144, line 18: unnecessary include

src/widgets/navigationwidget.cpp
Waqar144: yeah, that's my point. We are auto expanding based on the cursor position, so maybe expandAll is not necessary anymore. Will also be better perf wise, since expand is a slow operation.

Either way is fine for me.

Waqar144:
looks good.

src/widgets/navigationwidget.h
pbek: @fnkhz, is this needed?

src/widgets/qownnotesmarkdowntextedit.cpp
```diff
@@ -73,6 +73,15 @@ QOwnNotesMarkdownTextEdit::QOwnNotesMarkdownTextEdit(QWidget *parent)
         onZoom(/*in=*/ false);
     });
 
+    connect(this, &QPlainTextEdit::cursorPositionChanged, this,
+            [this]() {
+        if (!mainWindow) {
+            qWarning() << "No MainWindow! shouldn't happen!";
+            return;
+        }
+        mainWindow->selectNavigationItemAtPosition(textCursor().block().position());
```
pbek, line 82: We still need a setting in the settings dialog for that (default on should be ok), I guess this feature will drive some people crazy. 😉

Waqar144: maybe lets wait till someone comes up who wants to turn it off.

pbek: I may be the first already... I'm not sure yet. 😆 
And as soon as other stuff (like links) are shown in the navigation panel then that feature needs to be turned off..

src/widgets/navigationwidget.h
fnkhz: No, sorry, forgot to remove it.

src/widgets/qownnotesmarkdowntextedit.cpp
fnkhz: Will be adding the setting tonight.

pbek: Thank you!

src/dialogs/settingsdialog.cpp
```diff
@@ -1650,6 +1653,9 @@ void SettingsDialog::readPanelSettings() {
     ui->navigationPanelHideSearchCheckBox->setChecked(
         settings.value(QStringLiteral("navigationPanelHideSearch")).toBool());
 
+    ui->navigationPanelAutoSelectCheckBox->setChecked(
+        settings.value(QStringLiteral("navigationPanelAutoSelect")).toBool());
```
pbek, line 1657: Let's set the default to `true`.

src/widgets/qownnotesmarkdowntextedit.cpp
```diff
@@ -79,7 +79,12 @@ QOwnNotesMarkdownTextEdit::QOwnNotesMarkdownTextEdit(QWidget *parent)
             qWarning() << "No MainWindow! shouldn't happen!";
             return;
         }
-        mainWindow->selectNavigationItemAtPosition(textCursor().block().position());
+        const bool autoSelect =
+            QSettings().value(QStringLiteral("navigationPanelAutoSelect"), false)
```
pbek, line 83: Here too... Let's set the default to `true`.

pbek:
@fnkhz, I just found out this PR breaks the logging panel. Every click in the logging panel results in a warning `No MainWindow! shouldn't happen!`.

fnkhz:
Will check it asap.

pbek:
The issue might have the same root as https://github.com/pbek/QOwnNotes/issues/2391 😅

fnkhz:
> @fnkhz, I just found out this PR breaks the logging panel. Every click in the logging panel results in a warning `No MainWindow! shouldn't happen!`.

Should be fixed with https://github.com/pbek/QOwnNotes/pull/2392

Issue #2389, opened 2022-01-19
qfilesystemwatcher will only trigger once if the file update method was deleted and created


clevertension:
[issues] see this link https://stackoverflow.com/questions/18300376/qt-qfilesystemwatcher-signal-filechanged-gets-emited-only-once
some external editor will first delete file and create a new one, so QOwnNotes can't detect the file update

#### Steps to reproduce
i can use MegaSync to reproduce it when the note sync from one computer to another

#### Solution
so i can give a solution, and test it is ok now:
while(note.fileExists()==false) {
   sleep 1 seconds and retry max 30 times
}
if(note.fileExists()) {
   we can reload the current note
}

pbek:
> i can use MegaSync to reproduce it when the note sync from one computer to another

Can you please describe step by step what you did to reproduce this issue (best without any other tool that would be needed to be installed)?

> sleep 1 seconds and retry max 30 times

you can't really "sleep", it would freeze the application... and to do that in the background would cause a lot of side-effects

clevertension:
just like  QCoreApplication.processEvents(QEventLoop:AllEvents, 100); process all events every 100ms, this can avoid freeze application

pbek:
> just like QCoreApplication.processEvents(QEventLoop:AllEvents, 100); process all events every 100ms, this can avoid freeze application

Let's discuss that in a Pull Requests...

So far I cannot reproduce your issue... I removed a note externally and the note folder was reloaded, than I created the same file again externally and the note folder was reloaded...

clevertension:
you can simulator with following code, it delete file and create it immediatelly


`
#include <QCoreApplication>
#include <QDebug>
#include <QFile>
#include <QTextStream>
int main(int argc, char *argv[])
{
    QCoreApplication a(argc, argv);
    //read file content
    QString filename = "/Users/dan/Documents/MEGA/QOwnNotes/database/Postgresql.md";
    QFile file(filename);
    file.open(QIODevice::ReadOnly);
    QString s;
    QTextStream in(&file);
    s.append(in.readAll());
    file.close();
    //delete file
    QFile ff(filename);
    ff.remove();
    //create a new one
    QFile file1(filename);
    if (file1.open(QIODevice::ReadWrite)) {
        QTextStream stream(&file1);
        stream <<s<<"---->add something" << endl;
    }
    return a.exec();
}
`

pbek:
[...]

PR #2390, opened 2022-01-19
fix issue 2389

clevertension:
fix the issue [qfilesystemwatcher will only trigger once if the file update method was deleted and created]

referer https://github.com/pbek/QOwnNotes/issues/2389

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2792,12 +2792,13 @@ void MainWindow::notesWereModified(const QString &str) {
 
     QFileInfo fi(str);
     Note note = Note::fetchByFileName(fi.fileName());
-
+    delayCheckNotesExists(note);
+    if (note.fileExists()) {
+        noteDirectoryWatcher.addPath(note.fullNoteFilePath());
+    }
     // load note from disk if current note was changed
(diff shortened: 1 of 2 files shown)


pbek:
@Waqar144, can you reproduce #2389 and do you think this PR is a good idea?

src/mainwindow.cpp
```diff
@@ -2766,6 +2766,16 @@ void MainWindow::updateNoteTextFromDisk(Note note) {
     ScriptingService::instance()->onCurrentNoteChanged(&currentNote);
 }
 
+void MainWindow::delayCheckNotesExists(const Note &note) {
+    int max_checks = 10;
+    while(!note.fileExists() && max_checks-->0) {
+        QTime sleepTime = QTime::currentTime().addSecs(1);
+        while(QTime::currentTime() < sleepTime) {
+            QCoreApplication::processEvents(QEventLoop::AllEvents, 100);
```
Waqar144, line 2774: This almost always is a really bad idea.

A better way might be to do it using QTimer's and recursion with a limit or do it in a separate thread.

clevertension:
i test the latest code, it can detect the file change, so this PR can be closed

PR #2392, opened 2022-01-19
Fix logging panel issue caused by #2136

fnkhz:
Fix logging panel issue caused by #2136 implementation


pbek:
The issue is that the logging panel is not getting the mainWindow like it possibly should since https://github.com/pbek/QOwnNotes/commit/c48eae67c5b8173846288faddc0d3d5197b31dd6...

`ui->encryptedNoteTextEdit` would need the same treatment. Maybe don't use a lambda function, but a named slot and connect it to both text edits...

pbek:
Thank you!

PR #2399, opened 2022-01-20
Clean up logWidget

Waqar144:
Remove all qproperty based checks. Instead become a child of mainwindow
to have a somewhat clear ownership semantics. This will prevent using
extra hacks to handle cases where we are getting destroyed and logging.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -131,13 +131,11 @@ MainWindow::MainWindow(QWidget *parent)
     // static reference to us
     s_self = this;
 
-    // handle logging as signal/slot to even more prevent crashes when
-    // writing to the log-widget while the app is shutting down
-    connect(this, &MainWindow::log, LogWidget::instance(), &LogWidget::log);
+    _logWidget = new LogWidget(this);
+    connect(this, &MainWindow::log, _logWidget, &LogWidget::log);
 
     // use our custom log handler
     qInstallMessageHandler(LogWidget::logMessageOutput);
-    qApp->setProperty("loggingEnabled", true);
 
 #ifdef Q_OS_MAC
     // disable icons in the menu
@@ -724,14 +722,6 @@ MainWindow::~MainWindow() {
     // (#1269, may cause an interruption of the shutdown process)
     //    gitCommitCurrentNoteFolder();
 
-    qApp->setProperty("loggingEnabled", false);
-
-    if (showSystemTray) {
-        // if we are using the system tray lets delete the log window so the
-        // app can quit
-        delete (LogWidget::instance());
-    }
-
     delete ui;
 
     s_self = nullptr;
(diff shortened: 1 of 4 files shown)


src/mainwindow.cpp
```diff
@@ -724,14 +721,6 @@ MainWindow::~MainWindow() {
     // (#1269, may cause an interruption of the shutdown process)
     //    gitCommitCurrentNoteFolder();
 
-    qApp->setProperty("loggingEnabled", false);
-
-    if (showSystemTray) {
-        // if we are using the system tray lets delete the log window so the
```
pbek, line 730: Will this still work?

src/mainwindow.cpp
```diff
@@ -131,13 +131,10 @@ MainWindow::MainWindow(QWidget *parent)
     // static reference to us
     s_self = this;
 
-    // handle logging as signal/slot to even more prevent crashes when
```
pbek, line 134: And there will be no crashes because something gets logged while shutting down?

src/mainwindow.cpp
Waqar144: I tried this case especially + a couple of ways to quit and so far nothing crashed.

src/mainwindow.cpp
Waqar144: One case to try this (haven't done so myself yet,) would be to log in the mainWindow destructor. If it survives that, we are safe.

Waqar144: As long as the widget is alive, it shouldn't crash. The Qpointer usage is here for exactly this reason. On destruction it should become null and no further log calls will be processed.

Waqar144: Tried it, works okay. Here's a log ( I added an else case where we call `log()`).

```
Warning: deleting mainwindow (/home/waqar/projects/QOwnNotes/src/mainwindow.cpp:726, virtual MainWindow::~MainWindow())
Got log request
Warning: Deleting log widget (/home/waqar/projects/QOwnNotes/src/widgets/logwidget.cpp:99, virtual LogWidget::~LogWidget())
Got log request
Debug: deleting 0x560d8cb97b68 for "en_US"
log widget is dead now
Debug: deleting 0x560d8c726f20 for "en_US"
log widget is dead now
Debug: Removing loader:  Sonnet::Loader(0x560d8c0de720)
log widget is dead now
```

"Log widget is dead now" this is the else case, where the log widget has long gone.

pbek:
I can't see any status or scripting log messages that should be shown... Are you sure everything is handled like it was before? 😅

pbek:
Reloading the scripting engine for scripting log entry. Automatic script update on startup for status log entry.

Waqar144:
> Reloading the scripting engine for scripting log entry. Automatic script update on startup for status log entry.

Done, should work now.

pbek:
I'll be keeping my fingers crossed 😆

Waqar144:
> I'll be keeping my fingers crossed 😆

Let's see, I am positive about this ;)

pbek:
> Let's see, I am positive about this ;)

OMG. This PR breaks the note search! How is this even possible!

pbek:
(I just bisected my way to this PR)

pbek:
And notes aren't stored any more when you switch to another note... and you can't create notes from the note search lineedit... 🙄

pbek:
https://github.com/pbek/QOwnNotes/pull/2406 fixes some issues. But only some!

Waqar144:
> #2406 fixes some issues. But only some!

what else is there?

pbek:
"So far" I can only see https://github.com/pbek/QOwnNotes/issues/2412

Waqar144:
can't repro that.

Issue #2398, opened 2022-01-20
Enter key dos not go to new paragraph in markdown


SBaldrick:
Hello,

I seem to have a problem with the enter key in markdown. In markdown pannel, the behaviour is normal (New line). But in the preview pannel, it looks like the "paragraph" key has not been hit, all is queued on the same line. 

I can show you a screen cap if necessary.

Thanks for your help

pbek:
A newline in markdown isn't a line break in html (the preview). Only a new paragraph will do that (2 newlines).
Or you can enter two spaces at the end of the line (`Shift + Enter` will do that for you).

pbek:
Also see: https://www.qownnotes.org/getting-started/markdown.html#line-breaks

SBaldrick:
That seems to have done the trick perfectly ! Thank you !!

Although the markdown guide (your link) does say : "One newline will get you a new line in the same paragraph" which was what I'm going for. But maybe there's a subtlety there that I don't see, I'm new at this :p

pbek:
You are right, I have no idea what that means. I'll rewrite that. Thank you.

pbek:
> **Two newlines** will get you a **new paragraph**.

Or maybe the old markdown2html library did exactly that... 🤔

noureddin:
Sorry for replying to a closed issue.

Honestly I see "a new line in the same paragraph" is ambiguous and can mean either a soft break (new line in the editor but not in the preview) or a hard break (new line in both).

I guess more rewording is needed. I suggest something like:

> - Single newlines inside paragraphs allow you to divide your paragraphs to more lines for easier editing, without affecting the preview; in the preview, **line breaks inside a paragraph are ignored**.

Or something like (with the next bullet):

> - You can break a paragraph into more than a single line for easier editing; they still render as a single paragraph with no breaks.
> - You can force a line break inside a paragraph by ending a line with two spaces.

But then also the first bullet would need to change to maintain parallelism (starting the sentence with `You can` or similar).

I'm just suggesting a general idea; feel free to reword better.

(Or this should be a new issue? 😅)

pbek:
> Sorry for replying to a closed issue.

that's quite fine with me

> Honestly I see "a new line in the same paragraph" is ambiguous and can mean either a soft break (new line in the editor but not in the preview) or a hard break (new line in both).

[...]

PR #2403, opened 2022-01-21
[website] (re)mention paragraph breaks

noureddin:
#2398 mentioned that the documentation about line breaks is ambiguous. It was discussed and improved in that issue, and this PR continues improving it.

The last commit discussed in that issue had improved the wording about soft line breaks and hard line breaks (the second and third bullets [originally](https://github.com/pbek/QOwnNotes/blob/3215c4984f4f4bf0e6fabfdd87990e7ee9a86099/webpage/src/getting-started/markdown.md#line-breaks)), but it removed the first bullet, which was about paragraph breaks.

This PR tries to re-incorporate that in the same style of the new bullets.


pbek:
Thank you!

Issue #790, opened 2017-11-26
Feature request: ability to show notes in Subfolders panel


mpodshivalin:
Let's imagine that we have a following folder/notes structure:
```
Notes
|-- Markdown Cheatsheet (note)
|-- Work (folder)
    |-- Passwords (note)
    |-- Contacts (note)
    |-- Active work (folder)
        |-- ToDo List (note)
        |-- Meeting agenda 26.11.2017 (note)
    |-- Archive work (folder)
        |-- Meeting 20.01.2017 (note)
        |-- Meeting 20.02.2017 (note)
        |-- Meeting 20.03.2017 (note)
|-- Life (folder)
    |-- ToDo List (note)
    |-- Trips (folder)
        |-- 12.2017 trip (note)
        |-- 04.2018 trip (note)
    |-- ...
|-- Random documents
    |-- How to generate SSH keys (note)
    |-- ...
```

Currently we have 2 separate panels to switch between notes:
* Subfolders panel
* Note list panel

So I'll see this interface:
![image](https://user-images.githubusercontent.com/16121630/33241390-9f0ba550-d2d5-11e7-9e42-69b07105ee78.png)

If I need to switch from work ToDo List to life ToDo List to quickly add something, I'll have to:
* Navigate to Life (folder)
* Open ToDo List (note)
* Navigate back to Active work (folder)
* Open ToDo List (note)

This is not very convenient not only because of the number of actions, but because I have to explicitly remember the name of the folder I need to navigate to. I have tons of notes in different folders, so using "All notes" or history wouldn't be convenient either, so I propose to add the ability to show notes directly in Subfolders panel (thus making it possible to use only one panel instead to two), so it would look like this:

![image](https://user-images.githubusercontent.com/16121630/33241569-6300740c-d2d8-11e7-8f6b-3ac0b6c2f7fe.png)

Maybe it's not to everyone's taste, but in my case switching between notes would be much easier this way.

pbek:
I'm sorry, that didn't work out for 1000s of notes...

But you either can:

- use the bookmarking function to switch between notes (take a look at the keyboard shortcuts)
- use the keyboard keys or mouse keys to jump forward or backward in the note history
- use the note view dialogs to view multiple notes if you don't want to edit them

pbek:
## 19.10.11
- you now can choose to **show your notes together with your subfolders in a tree**
    - you can enable this in the *Panels settings*
    - this is work in progress and by far not fully implemented yet!
[...]

PR #2410, opened 2022-01-23
Fix noteTreeWidget context menu when clicking on folder item

Waqar144:


pbek:
What issue is this PR fixing?

Waqar144:
> What issue is this PR fixing?

There is no reported issue for it

pbek:
And builds are failing on https://github.com/pbek/QOwnNotes/runs/4913757419?check_suite_focus=true

pbek:
> There is no reported issue for it

So please describe the issue 😅

Waqar144:
Missing include `<memory>`

Waqar144:
Switch to tree mode note tree widget. Then try to use the context menu. You will see

Waqar144:
Does there always need to be an issue for PRs now?

pbek:
> Switch to tree mode note tree widget. Then try to use the context menu. You will see

You didn't mention that before it is about https://github.com/pbek/QOwnNotes/issues/790.

> Does there always need to be an issue for PRs now?

It would be great, but not in this case, because it seems you are talking about https://github.com/pbek/QOwnNotes/issues/790.

Waqar144:
Yes, that. But it broke during the refactoring, so this is a fix for that.

pbek:
> Missing include <memory>

Will you add the `include <memory>` so I can see if it runs through?

Waqar144:
Yes, but will have to wait until tomorrow when I get back to my pc.

Waqar144:
Ok build passed now.

pbek:
Our main build is down... https://github.com/pbek/QOwnNotes/issues/2415 😬

PR #2423, opened 2022-01-29
Fix #2421

### LINKED ISSUE #2421: "Don't ask again" on "allow note editing" makes the pressed key that allows it later always consumed
Using QOwnNotes 22.1.11.1.

#### Steps to reproduce

1. Disallow note editing (press Esc, or from the Note menu).
2. In the editor, type any letter.
3. Appears a dialog that informs you about the read-only mode and asks if you want to enable editing.
4. Tick the "Don't ask again" and press "OK".
5. Disallow note editing, again.
6. In the editor, type any letter.
7. Note editing is automatically re-enabled.

#### Expected behaviour

8. The letter you typed should appear in the editor where the cursor is.

#### Actual behaviour

8. Nothing. The letter only enables note editing but is forgotten. You have to type it again.

noureddin:
This should close #2421.

Two notes:

1. I'm not particularly happy about the WET code, i.e., all of the added lines except the comments and the `eventFilter` call. But I don't know a better way.
2. I only resend the event if the message box was overriden (to Yes), not if the answer is Yes in general. I think this is less surprising for the users.

### DIFF
--- a/src/widgets/qownnotesmarkdowntextedit.cpp
+++ b/src/widgets/qownnotesmarkdowntextedit.cpp
@@ -1201,6 +1201,21 @@ bool QOwnNotesMarkdownTextEdit::eventFilter(QObject *obj, QEvent *event) {
                         if (mainWindow != Q_NULLPTR) {
                             mainWindow->allowNoteEditing();
                         }
+                        // If the answer is overriden to Yes ("Don't ask again" with "Yes"),
+                        // what you type then only enables note editing, but is not typed in
+                        // the editor. We need to re-send the event after enabling editing.
+                        // BUT, we should do that only if the msgbox is overriden to Yes,
+                        // not if manually answered.
+                        // You may see: https://github.com/pbek/QOwnNotes/issues/2421
+                        // This check is partially copied from utils/gui.cpp showMessage()
+                        QSettings settings;
+                        const QString settingsKey = QStringLiteral("MessageBoxOverride/readonly-mode-allow");
+                        auto overrideButton = static_cast<QMessageBox::StandardButton>(
+                            settings.value(settingsKey, QMessageBox::NoButton).toInt());
+                        if ((overrideButton == QMessageBox::Yes)) {
+                            // overriden to answer yes: re-send the event
+                            return QMarkdownTextEdit::eventFilter(obj, event);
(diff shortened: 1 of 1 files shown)


pbek:
@Waqar144, are you fine with this?

src/widgets/qownnotesmarkdowntextedit.cpp
```diff
@@ -1230,6 +1230,21 @@ bool QOwnNotesMarkdownTextEdit::eventFilter(QObject *obj, QEvent *event) {
...
                         }
+                        // If the answer is overriden to Yes ("Don't ask again" with "Yes"),
+                        // what you type then only enables note editing, but is not typed in
+                        // the editor. We need to re-send the event after enabling editing.
+                        // BUT, we should do that only if the msgbox is overriden to Yes,
+                        // not if manually answered.
+                        // You may see: https://github.com/pbek/QOwnNotes/issues/2421
+                        // This check is partially copied from utils/gui.cpp showMessage()
+                        QSettings settings;
+                        const QString settingsKey = "MessageBoxOverride/readonly-mode-allow";
```
Waqar144, line 1241: Wrap with qstringliteral()

Waqar144:
+1

For me, even if we add a letter for the case where you manually clicked Yes would be okay. After all, you typed the letter and probably expected it to be there in the editor but forgot that it was in read only mode, but others might have different expectations.

noureddin:
I had several issues because I by mistake made my edits on an older commit. Long story short, I re-made the commit, with QStringLiteral, to a new branch with the same name, and force-pushed it. 😇

I also recompiled locally to make sure I didn't mess up the build.

.

> For me, even if we add a letter for the case where you manually clicked Yes would be okay. After all, you typed the letter and probably expected it to be there in the editor but forgot that it was in read only mode, but others might have different expectations.

I may be wrong, but I have two situations in mind:

1. A speed typist: types several letters before noticing the message box, only the first is re-send later. I think this is confusing.
2. A normal person: presses a letter, and a message box shows up. My expectation in such a situation would be that what is typed is not registered at all; it only made the message box appear.

That being said, I don't have a strong opinion on this issue. I just felt it's more natural that way. But I may wrong.

PR #2426, opened 2022-01-31
Changed danish language file

DotNetCoderDk:
Changed some of the danish translations so they are not worded as machine translations would sound


pbek:
I'm sorry, these files are generated automatically with Crowdin as source. 😅  You would need to enter it there. See https://www.qownnotes.org/contributing/translation.html.

DotNetCoderDk:
Crud... I will try to do it some time in the future.

pbek:
Thank you! 😁

PR #2438, opened 2022-02-06
Remove encryptedNoteTextEdit

Waqar144:
It is just unecessary code and complexity. Remove it. The main
noteTextEdit can handle everything.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -212,8 +212,6 @@ MainWindow::MainWindow(QWidget *parent)
     sortingOrder->addAction(ui->actionDescending);
     sortingOrder->setExclusive(true);
 
-    // hide the encrypted note text edit by default
-    ui->encryptedNoteTextEdit->hide();
     ui->multiSelectActionFrame->hide();
 
     // initialize the tag button scroll area
@@ -508,17 +506,13 @@ void MainWindow::initNotePreviewAndTextEdits()
     // set the search frames for the note text edits
     const bool darkMode = settings.value(QStringLiteral("darkMode")).toBool();
     ui->noteTextEdit->initSearchFrame(ui->noteTextEditSearchFrame, darkMode);
-    ui->encryptedNoteTextEdit->initSearchFrame(ui->noteTextEditSearchFrame,
-                                               darkMode);
 
     // set the main window for accessing its public methods
     ui->noteTextEdit->setMainWindow(this);
-    ui->encryptedNoteTextEdit->setMainWindow(this);
 
     // setup vim mode
     if (settings.value(QStringLiteral("Editor/vimMode")).toBool()) {
         initFakeVim(ui->noteTextEdit);
-        initFakeVim(ui->encryptedNoteTextEdit);
     }
 
     // do the navigation parsing after the highlighter was finished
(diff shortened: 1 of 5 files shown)


pbek:
> The main noteTextEdit can handle everything.

That would be really!
But did you actually test your PR? 😅 

Try encrypting a note, then change the text and then jump to another note and back to your note... The added text is gone...
I edited it again and then the whole encryption was gone. 😆

Waqar144:
Yeah, i tested.

- Encrypted a note
- Jumped to other notes, made edits
- Jumped back to encrypted one => everything working.

Waqar144:
Tried:

- Encrypt note
- "Edit Encrypted Note" => auto decrypts + shows the text
- Changed text
- jumped elsewhere
- jumped back => note is there but not encyrpted.

Can't seem to see anything worng?

Waqar144:
Btw, this behaviour of storing the password is really bad, from a security POV. Password should never be stored in memroy, *at all*. Once used, the memory should be purged.

pbek:
I can't even see any more when I'm editing an encrypted note (red border is gone)... and I loose either entered text or the encryption (buttons to newly encrypt note is active again) or both...

When I edit an encrypted note kate (with the note opened in it) shows me the plain text after a save!

pbek:
> Btw, this behaviour of storing the password is really bad, from a security POV. Password should never be stored in memroy, at all. Once used, the memory should be purged.

Yes, I don't like it neither, but it's very inconvenient to not have it even for a short amount of time, you wouldn't even be able to edit a note without entering it again and again the way it currently is implemented... If I'd like to heave real security I'd use the e.g. PGP script for the encryption or something else...

pbek:
...newline will be added silently with the next release.

Waqar144:
> > Btw, this behaviour of storing the password is really bad, from a security POV. Password should never be stored in memroy, at all. Once used, the memory should be purged.
> 
> Yes, I don't like it neither, but it's very inconvenient to not have it even for a short amount of time, you wouldn't even be able to edit a note without entering it again and again the way it currently is implemented... If I'd like to heave real security I'd use the e.g. PGP script for the encryption or something else...

Security comes with some loss in convenience. Even if you are going to store the password, this way is incorrect. Every OS provides some abstraction for a keystore. That should be used.

Waqar144:
> I can't even see any more when I'm editing an encrypted note (red border is gone)... and I loose either entered text or the encryption (buttons to newly encrypt note is active again) or both...
> 
> When I edit an encrypted note kate (with the note opened in it) shows me the plain text after a save!

What red border? I have never seen any red border in QON ever... with encrypted notes or otherwise

Waqar144:
Should work as before now

pbek:
> [](https://github.com/Waqar144)@[Waqar144](https://github.com/Waqar144) Waqar144 [force-pushed]

please no force pushes... what is the changes list of your commit?

pbek:
>  Even if you are going to store the password, this way is incorrect. Every OS provides some abstraction for a keystore. That should be used.

The password isn't persisted, it's "in-memory"... But using a keystore to store it permanently would be nice too, but it has two disadvantages.

1) if there is no keystore or the user doesn't use it QON encryption is broken or crippled
2) I dunno if using a keystore should enforce decryption.. I wouldn't like that. so better not, but that can be done

Waqar144:
> > @[Waqar144](https://github.com/Waqar144) Waqar144 [force-pushed]
> 
> please no force pushes... what is the changes list of your commit?

1. I don't like squash merges when I didn't do it myself
2. I had to rebase because of conflicts, so I needed to force push.

pbek:
Can we do the same styling as before?
1) current
2) your PR

image

Waqar144:
If one tries hard enough, yeah that can be done

pbek:
And I still have the issue when I switch to another note and back a couple of times then the encryption is gone and kate shows the unencrypted note.

pbek:
And sometimes the encrypted note gets locked while editing... I cannot enter anything any more.

pbek:
> Though I personally want that it should just lock itself on tab / note change. Once you are back, you can manually reopen the note for editing.

No, I meant I saw the encrypted note in "edit mode" but wasn't able to type any more

Waqar144:
Ah, hmm. I thought I fixed that. Will see.

PR #2449, opened 2022-02-10
Remove levenshtein similarity check

Waqar144:
It's slowing everything down too much.

https://github.com/pbek/QOwnNotes/issues/2433

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2790,14 +2790,6 @@ void MainWindow::notesWereModified(const QString &str) {
             const bool isCurrentNoteNotEditedForAWhile =
                 this->currentNoteLastEdited.addSecs(60) <
                 QDateTime::currentDateTime();
-            // If the current note wasn't edited for a while, we want that it is possible
-            // to get updated even with small changes, so we are setting a threshold of 0
-            const int threshold = isCurrentNoteNotEditedForAWhile ? 0 : 8;
-
-            // Check if the old note text is the same or similar as the one on disk
-            if (Utils::Misc::isSimilar(oldNoteText, noteTextOnDisk, threshold)) {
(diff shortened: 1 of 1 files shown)


pbek:
I don't really want to remove this until there is another solution. It works great for almost all notes to get around those "somethings gets updated by an external process while or after you were editing a note" issues.

Waqar144:
> I don't really want to remove this until there is another solution. It works great for almost all notes to get around those "somethings gets updated by an external process while or after you were editing a note" issues.

For tiny notes, yes.

pbek:
> For tiny notes, yes.

I wouldn't call a few 100kB tiny... If you want we can add a simple size check to decide if we want the similarity check...

Waqar144:
On linux, I haven't encountered any file watcher issues (probably because I don't sync and change things not very often). But on other systems its a common issue. And so I can't imagine the pain everyone would be going through everytime this code path is triggered (which is a lot given the number of externally modified issues here). This is completely broken for anything larger than a couple of kilobytes

pbek:
I wouldn't call 93.8 KB tiny neither for plain text files. All my notes are below 93.8 KB! Even the QOwnNotes Todo list...

Waqar144:
93K is tiny. You may have different standards, but for a computer that can process billions of instructions per second and has a throughput of multi GB/s, this is nothing.

pbek:
You are right, I re-evaluated. It takes too long even in single digit kB notes.

Waqar144:
I think `QSignalBlock`ing the watcher doesn't work well. On the other hand, if we disconnect -> reconnect, that is a lot more reliable as there will be no signals coming through.

For cases where there really was an external modification by an external program triggered by the user, the dialog is necessary.

Issue #2451, opened 2022-02-10
[I] Custom toolbar configurations not saved


dohliam:
Issue: Configurations and customizations of the toolbar do not persist after restart.

#### Expected behaviour

Changes such as adding and removing buttons from the toolbars should persist across sessions/restarts. New toolbars created through the toolbar settings window should have their configurations saved rather than disappearing.

#### Actual behaviour

Toolbar settings are not saved after restart. Buttons added to toolbars disappear/reset across sessions. Changes do not persist and all configurations and menus are reset after restarting.

#### Steps to reproduce

This issue can be easily reproduced on a fresh install of QOwnNotes and the latest version of Manjaro (KDE Plasma edition):

1. Install [latest Manjaro Pahvo KDE](https://manjaro.org/downloads/official/kde/).
2. Install QOwnNotes following the official Arch instructions [here](https://www.qownnotes.org/installation/arch.html).
3. Go to Settings > Scripting > Add script... > Search Script Repository and install the script [Headline depth buttons](https://github.com/qownnotes/scripts/tree/master/headline-depth) (or any script with custom actions -- this is just an example).
4. Go to Settings > Toolbars and click `New` to create a new toolbar which will be named `custom_toolbar_1` by default. In the same settings window, select `Custom Actions` from the dropdown menu on the left, choose the button from the menu labelled `#+` and click on `Insert` to add it to `custom_toolbar_1`.
5. At this point, clicking on the `Apply toolbar configuration` button does not make a difference to the result, so click it (or not).
6. Click `OK` to close the window. Observe that the new Headline depth button has been added to the toolbar at the top right.
7. Click on the `Quit` button to exit the app.
8. Restart QOwnNotes. Note that the toolbar has been reset to its previous configuration and the custom action button has disappeared.
9. Repeat with different toolbar buttons. None of them persist after a restart.

This has also been tested on KDE Neon and on several different machines with the same results. It does not seem to be possible to get the toolbar to retain any settings across restarts.

#### Output from the debug section in the settings dialog

<details><summary>Expand</summary>

QOwnNotes Debug Information
===========================

## General Info

**Current Date**: `Thu Feb 10 21:13:25 2022`
**Version**: `22.2.3`
**Build date**: `Feb 10 2022`
**Build number**: `947`
**Platform**: `linux`
**Operating System**: `Manjaro Linux`
**Build architecture**: `x86_64`
**Current architecture**: `x86_64`
**Release**: `OBS`
[...]

PR #2452, opened 2022-02-13
Fix crash in nav widget

Waqar144:
The doc can change while we are parsing it resulting in crashes.


src/widgets/navigationwidget.cpp
```diff
@@ -83,12 +70,19 @@ void NavigationWidget::parse(const QTextDocument *document, int textCursorPositi
...
     _cursorPosition = textCursorPosition;
+}
 
-    const QFuture<QVector<Node>> future =
-        QtConcurrent::run(&NavigationWidget::parseDocument, document);
-    this->_parseFutureWatcher->setFuture(future);
+void NavigationWidget::doParse()
+{
+    _delay.stop();
+    qWarning() << "Parsing now!";
```
pbek, line 83: Do you need this warning?

pbek:
There now is a 1.4. to 1.8sec delay for the navigation widget to render when switching notes. Was that on purpose?

Waqar144:
> > @[Waqar144](https://github.com/Waqar144) Waqar144 [force-pushed](https://github.com/pbek/QOwnNotes/compare/30eecf49a89bb64348e870bf98ce530fd9035f09..0ec40f7aa4e04d0fe4e8698a99efb269d168f21a) the work/nav-parser-fix branch from [30eecf4](https://github.com/pbek/QOwnNotes/commit/30eecf49a89bb64348e870bf98ce530fd9035f09) to [0ec40f7](https://github.com/pbek/QOwnNotes/commit/0ec40f7aa4e04d0fe4e8698a99efb269d168f21a) 2 hours ago
> 
> please no force pushes in PRs (or I need to review the whole PR again every time)

I will say it one last time. I will force push when I think it's better.

Waqar144:
> There now is a 1.4. to 1.8sec delay for the navigation widget to render the new tree when switching notes. Was that on purpose?

Yes, to avoid triggering it too much which is just wasteful. It will update once you stop typing.

pbek:
> I will say it one last time. I will force push when I think it's better.

What is better (or when it is better) when you force-push? And how I am supposed to review the changes between your force pushes?

pbek:
> Yes, to avoid triggering it too much which is just wasteful. It will update once you stop typing.

No, I didn't mean typing, that's ok. I meant when switching between notes...

Waqar144:
> > Yes, to avoid triggering it too much which is just wasteful. It will update once you stop typing.
> 
> No, I didn't mean typing, that's ok. I meant when switching between notes...

you can change the delay to whatever you like, I set it to 1 second for now.

pbek:
> you can change the delay to whatever you like, I set it to 1 second for now.

for typing the delay is great.
But I'd like no delay when browsing notes, like it was before.

Waqar144:
Removed the delay completely. Now it will run more but there shall be no delays.

Waqar144:
> What is better (or when is it better) when you force-push? 

When the change I am making belongs to the commit that is already there, and is not a standalone change.

Waqar144:
> And how I am supposed to review the changes between your force pushes?

The same way I do yours :/

pbek:
> The same way I do yours :/

When do I make a force push and when you do need to review it?

Waqar144:
> > The same way I do yours :/
> 
> When do I make a force push and when you do need to review it?

Never. And I have never reviewed anything by you I think.

Waqar144:
So, Is there something missing in the PR? Or are you purposely delaying it for no reason?

pbek:
I haven't tested it again. And I need to look it through from the start because of the force pushes. Need more time.

pbek:
Most of my spare time currently goes into support, https://github.com/qownnotes/web-companion/issues/38 and https://github.com/qownnotes/web-companion/issues/37. Reviewing things over and over again (extra) takes time...

Waqar144:
Generally when you comment at some place, I usually will only update that part.

If there need to be new changes, then I usually try to add them in a new commit.

Besides, you don't have to review so hard, since things break anyways and I am always fixing them. But I guess you have trust issues so yeah.

pbek:
Since 22.1.0 QOwnNotes did become very unstable and got a bad reputation for being so. I had to deal with hours and hours of support requests on all kind of channels every day and week, testing, reproducing, finding out where the issues are, reviewing your changes, finding "Easter-eggs" in those changes, preparing releases, ...

If you want to help with e.g. support you can start answering issues, like https://github.com/pbek/QOwnNotes/issues/2451.

Waqar144:
Well, when you make releases with zero testing after 1 hour then that's bound to happen, especially for changes that are large there are going to be bugs.

In any case, let me know of any changes that I have made but ignored the bugs and regressions that were caused by it, because it seems like you think I am just causing bugs and not helping anymore. And even though you have been doing support, close to 90% of the Qt6 porting work is done by me and a large number of features are written by me, but that doesn't count as help I guess. Also I think I will not make any new features, optimizations, refactorings or porting anymore since it makes things unstable for you and gives bad reputation. But if there are any bugs that I have caused I will take a look, and that will be it.

pbek:
Thank you! I'm sorry too. I'm under a lot of stress too in my non-dev life lately and now a lot of extra stress with the red alerts of https://github.com/qownnotes/web-companion/issues/38 and https://github.com/qownnotes/web-companion/issues/37.

sanderboom:
Happy you guys reconciled :wink: You both do great work! :rocket: 

Probably this has been discussed before, but to lower stress on both the development-side and user-side: why not use a dev and a stable channel? Have a small group test dev continuously and cut a release for stable every month. 

Anyway, thanks for all the effort :handshake:

pbek:
> why not use a dev and a stable channel?

because that would be a crazy amount of work to streamline that for all build processes and the users...
but for example we already build a windows version as build artifact on every commit

Waqar144:
If we delay releases, who will test it? But I agree on some delay before releasing, and maybe some tests to harden things a bit more.

sanderboom:
> who will test it

At least us three :wink: and maybe other enthusiastic users.

I'm thinking of Sublime Text's dev channel (https://www.sublimetext.com/dev). I packaged this for AUR and so I know it was quite populair. 

> because that would be a crazy amount of work

:grimacing: I bet

pbek:
This PR broke some builds on https://build.opensuse.org/package/show/home:pbek:QOwnNotes/desktop, including the AppImage.

PR #2456, opened 2022-02-14
Re-enable filewatch with a delay

Waqar144:
FileWatchDisabler isn't enough on its own for some cases and we still
get externally modified signals.

Case that this fixes:

Create a note with a single checkbox item on the first line and
check/uncheck it from the note preview. This triggers file rename which
shouldn't result in externally modified notifications. With this change,
that should be fixed.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -128,7 +128,7 @@ struct FileWatchDisabler {
         : _mainWindow(mw)
     {
         Q_ASSERT(mw);
-        mw->disconnect(&mw->noteDirectoryWatcher, nullptr, nullptr, nullptr);
+        QObject::disconnect(&mw->noteDirectoryWatcher, nullptr, nullptr, nullptr);
     }
 
     ~FileWatchDisabler()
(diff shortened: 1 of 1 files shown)


pbek:
Hard to test this PR.
QOwnNotes currently crashes with our without this PR when I build QOwnNotes with clion when I add or remove notes externally. 😆 

In the current release 22.2.5 externally added notes are currently not even detected. 🙈  Do you have the same issue?

Waqar144:
> [](https://github.com/pbek)In the current release 22.2.5 externally added notes are currently not even detected (deleted are). see_no_evil Do you have the same issue?

Hmm, no. I just noticed this while trying out the checkbox feature. Try creating a note like this:

```
- [ ] first
```

Then check/uncheck using note preview => causes rename, which is expected and we shouldn't get a `filesChanged` signal. For normal users this would not be noticeable as you don't get the dialog because file name has changed i.e., it falls in the `else {` case of `notesWereModified` function.

But anyways, for the crash do you have a backtrace?

Waqar144:
One other thing, this issue will only reproduce if you have auto save timer set to a small value like 2 seconds or less (I think, not sure)

pbek:
> But anyways, for the crash do you have a backtrace?

Nothing very usable...

```
QHashData::nextNode 0x00007fdfba713ba7
QHash::iterator::operator++ qhash.h:356
QPollingFileSystemWatcherEngine::timeout 0x00007fdfba8298d6
doActivate<…> 0x00007fdfba8f61f0
QMetaObject::activate 0x00007fdfba8ef587
QTimer::timeout 0x00007fdfba8facfe
QObject::event 0x00007fdfba8ecc55
QApplicationPrivate::notify_helper(QObject *, QEvent *) 0x00007fdfbb565dc3
QApplication::notify(QObject *, QEvent *) 0x00007fdfbb56ebb8
QCoreApplication::notifyInternal2 0x00007fdfba8bf19a
QTimerInfoList::activateTimers 0x00007fdfba918d40
timerSourceDispatch 0x00007fdfba919674
idleTimerSourceDispatch 0x00007fdfba919674
g_main_context_dispatch 0x00007fdfb968317d
<unknown> 0x00007fdfb9683400
g_main_context_iteration 0x00007fdfb96834a3
QEventDispatcherGlib::processEvents 0x00007fdfba9199f2
QEventLoop::exec qflags.h:69
QCoreApplication::exec qflags.h:121
main main.cpp:643
__libc_start_main 0x00007fdfba0f10b3
_start 0x000055daa33442de
```

Waqar144:
That's not linux, is it?

I checked on my system and yes, added notes are not being detected. Putting a `qWarning()` in `notesDirectoryWasModified` I get:

```
[18:41:43] [warning] Note Directory was modified "/home/waqar/Documents/Notes/qowntestings", _isNotesDirectoryWasModifiedDisabled: true
```

that `true` blocks updating

Waqar144:
I wonder why `QPollingFileSystemWatcherEngine` polling file watcher is being used. Can you explicitly do a debug build using `cmake -DCMAKE_BUILD_TYPE=Debug ..`, that should give better backtrace.

PR #2457, opened 2022-02-15
Downgrade to C++14

Waqar144:
gcc 4.8.x doesn't support C++17

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -17,7 +17,7 @@ add_library(md4c
 add_executable(QOwnNotes "")
 
 set_target_properties(QOwnNotes PROPERTIES
-    CXX_STANDARD 17
+    CXX_STANDARD 14
     CXX_STANDARD_REQUIRED ON
     AUTOMOC ON
     AUTOUIC ON
--- a/src/QOwnNotes.pro
+++ b/src/QOwnNotes.pro
@@ -102,13 +102,13 @@ TRANSLATIONS = languages/QOwnNotes_en.ts \
     languages/QOwnNotes_sq.ts
 
 CODECFORTR = UTF-8
-CONFIG += c++17
+CONFIG += c++14
 
 # for older Qt versions
 win32-msvc {
-    QMAKE_CXXFLAGS += "/std:c++17"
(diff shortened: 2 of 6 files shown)


pbek:
There still seems to be a problem on Ubuntu 18.04 for the AppImage build: https://github.com/pbek/QOwnNotes/runs/5195670986?check_suite_focus=true

Waqar144:
> There still seems to be a problem on Ubuntu 18.04 for the AppImage build: https://github.com/pbek/QOwnNotes/runs/5195670986?check_suite_focus=true

pushed a fix, but that only affects the github CI or anyone using DEV_MODE

Waqar144:
I have disabled `-pedantic` but that is the wrong fix. Better would be that you disable `DEV_MODE` for this specific appimage build instead. That flag is important if you want to be sure that we only have C++14 code, and not C++17 and thus prevent failures in places like OBS etc.

pbek:
> Better would be that you disable DEV_MODE for this specific appimage build instead. 

For OBS you could attempt to do that in https://github.com/pbek/QOwnNotes/blob/develop/obs/appimage.yml
I wonder why the GitHub AppImage build went through...

Waqar144:
> 

The build went through because I disabled the flag. OBS builds should succeed now anyways. I meant that we should disable `config+=DEV_MODE` for github appimage build to catch build errors early here rather than find out in OBS.

pbek:
Still a lot of broken builds on OBS: https://build.opensuse.org/package/show/home:pbek:QOwnNotes/desktop

Waqar144:
> Still a lot of broken builds on OBS: https://build.opensuse.org/package/show/home:pbek:QOwnNotes/desktop

Which distro does the AppImage use?

pbek:
Hard to tell from https://build.opensuse.org/package/live_build_log/home:pbek:QOwnNotes/desktop/AppImage/x86_64, the user has no control about that.

Waqar144:
can the compiler be updated there?

pbek:
Not that I know of. You only have control over the appimage.yml.

Waqar144:
we can use `-std=c++1y`, but that flags implies experimental 14 support, which means somethings might not work at some point.

PR #2460, opened 2022-02-15
Use FileWatchDisabler in removeConflictedNotesDatabaseCopies

Waqar144:
This fixes the bug that note created outside QON wasn't detected. The
problem was that if we early returned on

```
if (count == 0) {
   return;
}
```
we left file watch disabled forever.

Using FileWatchDisabler does the same thing, but is less error prone
as it will auto enable file watch on scope exit.


pbek:
Awesome, no crash any more too! 😆

PR #2461, opened 2022-02-15
Try to reenable -pedantic

Waqar144:


.github/workflows/build-test.yml
```diff
@@ -146,12 +146,12 @@ jobs:
     - if: contains( matrix.os, 'ubuntu-latest') && steps.changes.outputs.DO_BUILD
       name: Do Linux tests
       run: |
-        cd ../tests && qmake CONFIG+=debug CONFIG+=DEV_MODE DEFINES+=INTEGRATION_TESTS && make -s -j$(nproc)
+        cd ../tests && qmake CONFIG+=debug CONFIG+=DEV_MODE DEFINES+=INTEGRATION_TESTS && make -s -j 8
```
pbek, line 149: Why 8? And why not `$(nproc)`?

Waqar144: nproc doesn't work on windows so had to fallback.

PR #2467, opened 2022-02-18
Update Note Text when reload by external editor

clevertension:
when i open a note, and this current note is synced with delete and add way, the note text can't change if the code is below
setCurrentNote(std::move(this->currentNote), **false**);

so i change **false** to **true**, and now it is ok

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2944,7 +2944,7 @@ void MainWindow::notesWereModified(const QString &str) {
 
         // rebuild and reload the notes directory list
         buildNotesIndexAndLoadNoteDirectoryList();
-        setCurrentNote(std::move(this->currentNote), false);
+        setCurrentNote(std::move(this->currentNote), true);
     }
 }
 


pbek:
You shouldn't even be in the that else branch if the current note was modified.

> when i open a note, and this current note is synced with delete and add way, the note text can't change if the code is below

I don't understand "current note is synced with delete and add way".
Can you please describe step by step what you did to reproduce this issue?

@Waqar144, can this have anything to do with the latest changes?

src/mainwindow.cpp
```diff
@@ -2944,7 +2944,7 @@ void MainWindow::notesWereModified(const QString &str) {
 
         // rebuild and reload the notes directory list
         buildNotesIndexAndLoadNoteDirectoryList();
-        setCurrentNote(std::move(this->currentNote), false);
+        setCurrentNote(std::move(this->currentNote), true);
```
pbek, line 2947: I don't think that's a good idea... If any random note in the note folder was changed it would trigger that the note text edit is refreshed and that is very bad while you are editing the note.

pbek:
> current note is synced with delete and add way

Do you mean if the current note is removed externally and then added again externally?

@Waqar144, shouldn't that be detected if the current note was removed externally?

Waqar144:
> > current note is synced with delete and add way
> 
> Do you mean if the current note is removed externally and then added again externally?
> 
> @Waqar144, shouldn't that be detected if the current note was removed externally?

I don't think this PR is addressing detection, rather the change is detected by the text isn't being updated.

pbek:
> I don't think this PR is addressing detection, rather the change is detected by the text isn't being updated.

It's about the current note and that code tree (the "else") shouldn't be called if it's about the current note. That tree is for "if some other note changed, not the current note.

Waqar144:
> > I don't think this PR is addressing detection, rather the change is detected by the text isn't being updated.
> 
> It's about the current note and that code tree (the "else") shouldn't be called if it's about the current note. That tree is for "if some other note changed, not the current note.

Yes, but if your current note was renamed? The else case will get hit in that case.

pbek:
I think the current note file was removed and than created anew.

Waqar144:
> I think the current note file was removed and than created anew.

But generally falling in the else case for this is okay if you triggered the rename yourself from within kate. But I am really curious as to what triggers this case. Without understanding that we won't really know if the fix is right or not.

clevertension:
i am use https://mega.nz/ as QON storage, when i modify file, the MEGA will sync the notes from one machine to another, so if the two machine open same note, i find that note modify will be detected, but the latest note text can be reloaded, so i change the false flag to true, and it is ok now

clevertension:
i test the MEGA sync way is remove and add note in my mac OS

Waqar144:
So it's basically that the current note was modified + renamed outside QOwnNotes.

pbek:
> So it's basically that the current note was modified + renamed outside QOwnNotes.

Not removed and created again?

Waqar144:
> Not removed and created again?

That happens every time you save / rename / modify a file. File is deleted and new one with new content is create in its place i.e., there is no "update". So true here as well.

pbek:
Question is, should the current note be reset after the "delete"?

pbek:
I'll close this PR, because it would break note-editing as mentioned in the code-comment. More debate in #2493.

PR #2470, opened 2022-02-19
Fix javascript highlighting (resolves #2468)

### LINKED ISSUE #2468: Javascript in code blocks treated as Java
Issue: Using `javascript` as the language name causes code blocks to be highlighted as Java. Using `js` highlights the code correctly.

#### Expected behaviour

Using either `js` and `javascript` should have the same effect on highlighted code.

#### Actual behaviour

`javascript` causes code to be highlighted as if it was Java, while `js` correctly highlights code as JavaScript.

#### Steps to reproduce

Paste the following code in QOwnNotes:

	```javascript
	function random(min, max) {
	   let num = Math.floor(Math.random() * (max - min)) + min;
	   return num;
	}
	```

	```js
	function random(min, max) {
	   let num = Math.floor(Math.random() * (max - min)) + min;
	   return num;
	}
	```

Screenshot showing how QOwnNotes highlights the two examples above:

image

## Solution

This issue appears to be caused by [this line](https://github.com/pbek/QOwnNotes/blob/develop/src/helpers/codetohtmlconverter.cpp#L35) in `codetohtmlconverter.cpp`:

```cpp
        {QStringLiteral("javascript"), CodeToHtmlConverter::CodeJava},
```

It looks like this line should be changed to the following:

```cpp
        {QStringLiteral("javascript"), CodeToHtmlConverter::CodeJs},
```

If this is correct, I would be happy to submit a PR to fix the issue.

dohliam:
This change fixes the issue reported in #2468 by making `javascript` in code blocks highlight as JavaScript rather than Java.

### DIFF
--- a/src/helpers/codetohtmlconverter.cpp
+++ b/src/helpers/codetohtmlconverter.cpp
@@ -32,7 +32,7 @@ void CodeToHtmlConverter::initCodeLangs() Q_DECL_NOTHROW {
         {QStringLiteral("html"), CodeToHtmlConverter::CodeXML},
         {QStringLiteral("ini"), CodeToHtmlConverter::CodeINI},
         {QStringLiteral("java"), CodeToHtmlConverter::CodeJava},
-        {QStringLiteral("javascript"), CodeToHtmlConverter::CodeJava},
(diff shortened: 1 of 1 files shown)


pbek:
Thank you!

Issue #2076, opened 2021-03-21
[F] Issue/Feature Request: Remote git repositories


teotikalki:
#### Expected behaviour

Hi. I'm a happy QOwnNotes user for several years now. I use it with my own local Nextcloud installation for sync.

I have used the 'Use git to keep local history of your notes' feature since the beginning. I have noticed that syncing the .git folder is MOST of the work that my Nextcloud is actualy doing! The git folder is also (by FAR) most of my Notes foldeer itself (see below). I have long thought that the best solution to this is to allow for a remote git repository.

What I would like (Expected behaviour):

+ Each client can use QON and Nextcloud
+ Nextcloud will sync the notes folder
	+ Nextcloud will IGNORE the git folder (really all .git folders)
	+ This sync will be very fast as THOUSANDS of tiny objects will no longer be handled
+ Each client will push to a remote git repository (in my case, my local kallithea instance)

From the UI, the interface could simply be updated to:

'Use Git to keep a history of your notes?' -> Y
'Keep local git repo' or 'Enter url for remote repository'

#### Actual behaviour

```
teoti@avalon:~/Nextcloud$ du -ms Notes/
866	Notes/
```
```
teoti@avalon:~/Nextcloud$ du -ms Notes/.git
855	Notes/.git
```

I frequently have 'Conflicted copy' sync errors which I have gotten used to ignoring/dealing with by deleting the conflicted copy. I am currently having a serious issue with corruption in the git repo, which I started looking into because one or two folders have selectively stopped syncing.

#### Steps to reproduce

1. Use a git repository to backup your notes AND sync your notes folder. 
2. Wait

#### Relevant log output in the Log panel

<details><summary>Expand</summary>

```

```
</details>

#### Information about the application, settings and environment

<details><summary>Expand</summary>

QOwnNotes Debug Information
===========================

## General Info

**Current Date**: `Sun. Mar. 21 11:33:41 2021`
**Version**: `20.9.2`
**Build date**: `Sep  3 2020`
**Build number**: `757`
**Platform**: `linux`
**Operating System**: `Debian GNU/Linux bullseye/sid`
**Build architecture**: `x86_64`
**Current architecture**: `x86_64`
**Release**: `OBS`
**Qt Version (build)**: `5.11.3`
**Qt Version (runtime)**: `5.12.5`
**Portable mode**: `no`
**Settings path / key**: `/home/teoti/.config/PBE/QOwnNotes.conf`
[...]

Issue #744, opened 2017-10-12
Push to git after a number of commits


sanderboom:
#### Feature request: 

Push to git after a number of commits. If anything goes wrong, the process stops until the application is restarted. The user gets a message and needs to solve everything manually.

I looked through the hooks, but there is no post GIT commit hook, right? 

Is pushing something you are inclined to add to the main application, or do you rather see this in a script? In case of the latter, what hook do you suggest to use?

pbek:
I never did a `git push` because an external repository needs to be in place and conflict management can not be done automatically. But you could use a timer, like in <https://github.com/pbek/QOwnNotes/blob/develop/doc/scripting/example.qml> to do that push every x seconds.

Does that answer your question?

sanderboom:
Every x seconds is not optimal, but doable I think. I'll experiment with creating a script for it. Thanks.

pbek:
If you come up with a great solution please consider doing a pull request to <https://github.com/qownnotes/scripts>!

pbek:
@sanderboom, maybe you want to try https://github.com/muesli/gitomatic/

PR #2512, opened 2022-04-12
WIP: Add support for git push on every commit

grisuthedragon:
Since I am using my own git server instead of nextcloud to keep my files in sync, I often forgot to push the changes to my notes' folder. For this reason, I created this patch, which allows performing a `git push` after each commit. In this way, my remote git repository is kept up to date when using QOwnNotes. 

The patch adds an option to the notes folder in order to enable/disable the git push functionallity. This requires the database version of the notesFolder table to be updated to `42`.

### DIFF
--- a/src/dialogs/settingsdialog.ui
+++ b/src/dialogs/settingsdialog.ui
@@ -4766,7 +4766,7 @@ Just test yourself if you get sync conflicts and set a higher value if so.</stri
                 <string notr="true">&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
 &lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
 p, li { white-space: pre-wrap; }
-&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'Ubuntu'; font-size:11pt; font-weight:400; font-style:normal;&quot;&gt;
+&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'Noto Sans'; font-size:11pt; font-weight:400; font-style:normal;&quot;&gt;
 &lt;p style=&quot; margin-top:12px; margin-bottom:12px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;span style=&quot; font-family:'Sans Serif'; font-size:9pt;&quot;&gt;portable mode information will be loaded here&lt;/span&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
                </property>
                <property name="openExternalLinks">
--- a/src/dialogs/settingsdialog.ui
+++ b/src/dialogs/settingsdialog.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>1127</width>
-    <height>1218</height>
+    <width>800</width>
+    <height>600</height>
    </rect>
   </property>
   <property name="sizePolicy">
@@ -3558,8 +3558,8 @@
                           <rect>
                            <x>0</x>
                            <y>0</y>
-                           <width>338</width>
-                           <height>467</height>
+                           <width>360</width>
+                           <height>532</height>
                           </rect>
                          </property>
                          <layout class="QGridLayout" name="gridLayout_55">
--- a/src/dialogs/settingsdialog.ui
+++ b/src/dialogs/settingsdialog.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>800</width>
-    <height>600</height>
+    <width>1127</width>
+    <height>1218</height>
    </rect>
   </property>
   <property name="sizePolicy">
@@ -409,7 +409,7 @@
          <item>
           <widget class="QStackedWidget" name="settingsStackedWidget">
            <property name="currentIndex">
-            <number>19</number>
+            <number>0</number>
            </property>
            <widget class="QWidget" name="noteFoldersPage">
             <layout class="QGridLayout" name="gridLayout_19">
@@ -557,7 +557,7 @@
                    <enum>QFrame::Raised</enum>
                   </property>
                   <layout class="QGridLayout" name="gridLayout_22">
-                   <item row="12" column="0" colspan="2">
+                   <item row="13" column="0" colspan="2">
                     <widget class="QFrame" name="noteFolderRemotePathTreeWidgetFrame">
                      <property name="frameShape">
                       <enum>QFrame::NoFrame</enum>
@@ -648,24 +648,24 @@
                      </layout>
                     </widget>
                    </item>
-                   <item row="8" column="0" colspan="2">
+                   <item row="9" column="0" colspan="2">
                     <widget class="QLabel" name="noteFolderCloudConnectionLabel">
                      <property name="text">
                       <string>Cloud connection:</string>
                      </property>
                     </widget>
                    </item>
-                   <item row="9" column="0" colspan="2">
+                   <item row="10" column="0" colspan="2">
                     <widget class="QComboBox" name="noteFolderCloudConnectionComboBox"/>
                    </item>
-                   <item row="10" column="0" colspan="2">
+                   <item row="11" column="0" colspan="2">
                     <widget class="QLabel" name="noteFolderRemotePathLabel">
                      <property name="text">
                       <string>Path on your ownCloud server:</string>
                      </property>
                     </widget>
                    </item>
-                   <item row="13" column="0">
+                   <item row="14" column="0">
                     <widget class="QFrame" name="noteFolderVerticalSpacerFrame">
                      <property name="frameShape">
                       <enum>QFrame::NoFrame</enum>
@@ -722,6 +722,16 @@
                      </property>
                     </widget>
                    </item>
+                   <item row="7" column="0" colspan="2">
+                    <widget class="QCheckBox" name="noteFolderGitPushCheckBox">
+                     <property name="toolTip">
+                      <string>You need to have an upstream configured in your git setup!</string>
+                     </property>
+                     <property name="text">
+                      <string>Perform a git push after each commit</string>
+                     </property>
+                    </widget>
+                   </item>
                    <item row="2" column="0" colspan="2">
                     <widget class="QCheckBox" name="noteFolderActiveCheckBox">
                      <property name="text">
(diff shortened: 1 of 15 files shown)


pbek:
As much I would like this to include, I think it still doesn't address the fundamental issue mentioned in https://github.com/pbek/QOwnNotes/issues/2076#issuecomment-803638320 or https://github.com/pbek/QOwnNotes/issues/744#issuecomment-336156850... The part with conflict resolution. I can already see issues like "QOwnNotes isn't pushing my commit, pls fix". 🙈 

Do you have a suggestion?

src/dialogs/settingsdialog.ui
```diff
@@ -4756,7 +4766,7 @@ Just test yourself if you get sync conflicts and set a higher value if so.</stri
                 <string notr="true">&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
 &lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
 p, li { white-space: pre-wrap; }
-&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'Noto Sans'; font-size:11pt; font-weight:400; font-style:normal;&quot;&gt;
```
pbek, line 4759: Can you please not change the default font. 😁  I know, QtCreator tends to do this...

grisuthedragon: Sorry I tried to remove everything, which was not part of adding the checkbox, but it seems that I overlooked something.

pbek: One of the quirks of QtCreator...

pbek:
Another issue, especially for new users of QOwnNotes and git, will be how to set a git remote...

grisuthedragon:
For the beginning, I would say, they have to do this on the CLI. If the repository is clone, the remote is set any way.

pbek:
> So the option I could think of would be to build a pull/push around the commit and make some options like "On conflict use mine" "On conflict use theirs" to obtain a default handling of the conflict situation. 

That would work... Another option then.

> For the beginning, I would say, they have to do this on the CLI. If the repository is clone, the remote is set any way.

Then we need a guide to explain the steps...
For example a `git.md` in https://github.com/pbek/QOwnNotes/tree/develop/webpage/src/getting-started...

grisuthedragon:
I found another problem, which is causes trouble in case of the pulls. The detection if a file was changed on disk does not detect the stuff going on when pulling and thus changes are lost. 

So another idea, which may lead to an easier to handle solution. (Regarding where I was coming from, that I work on the notes, close the laptop and then recognized that the stuff was not pushed). 
 1. Add a Menu Entry or a Toolbar Button for "Git Pull", which raises an error message in case of a merge conflict. 
 2. Do the push on every commit, raising an error message, that a pull is required before. 
 
Does this sound reasonable?

pbek:
Seems like there is a reason that tools like https://github.com/muesli/gitomatic/ exist. 😅 
I don't think anyone would like a manual pull.... Maybe just force-push (and possibly destroying anything on the remote repository)?

grisuthedragon:
I tried a bit how to get a reasonable git workflow integrated, which tries to merge if possible, otherwise throws an error message. In a normal one-person workflow this seems to be reasonable. The only problem I have left now, is that after doing the pull/push, all currently open notes have to be reloaded in case the have been changed by the merge. @pbek  Can you help me implementing this?

pbek:
> all currently open notes have to be reloaded in case the have been changed by the merge.

Doesn't that already work?
The note folder should reload automatically if files are changed if you do not actively block that FileWatcher and DirectoryWatcher event in the source code.

src/utils/git.cpp
```diff
@@ -55,7 +55,7 @@ void Utils::Git::commitCurrentNoteFolder() {
     if (!executeGitCommand(git, QStringList{"init"}, process) ||
         !executeGitCommand(git, QStringList{"config", "commit.gpgsign", "false"}, process) ||
         !executeGitCommand(git, QStringList{"add", "-A"}, process) ||
-        !executeGitCommand(git, QStringList{"commit", "-m", "QOwnNotes commit"}, process)) {
+        !executeGitCommand(git, QStringList{"commit", "-m", "QOwnNotes commit"}, process, false, true)) {
```
pbek, line 58: Showing an error dialog here will cause the user to be spammed with error messages every X seconds. And if the user isn't in front of a computer there maybe will be 1000s of dialogs 😅

src/utils/git.cpp
```diff
@@ -79,7 +79,25 @@ void Utils::Git::pushCurrentNoteFolder() {
     auto* process = new QProcess();
     process->setWorkingDirectory(NoteFolder::currentLocalPath());
 
+    if (!executeGitCommand(git, QStringList{"pull", "--no-commit", "--no-ff"}, process)) {
```
pbek, line 82: What if the error is "git command not found"?

src/utils/git.cpp
```diff
@@ -79,7 +79,25 @@ void Utils::Git::pushCurrentNoteFolder() {
...
+                QObject::tr("The pull and merge of the remote repository failed. Please check on the command line whether the remote is configured correctly"
+                    "or if a manual merge is required."),"git-pull-merge-failed.");
+        delete(process);
+        return;
+    } else {
+        executeGitCommand(git, QStringList{"commit", "-am", "Merge commit"}, process, false, true);
+    }
+
     if (!executeGitCommand(git, QStringList{"push", "-q"}, process)) {
+        Utils::Gui::warning(
```
pbek, line 97: Again, this will spam the user with error dialogs...

src/utils/git.cpp
grisuthedragon: That one does not show up with error messages, the second boolean parameter means, that the return value is ignored.

src/utils/git.cpp
grisuthedragon: Moving this to the status bar?

grisuthedragon:
> > all currently open notes have to be reloaded in case the have been changed by the merge.
> 
> Doesn't that already work? The note folder should reload automatically if files are changed if you do not actively block that FileWatcher and DirectoryWatcher event in the source code.

Unfortunately not. I tried it in the following way. I changed a file at the beginning in the git remote and in the end using Qownnotes. Thus, the merge will be successful. In order to see the change, I have to close the note and open it again.

src/utils/git.cpp
grisuthedragon: Since the commit is done before, there would be an error message earlier.

src/utils/git.cpp
pbek: It will be hard to do that from a library (`git.cpp`).

pbek:
> Unfortunately not. I tried it in the following way. I changed a file at the beginning in the git remote and in the end using Qownnotes. Thus, the merge will be successful. In order to see the change, I have to close the note and open it again.

...or reload the note folder (see *Note* menu)
Or in code like in `MainWindow::on_action_Reload_note_folder_triggered()`, but maybe not reload the current note, that might be annoying while you are typing into the note...

grisuthedragon:
> > > all currently open notes have to be reloaded in case the have been changed by the merge.
> > 
> > 
> > Doesn't that already work? The note folder should reload automatically if files are changed if you do not actively block that FileWatcher and DirectoryWatcher event in the source code.
> 
> Unfortunately not. I tried it in the following way. I changed a file at the beginning in the git remote and in the end using Qownnotes. Thus, the merge will be successful. In order to see the change, I have to close the note and open it again.

It seems that I was too fast during testing. Now it works.

grisuthedragon:
I moved the error and status message to the status bar in the main window. Now it seems to work, at least for my purpose.

src/utils/git.cpp
grisuthedragon: The errors are now shown in the status bar. So there will be no flood of dialogs.

grisuthedragon: > ~Showing an error dialog here will cause the user to be spammed with error messages every X seconds. And if the user isn't in front of a computer there maybe will be 1000s of dialogs~
> 
> Ah, new parameter

Yes, to distinguish between "allowed/accepted" errors and the ones which are fatal.

src/utils/git.cpp
grisuthedragon: I pass a pointer from the mainwindow in, when the commit or pull is called.

src/utils/git.cpp
```diff
@@ -23,6 +23,7 @@
 
 #include "gui.h"
 #include "misc.h"
+#include "mainwindow.h"
```
pbek, line 26: MainWindow in a utility library... 🙈

pbek:
When I activate the "git push checkbox" (with the default settings of 10sec interval) QOwnNotes takes about 19sec to start (instead of 0.1sec). When I start typing (or even if I don't do anything at all) the UI is frozen for a long time, then responds for a sec or so just to be unresponsive again...

I had this issue when an external git push wasn't correctly setup (remote git repository on gitlab.com) and also after it was setup and no error was shown in QOwnNotes. Most of the time I had to kill the QOwnNotes process. I'm not even able to interact with the application after the startup anymore.

Everything still works fine when I don't use the "put push" feature.

PS: It should not be possible to enable the "git push" checkbox when "git support" isn't enabled in the settings dialog.

pbek:
> the UI is frozen for a long time,

Because `process->waitForFinished()` is running the process in the main thread to wait for the result.

pbek:
...a QTimer::singleShot around the whole remote git pull/push operation (and being very diligent what gets accessed in that time) should help.

and because the remote git pull/push can take ages there also needs to be some kind of locking so that not multiple git pull/push operations can be active at the same time (and cause havoc on git and DOS the local computer and the server)

src/utils/git.cpp
```diff
@@ -50,12 +56,59 @@ void Utils::Git::commitCurrentNoteFolder() {
     if (!executeGitCommand(git, QStringList{"init"}, process) ||
         !executeGitCommand(git, QStringList{"config", "commit.gpgsign", "false"}, process) ||
         !executeGitCommand(git, QStringList{"add", "-A"}, process) ||
-        !executeGitCommand(git, QStringList{"commit", "-m", "QOwnNotes commit"}, process)) {
+        !executeGitCommand(git, QStringList{"commit", "-m", "QOwnNotes commit"}, process, false, true)) {
+        if ( mw != nullptr ) {
```
pbek, line 60: Can you please use the same coding style for brackets and spaces like in the rest of the project. 😉

pbek:
> a QTimer::singleShot around the whole remote git pull/push operation (and being very diligent what gets accessed in that time) should help.

`QTimer::singleShot` still blocks the UI thread. 😬  seems like QThread needs to be used to run the commit and push methods.

pbek:
I will close this until there is more information.

Issue #2528, opened 2022-05-15
Segfault when running script with shortcut


louwers:
<!--
If you are able to open the application the best way gather all needed
information and post questions, feature requests or issues is to use
the issue assistant in the "Help menu"!

The menu entry is named "Post questions, feature requests or issues".

If you have an issue with a script please open an issue on
https://github.com/qownnotes/scripts/issues and mention the authors of the script.
You will find the authors in the *Script repository*.
-->

I have [this script](https://gist.github.com/louwers/41ece664d27b72ff870d4cfa8c471915) bound to Alt+D, but when I run it with Alt+D, the application crashes (segfault).

When I run it from the menu it always runs fine.

#### Expected behaviour

Script runs the same as when run from the menu.

#### Actual behaviour

Script runs without segfault.

#### Steps to reproduce

I am on Linux. Import the script above. Bind it to Alt + D. Not immediately, but after a restart it will likely segfault. Re-binding the script makes it work again for a bit, until it will start segfaulting again.

#### Output from the debug section in the settings dialog

Here is a stack trace:

```                                                                                                                                                                                                            1   QObject::objectName() const                                                                                                                                                                                   0x7ffff64a25d4 
2   operator()                                                                                                                                                                             mainwindow.cpp     658 0x555555acb88e 
3   QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, MainWindow::initGlobalKeyboardShortcuts()::<lambda()>>::call(struct {...} &, void * *)                       qobjectdefs_impl.h 146 0x555555b1aeda 
4   QtPrivate::Functor<MainWindow::initGlobalKeyboardShortcuts()::<lambda()>, 0>::call<QtPrivate::List<>, void>(struct {...} &, void *, void * *)                                          qobjectdefs_impl.h 256 0x555555b1a7f4 
5   QtPrivate::QFunctorSlotObject<MainWindow::initGlobalKeyboardShortcuts()::<lambda()>, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase *, QObject *, void * *, bool *) qobjectdefs_impl.h 443 0x555555b19903 
6   ??                                                                                                                                                                                                            0x7ffff64b1463 
7   QHotkey::activated                                                                                                                                                                     moc_qhotkey.cpp    278 0x555555bba2fa 
[...]

PR #2529, opened 2022-05-15
Prevent segfault when findAction returns nullptr

louwers:
"Solves" #2528, but the question is: why can't it find the custom action even though I see it in my menu?


pbek:
Ah, sorry. I fixed it already... 😅

PR #2601, opened 2022-09-02
Avoid reading settings twice

Waqar144:


src/mainwindow.cpp
```diff
@@ -5124,7 +5124,6 @@ void MainWindow::updateCurrentFolderTooltip() {
  */
 void MainWindow::openSettingsDialog(int page, bool openScriptRepository) {
     QPointer<SettingsDialog> settingsDialog = new SettingsDialog(page, this);
-    settingsDialog->readSettings();
```
pbek, line 5127: Hm, may cause regressions...

pbek: Ah, never mind, that was redundant since a new instance of the dialog is used every time.

Waqar144: yup

PR #2605, opened 2022-09-11
where for a default time format

GarryFlemings:
This PR affects only file webpage/src/editor/time-formats.md and adds only one paragraph.

The added paragraph tells readers where they should place the result of using this page.

Maybe there's a need to remove the preceding sentence.

If that preceding sentence refers to a syntax useable at other than the menu location my sentence points to, maybe that sentence is fine as it is.


pbek:
Thank you 😉

PR #2635, opened 2022-10-27
Use MainWindow::instance everywhere

Waqar144:
Also includes a fix for context menu positioning in a separate commit

Context menu positioning for preview and edit seem to be broken on wayland for some reason. Not sure why. Code looks correct


pbek:
Thank you very much!

pbek:
Ah, too early. It breaks the windows builds https://github.com/pbek/QOwnNotes/actions/runs/3336646627/jobs/5522049242

pbek:
I fixed it

Waqar144:
Thanks

PR #2641, opened 2022-10-31
Update md4c

Waqar144:


pbek:
Can I pull in https://github.com/qownnotes/md4c/tree/work/update-md4c already?

Waqar144:
Yes, but it won't build without this patch.

So I would suggest something like Locally bump and apply this patch on top and push

pbek:
Tests are failing, need some project adapting too for sure.
https://github.com/pbek/QOwnNotes/actions/runs/3363372327/jobs/5576417624#step:14:24

PR #2659, opened 2022-11-14
Fix highlighting with unterminated string

Waqar144:


pbek:
Awesome, thank you very much!
But seems to break snap building: <https://github.com/pbek/QOwnNotes/actions/runs/3460256746/jobs/5776603664#step:19:4539>

pbek:
Thank you very much!

Issue #2670, opened 2022-11-22
Indented code block preview error


pbek:
@Waqar144, as follow up from #2667.

~~~
    ```ini
    test1=test2
    ```

everything that comes **afterward** is broken
~~~

![image](https://user-images.githubusercontent.com/1798101/203304617-2017d339-2ab2-417a-bd71-b2d1ab765f8c.png)

Waqar144:
Probably broken since forever. Solution for now is to Just write simpler markdown. Above snippet doesn't make a lot of sense

Waqar144:
if you want the tildes in the preview, sure

pbek:
Yes, to show a code block with markdown code...

![image](https://user-images.githubusercontent.com/1798101/203328962-914f1898-7fec-4774-b914-306e256b908b.png)

pbek:
But doing this with tilde characters would (almost) work:

    ~~~
    ```ini
    test1=test2
    ```
    ~~~

![image](https://user-images.githubusercontent.com/1798101/203329798-3620d0bd-c5b3-48bf-a185-1b113b19478d.png)

There we have another bug. 😅

pbek:
I put that part into a new issue #2671.

pbek:
#2673 helped a bit, but if there is another code block before it the preview is still broken:

![image](https://user-images.githubusercontent.com/1798101/203504383-9a72eb23-fee0-45a9-802b-59dcbb8a08e8.png)

Waqar144:
try
```diff
diff --git a/src/entities/note.cpp b/src/entities/note.cpp
index 620ed9acf..304554dfa 100644
--- a/src/entities/note.cpp
+++ b/src/entities/note.cpp
@@ -2365,8 +2365,8 @@ static void highlightCode(QString &str, const QString &type, int cbCount) {
                 break;
             }
 
-            if (firstBlock >= 4) {
-                bool fourSpaces = std::all_of(str.cbegin() + (firstBlock - 4), str.cbegin() + firstBlock, [](QChar c){
+            if (currentCbPos >= 4) {
+                bool fourSpaces = std::all_of(str.cbegin() + (currentCbPos - 4), str.cbegin() + currentCbPos, [](QChar c){
                     return c.isSpace();
                 });
                 if (fourSpaces) {
```

pbek:
That works! Thank you! 🥳

PR #2673, opened 2022-11-22
Fix fences prefixed with 4 spaces

Waqar144:


pbek:
Helped a bit, but still broken when there is another code block before it: https://github.com/pbek/QOwnNotes/issues/2670#issuecomment-1324719543

PR #2675, opened 2022-11-24
Don't clone history

tim-gromeyer:


README.md
```diff
@@ -157,7 +157,7 @@ cd src
 lrelease QOwnNotes.pro
 
 qmake
-make
+make -j4
```
pbek, line 160: That depends on the amount of cores the system has...

tim-gromeyer: Yeah, but [here](https://github.com/pbek/QOwnNotes#building-qownnotes-for-development) it's the same

pbek: Ok, then :)

pbek:
Thank you

PR #2703, opened 2023-01-14
Update debian installation instructions

mikkoi:
New APT security model allows public keys to be connected only to specific
repositories. This eliminates the problem that any key could be used
in conjunction with any repository, not just the one which distributed
its key.

"[..] The key MUST NOT be placed in /etc/apt/trusted.gpg.d or loaded by apt-key add.
If future updates to the key will be managed by an apt/dpkg package as recommended below,
then it SHOULD be downloaded into /usr/share/keyrings using the same filename
that will be provided by the package. If it will be managed locally ,
it SHOULD be downloaded into /etc/apt/keyrings instead."

[OpenPGP Key Distribution](https://wiki.debian.org/DebianRepository/UseThirdParty#OpenPGP_Key_distribution)

https://manpages.debian.org/bullseye/apt/sources.list.5.en.html
https://wiki.debian.org/DebianRepository/UseThirdParty


pbek:
Thank you!

PR #2712, opened 2023-01-26
Run Clang format script over the codebase

Waqar144:
Column length is increased to 100 as 80 is too short imo

Also add a git-blame-ignore-revs file which ignores the formatting revisions while git-blaming


pbek:
Sound good. Thank you.
Care to share what commands you used?
And did you use `scripts/clang-format-project.sh`?

Waqar144:
> Sound good. Thank you. Care to share what commands you used? And did you use `scripts/clang-format-project.sh`?

Yes, used just the script

PR #2713, opened 2023-01-26
Qt6: Fix hunspell

Waqar144:
Enable core5compat to be able to use QTextCodec. QStringConverter doesn't support many encodings that we need to support all dicts

### DIFF
--- a/.github/workflows/build-release-qt6.yml
+++ b/.github/workflows/build-release-qt6.yml
@@ -24,7 +24,7 @@ env:
   PUBLISHER: "Patrizio Bekerle"
   REPO_DIR: "/home/runner/work/QOwnNotes"
   QT_VERSION: 6.5.0
-  QT_MODULES: "qtwebsockets"
+  QT_MODULES: "qtwebsockets qt5compat"
   QMAKE: qmake6
   CORES: 16
 
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -120,7 +120,7 @@ jobs:
         aqtversion: '==2.0.0'
         version: ${{ matrix.qt-version }}
         arch: win64_mingw81
-        modules: qtwebsockets
+        modules: qtwebsockets qt5compat
         cache: ${{ steps.cache-qt.outputs.cache-hit }}
     - if: false == contains( matrix.os, 'windows') && ( matrix.qt-version != '6.2.0' )
       name: Install Qt < 6 on Linux/macOS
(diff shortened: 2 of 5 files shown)


pbek:
Do we need to add core5compat? What will be the migration path away from it?

Waqar144:
> Do we need to add core5compat?

Yes. As noted in the description.

> What will be the migration path away from it?

No idea, maybe some third party library. Maybe Qt will add support for more encodings. Will have to see when core5compat is no longer supported.

Waqar144:
We can also pull in QTextCodec if really needed. The class seems to be fairly standalone

pbek:
> We can also pull in QTextCodec if really needed. The class seems to be fairly standalone

Tempting, I'd vote for that.

Waqar144:
Right, but its not as simple as just copying the files and pasting here. It will likely need a lot of adjustment before it is usable.

pbek:
Do you mean adjustment in the QTextCodec or in QOwnNotes? Adjustments of what kind?

Waqar144:
In QTextCodec of course. It is spread around in 10+ files

pbek:
Thank you. Good question what to do... I think it still would be great not to depend on core5compat, what do you think?

Waqar144:
It would be, yes, but unfortunately the new classes don't support all the encodings

pbek:
You mean the ones from core5compat?

Waqar144:
No. The ones in core5compat are not new classes, they are old ones. New ones are the ones being removed here.

pbek:
I'm a bit slow today, please help me out... 🤪 
So are we talking about QStringConverter (that doesn't support many encodings) now?
Or whether to fork QTextCodec or use the one from core5compat?

I was talking about "forking QTextCodec".

Waqar144:
> So are we talking about QStringConverter (that doesn't support many encodings) now?

Yes

Forking QTextCodec is possible but it is a lot of work I think. So I would suggest using core5compat. If Qt drops it, we can think about forking or maybe someone else will do it before us and we can use that.

pbek:
core5compat it is. Can you please add the module to the build processes?

Waqar144:
Added the module, not sure why its not picking it up

pbek:
You want to also add it in https://github.com/pbek/QOwnNotes/blob/develop/.github/workflows/build-test.yml

Waqar144:
Ok, seems to be working now. Thanks a lot

pbek:
Thank you!

Waqar144:
I think we can drop core5compat, qt6 does support those codecs. But the fix will need to be different from what we already had before this pr so reverting won't be needed.

pbek:
That would be awesome! 😁 
Btw. I renamed the branch `develop` to `main` to be more "conform"...

PR #2787, opened 2023-05-29
nix: improve nix-shell

wineee:
In general though, nativeBuildInputs is useful for cross-compilation as commands from those derivations will be available on the buildPlatform and execute at build time. Whereas buildInputs will likely be the architecture of the hostPlatform, so the derivation can link against those inputs (and be used at run-time).

https://discourse.nixos.org/t/use-buildinputs-or-nativebuildinputs-for-nix-shell/8464


pbek:
That looks nice. Thank you, @wineee!

pbek:
And thank you for the explanation!

PR #2813, opened 2023-08-16
i18n: use qt6 dir for translations when building on qt6

charliewolf:
hi, i'm working on updating a qownnotes rpm spec to support qt6 and ran into some issues with this hardcoded path.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -49,8 +49,13 @@ void loadTranslations(QTranslator *translator, const QString &locale) {
     loadTranslation(translator[5], appPath + "/languages/QOwnNotes_" + locale);
     loadTranslation(translator[6], appPath + "/QOwnNotes_" + locale);
     loadTranslation(translator[7], "../src/languages/QOwnNotes_" + locale);
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+    loadTranslation(translator[8], "../share/qt6/translations/QOwnNotes_" + locale);
+    loadTranslation(translator[9], appPath + "/../share/qt6/translations/QOwnNotes_" + locale);
+#else
     loadTranslation(translator[8], "../share/qt5/translations/QOwnNotes_" + locale);
     loadTranslation(translator[9], appPath + "/../share/qt5/translations/QOwnNotes_" + locale);
+#endif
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -49,8 +49,13 @@ void loadTranslations(QTranslator *translator, const QString &locale) {
     loadTranslation(translator[5], appPath + "/languages/QOwnNotes_" + locale);
     loadTranslation(translator[6], appPath + "/QOwnNotes_" + locale);
     loadTranslation(translator[7], "../src/languages/QOwnNotes_" + locale);
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
```
pbek, line 52: You'll need a `>=` check here, wouldn't you?

charliewolf: indeed, fixed.

pbek:
Will need changes too:

https://github.com/pbek/QOwnNotes/blob/68c860e109c0e9f99f5541aa48ebec4cc34e3052/src/main.cpp#L62

src/main.cpp
pbek: thank you

charliewolf:
got it, added a check in `loadReleaseTranslations` as well.

pbek:
I tested building on NixOS with Qt 6 and for some reason the `qt6/translations` wasn't even created 🤔 .

charliewolf:
i'm not familiar with nix but this patch worked for me on RHEL 9. did it create the qt5 dir or what did it do?

pbek:
I got it working now. LGTM. Thank you.

pbek:
Hm, seems like a lot of Qt6 builds break...
https://github.com/pbek/QOwnNotes/actions/runs/5887832027

barolo:
> Hm, seems like a lot of Qt6 builds break... https://github.com/pbek/QOwnNotes/actions/runs/5887832027

Hi is support for Qt6  complete now?

pbek:
> Hi is support for Qt6 complete now?

Yes, quite a while ago...

barolo:
> > Hi is support for Qt6 complete now?
> 
> Yes, quite a while ago...

Oh. Would you have some time to update ebuilds then?

PR #2850, opened 2023-09-16
Resolve QA notice from bug #2821

FireUnderTheMountain:
Update xdg icon cache during install/uninstall to silence notification during install.


FireUnderTheMountain:
One moment.

FireUnderTheMountain:
Ok, cleared for acceptance :)

Experienced mystery 'command not found' in the middle of 12 other successful compiles, but can't reproduce.

pbek:
Thank you 😁

Issue #2847, opened 2023-09-15
[BUG] qownnotes-23.9.4 build on gentoo fails due to lack of QT concurrent module


erfanara:
Hi, 

installing app-office/qownnotes-23.9.4 fails:

`emerge --ask app-office/qownnotes` leads to:
```
 * Running qmake ...
Info: creating stash file /var/tmp/portage/app-office/qownnotes-23.9.4/work/qownnotes-23.9.4/.qmake.stash
Project MESSAGE: Currently in RELEASE mode.
Project ERROR: Unknown module(s) in QT: concurrent                                                                                                                    [ !! ]
```

problem can be fixed by installing `dev-qt/qtconcurrent:5`

I would be happy to help you in fixing and testing ebuilds, should i open a PR?

pbek:
Is this maybe related to #2821.
@catthou, @werdnab, can you also reproduce this issue?

catthou:
I've got concurrency installed for something else and 23.9.4 installed and runs good. I don't dare to uninstall it because QT is very fickle on gentoo!

FireUnderTheMountain:
The suggested package, `dev-qt/qtconcurrent:5`, is probably required... @erfanara, were you able to install any prior version of QOwnNotes? Just curious, because I don't see a recent commit that would add this as a dependency.

@pbek - Is the ebuild is supposed to be maintained for all release versions, or out-of-the-box should it only "support" the most recent stable release? I don't really want to add a new global dependency if it's only needed for specific versions. Usually gentoo ebuilds have their version number in the ebuild name so you can isolate differences between versions, example: `qownnotes-23.9.4.ebuild`. On the other hand, if the ebuild is meant to only install the most recent release, then it could be added as a dependency without concern.

Note for Qt6, there is no `dev-qt/qtconcurrent` package, it looks like a module is built using the `concurrent` useflag in `dev-qt/qtbase` - though I'm not entirely familiar with Qt to be exactly sure.

erfanara:
Well actually i tried to install QOwnNotes on a minimal fresh gentoo and i got that error, 
i just tried out some older versions like 23.9.1 and the issue persists.

`src/QOwnNotes.pro` shows QT dependency modules needed to compile the project i guess:
```
QT       += core gui widgets sql svg network xml printsupport qml websockets concurrent
```
Looking at the git blame shows that this line is 3 years old 😅,
So i think the concurrent module can be installed as a dependency even on old ebuilds.

[...]

PR #2857, opened 2023-09-22
#2847: Require qtconcurrent for Qt5

FireUnderTheMountain:
The need for qtconcurrent predates all versions available in the gentoo overlay, so you don't need to push a new release version just for this: I'll go in and submit a PR to that repo to fix those.


pbek:
Thank you!

PR #2866, opened 2023-09-29
Update donate.md for Privacy Coin Options

freQniK:


pbek:
What is the reason for this change?

freQniK:
Purely personal. I use QOwnNotes on multiple computers (I think 5?) and I wanted to donate but prefer to donate with privacy coins, since real crypto is hidden.

pbek:
I don't really have a wallet for those, and Coinbase doesn't support them neither... I never really got into Monero yet and I don't even know Pirate Chain.

freQniK:
Well, my suggestion would be to keep this pull request open until you educate yourself on those, understand why they are important, and find the different exchanges where you can trade them.  

In the meantime, I will donate some satoshis to your Bitcoin address for thanks for creating a wonderful app that I use every single day on all three major platforms.

freQniK:
Also, you should always self-custody your crypto and not have them sent to exchanges unless you plan on cashing out. That's pretty much the whole point of crypto in case you were not aware.

pbek:
The thing is there hardly are any donations in any form of money 😁  And the last donation even in BTC was the first one many years ago. But thanks a lot for your BTC donation now!

Keeping an offline wallet is also a lot of responsibility (and work), but I know very well the pitfalls of an online one!

freQniK:
It's sad really. I'm a FOSS dev as well and the amount of donations is so pathetic. I will say that people in the Monero community, which I am part of on the various channels, really do like donating XMR to FOSS projects. 

Honestly, I've donated over $400 in the last year because of the simplicity and anonymity with crypto. I post on twitter everytime I make a donation to hopefully get people to follow suit.

pbek:
I'll close this for now until I get back to it.

PR #2878, opened 2023-10-12
fakevim: fix build with qt 6.6.0

NickCao:
Reference: https://github.com/qt-creator/qt-creator/commit/e56e3b6f374e00179eb0537198437864dddc47f2


pbek:
They needed to move the method? 😁

NickCao:
Just to move it below the definition of `class FakeVimHandler::Private`

pbek:
Thank you!

Issue #2520, opened 2022-04-25
Note List Tree Expansion Persistency


offset-torque:
Firstly thanks for this gem of application. I have tried many applications like Joplin and Obsidian but always came back to QOwnNotes for its features.

**Context**
I am using `Note List` as `Note Tree`.
I have a modest number of folders containing notes and some of them are always in an expanded state because I continously update notes in that folders.

**Current behaviour**
Whenever I delete a file from a folder or add a file to a folder, `Note Tree` collapses.
I am not sure if this is a bug (probably, because some folders are unaffected) or intended behavior (I hope not) but this should never happen. After adding a note, I have to click around 5-10 folders to expand them again. Then when I add another note, same thing again. Delete a note, more clicks etc.

I don't have a big knowledge base and even now it is tedious to re-expand all folders again and again. With more complex folder structures, this behavior would definitely render the application unusable. Expanding 50-100 folders each time you add/delete a note is very bad user experience and a deal breaker for this type of application.

![qownnotes](https://user-images.githubusercontent.com/84420582/165051447-cea7d56e-58de-4510-8010-9e8992a251f5.png)

**Expected behaviour**
Note tree folders should always keep their collapsed/expanded state whatever happens (for example as in **Obsidian**).

**Alternative solution**
If keeping track of expanded state of folders is **impossible/too much work** then QOwnNotes should at least give the user an option like **`Keep tree notes always expanded`** so all the levels are visible all the time. Although not optimal, this is better than the Sisyphus re-clicking alternative.

pbek:
I guess you are talking about #790, which is still "work in progress" (like the setting in the settings dialog says) 😁 
I guess it's best to deposit your wishes there...

But since there is no persistence of folder information, because all note sub-folders are just folders in the file-system the collapsed-state of the tree items will most likely also never be persisted... 😅 

> Firstly thanks for this gem of application.

Thank you 😉

offset-torque:
I found this exactly same issue in https://github.com/pbek/QOwnNotes/issues/790 as CLOSED:
[Prevent note list panel from collapsing after file name change](https://github.com/pbek/QOwnNotes/issues/2181)

> But since there is no persistence of folder information, because all note sub-folders are just folders in the file-system the collapsed-state of the tree items will most likely also never be persisted...

I understand implementing this is a big effort if there is no persistence structure currently. 

In this case it seems my alternative suggestion is the best workaround: 

[...]

Issue #2204, opened 2021-08-13
[I] moving a note closes folder in folder view


knodalyte:
#### Expected behaviour

If I select a note in the folder view and use the context menu to move the note to another folder, the original folder in the folder view should remail open.

This happens on other folder/note context operations, like Remove note.

#### Actual behaviour

folder is closed after moving a note to another folder

#### Steps to reproduce

open a subfolder in the folder tree
select a note and right click to open context menu
select move to folder and select a folder

#### Relevant log output in the Log panel

<details><summary>Expand</summary>

```

```
</details>

#### Information about the application, settings and environment

<details><summary>Expand</summary>

QOwnNotes Debug Information
===========================

## General Info

**Current Date**: `Fri Aug 13 12:03:29 2021`
**Version**: `21.8.3`
**Build date**: `Aug 12 2021`
**Build number**: `867`
**Platform**: `linux`
**Operating System**: `SparkyLinux 6 (Po-Tolo)`
**Build architecture**: `x86_64`
**Current architecture**: `x86_64`
**Release**: `OBS`
**Qt Version (build)**: `5.11.3`
**Qt Version (runtime)**: `5.15.2`
**Portable mode**: `no`
**Settings path / key**: `/home/cba/.config/PBE/QOwnNotes.conf`
**Application database path**: `/home/cba/.local/share/PBE/QOwnNotes/QOwnNotes.sqlite`
**Application arguments**: `/usr/bin/QOwnNotes`
**Qt Debug**: `0`
**Locale (system)**: `en_US`
**Locale (interface)**: *empty*
**Primary screen resolution**: `1920x1080`
**Screen resolution(s)**: `1920x1080, 1920x1080`
**Icon theme**: `breeze-dark-qownnotes`
**Notes in current note folder**: `191`
**Calendar items**: `0`
**Enabled scripts**: `7`

## Server Info

**serverUrl**: `https://cloud01.opsdata.ch`
**appIsValid**: `no`
**notesPathExists**: *empty*
**connectionErrorMessage**: *empty*

## Spellchecking

**Enabled**: `true`
**Selected language**: `en_US`
**Language codes**: `en_US`
**Language names**: `American English (United States)`
**Application dictionaries path**: `/home/cba/.local/share/PBE/QOwnNotes/dicts`

## Note folders

**currentNoteFolderId**: `13`

### Note folder `pkb`

**id**: `2`
**isCurrent**: `no`
**activeTagId**: `-1`
[...]

Issue #2181, opened 2021-07-06
[F] Prevent note list panel from collapsing after file name change


peterschussheim:
#### Expected behaviour

Ideally, this wouldn't show to the user.  I'd be happy to fix this provided you are OK with the changes and also with a bit of guidance from yourself on the source files relevant.

#### Actual behaviour

The left note list panel collapses when I update the file name of an open document (note that there is a timeout before QOwnNotes saves the file).

![1tD258GyQz](https://user-images.githubusercontent.com/3654683/124655153-77582900-de6d-11eb-9858-d0c49c905fc8.gif)

#### Steps to reproduce

1) enable "Note List" under Window->Panels
2) expand a folder
3) create new note
4) change default note title

#### Relevant log output in the Log panel

<details><summary>Expand</summary>

```

```
</details>

#### Information about the application, settings and environment

<details><summary>Expand</summary>

QOwnNotes Debug Information
===========================

## General Info

**Current Date**: `Tue Jul 6 15:17:20 2021`
**Version**: `21.7.3`
**Build date**: `Jul  5 2021`
**Build number**: `852`
**Platform**: `windows`
**Operating System**: `Windows 10 Version 2009`
**Build architecture**: `x86_64`
**Current architecture**: `x86_64`
**Release**: `GitHub Actions`
**Qt Version (build)**: `5.15.2`
**Qt Version (runtime)**: `5.15.2`
**Portable mode**: `no`
**Settings path / key**: `\HKEY_CURRENT_USER\Software\PBE\QOwnNotes`
**Application database path**: `C:\Users\peter\AppData\Roaming\PBE\QOwnNotes\QOwnNotes.sqlite`
**Application arguments**: `C:\Users\peter\QOwnNotes\QOwnNotes.exe`, `--after-update`, `C:\Users\peter\AppData\Local\Temp\QOwnNotes-KKHDDe-updater.bat`
**Qt Debug**: `0`
**Locale (system)**: `en_US`
**Locale (interface)**: *empty*
**Primary screen resolution**: `1920x1080`
**Icon theme**: `breeze-qownnotes`
**Notes in current note folder**: `79`
**Calendar items**: `0`
**Enabled scripts**: `5`

## Server Info

**serverUrl**: *empty*
**appIsValid**: `yes`
**notesPathExists**: *empty*
**serverVersion**: *empty*
**appVersion**: *empty*

## Spellchecking

**Enabled**: `true`
**Selected language**: `en_US`
[...]

PR #2928, opened 2023-12-20
bugfix: save collapse-state for note tree

### LINKED ISSUE #2924: [BUG] Notes list view resets when changing note name
Hi! This app is really good, I just noticed one problem that occurs ofter: when I change the name of a note (by editing the first line), and then save it, the note list panel resets. What I mean is that it scrolls to the top, and categories collapse/uncollapse randomly. This makes it hard to keep track of where I am. I'm using the experimental feature.

#### Output from the debug section in the settings dialog

<details><summary>Expand</summary>

</details>

#### Relevant log output in the Log panel

<details><summary>Expand</summary>

</details>

discapes:
Part of the feature #790 
Fixes #2924, #2520, #2204 and #2181
Woah, four in one!

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2123,6 +2123,8 @@ void MainWindow::loadNoteDirectoryList() {
 
         ui->noteSubFolderTreeWidget->buildTreeForParentItem(noteFolderItem);
         noteFolderItem->setExpanded(true);
+        connect(ui->noteTreeWidget, &QTreeWidget::itemExpanded, ui->noteSubFolderTreeWidget, &NoteSubFolderTree::onItemExpanded);
+        connect(ui->noteTreeWidget, &QTreeWidget::itemCollapsed, ui->noteSubFolderTreeWidget, &NoteSubFolderTree::onItemExpanded);
 
         itemCount = Note::countAll();
     } else {
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -506,6 +506,8 @@ void MainWindow::initTreeWidgets() {
             &MainWindow::onMultipleSubfoldersSelected);
     connect(ui->noteSubFolderTreeWidget, &NoteSubFolderTree::currentSubFolderChanged, this,
             &MainWindow::onCurrentSubFolderChanged);
+    connect(ui->noteTreeWidget, &QTreeWidget::itemExpanded, ui->noteSubFolderTreeWidget, &NoteSubFolderTree::onItemExpanded);
+    connect(ui->noteTreeWidget, &QTreeWidget::itemCollapsed, ui->noteSubFolderTreeWidget, &NoteSubFolderTree::onItemExpanded);
 }
 
 void MainWindow::initNotePreviewAndTextEdits() {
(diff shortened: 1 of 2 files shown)


src/widgets/notesubfoldertree.h
```diff
@@ -21,8 +21,9 @@ class NoteSubFolderTree : public QTreeWidget {
 
     static QMenu *contextMenu(QTreeWidget *parent);
 
-   private Q_SLOTS:
```
pbek, line 24: What's the reason for this change?

pbek: I mean it not being a slot anymore...

discapes: I just needed to make it public, and I wasn't sure if it's also required to be a slot, or how to do it.

src/mainwindow.cpp
```diff
@@ -2123,6 +2123,8 @@ void MainWindow::loadNoteDirectoryList() {
 
         ui->noteSubFolderTreeWidget->buildTreeForParentItem(noteFolderItem);
         noteFolderItem->setExpanded(true);
+        connect(ui->noteTreeWidget, &QTreeWidget::itemExpanded, ui->noteSubFolderTreeWidget, &NoteSubFolderTree::onItemExpanded);
```
pbek, line 2126: Since `loadNoteDirectoryList()` is called quite often, aren't those `connect` statements executed multiple times too?

discapes: Yeah they should probably be moved someplace better. But I'm wondering why in the firstplace is the note list used for the tree, and not the directory view. Then no extra connect calls would be needed.

pbek: It doesn't even make sense to me that the NoteSubFolderTree is involved at all, the notes are also added to the NoteSubFolderTree now. I suspect that might be a big regression from a pull request from two years ago that moved lots of stuff around.

pbek: > the notes are also added to the NoteSubFolderTree now

Ok, the note sub-folder tree is disabled when I restart the application. Still strange.

pbek: > But I'm wondering why in the firstplace is the note list used for the tree, and not the directory view.

Because the directory view lacks far more interaction for making notes work there than the notetree widget.

src/widgets/notesubfoldertree.h
discapes: Yeah it has to be public in order to be accessed in the connect() call.

discapes: Okay, I made them slots in the latest commit.

src/mainwindow.cpp
discapes: > Because the directory view lacks far more interaction for making notes work there than the notetree widget.
Got it. I guess NoteSubFolderTree is involved because the `buildTreeForParentItem` method was found useful. It confused me at first since it's getting called on an item that's not a part of the NoteSubFolderTree, and I'm surprised it works. But that's fine. Anyway, I moved the connect() calls to a better place in the last commit.

src/mainwindow.cpp
```diff
@@ -506,6 +506,8 @@ void MainWindow::initTreeWidgets() {
             &MainWindow::onMultipleSubfoldersSelected);
     connect(ui->noteSubFolderTreeWidget, &NoteSubFolderTree::currentSubFolderChanged, this,
             &MainWindow::onCurrentSubFolderChanged);
+    connect(ui->noteTreeWidget, &QTreeWidget::itemExpanded, ui->noteSubFolderTreeWidget, &NoteSubFolderTree::onItemExpanded);
```
pbek, line 509: This only should be done in the experimental note tree mode!

src/mainwindow.cpp
pbek: I think a lot got mixed up in the "cleanup contribution" two years ago...

src/mainwindow.cpp
discapes: I'm not sure the notelist can be expanded if not in experimental note tree mode?

pbek: No, the note list is a list by default.

pbek: > This only should be done in the experimental note tree mode!

Do you want to add the check? There is a method for it.

discapes: Done !

pbek: Thank you!

Issue #2998, opened 2024-04-02
[FEATURE] Progress bar and/or tags for Todo's


Akselmo:
#### Is your feature request related to a problem? Please describe.
I like to use tags and progress in todos, so I would like to have those exposed
in the todo editor.

#### Describe the solution you'd like
Basically a slider for todo progress, input for comma separated tags, and change
the priority to spinbox or something i think.

#### Describe alternatives you've considered
I can change these from nextcloud or calendar apps directly but it seems those values
are overwritten when modified from qownnotes.

#### Additional context
I would not mind helping out with this and trying to create the feature(s) myself! :)

Akselmo:
My problem with the deck for now is that I can't move items between boards from Qownnotes (unless I've missed something?)

But yeah I totally understand. I can give it a look myself at some point and see how much pain it is lol.


> Thank you for working on KDE. 😉 I'm a bit of a KDE fanboy myself. 😉

And thank you for using KDE stuff! :)

Akselmo:
Hello, I think I got it to work, at least for me it seems to work nicely :) https://github.com/pbek/QOwnNotes/pull/3006

No tag support, just progress, which is good enough for my usecase.

Akselmo:
Meanwhile I started the progress bar thing, i will look also into the tag support.

Considering we just read the tags from the incoming file, save them somewhere, expose a input field that allows editing them and then send them modified onwards, it should be relatively simple (famous last words), especially since we don't necessarily need to save them to the database at all.

Seems tags are just comma separated list of items `CATEGORIES:tag1,tag2` 

Honestly if progress bar is too much effort to get work well enough, tag system would also work for me, as I can then use tags to indicate the progress of the item (TODO, WAITING, CANCELED, DOING...). For me that would probably be even more useful than progress bar.

Of course would be cool to have both, but again if progress is too difficult and messy, tags will suffice for my usecase.

I'll start hacking on it and see how much effort it will be. :)

Akselmo:
Alright, I did not make pull request yet but here's the tags support commit https://github.com/Akselmo/QOwnNotes/commit/73aabe291dfa6f40fa002389f513f296cc312ee4

Very similar and I had no issues, especially compared to progress bar.

I wonder if the input field should be something different? Something like the note tag system is, or drop down with input field or something.. 

![image](https://github.com/pbek/QOwnNotes/assets/13801236/3d989136-8d73-408d-85b7-807f1cb15b29)

[...]

PR #3006, opened 2024-04-14
Add progress bar support to to-do items

Akselmo:
As per discussion in this issue https://github.com/pbek/QOwnNotes/issues/2998

I managed to get it to work, tested it out and so far no regressions! :)

This would help my workflow a lot, so I hope you like it as well.

If I add tag support, I'll do it in another PR, however I find progress bar way more useful.

I was also thinking of adding it as a sortable item to the list of tasks, but this now just adds the basic functionality.

image

## Changes

- Maximizing the slider turns the completion checkbox on when saved.
- On hover, there's tooltip for the progress.
- I added tickmarks to the priority slider so it's easier to distinguish between the two
- Clicking on checkbox sets the progress to 100 or 0
- Depending on progress, the completion date is saved or removed

## Todos
- [ ] If subtask or top-level task with subtasks progress is set 100, the upload fails. Clicking on complete checkbox works however.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ ui_*.h
 /src/QOwnNotes
 secrets/*
 
+.cache/clangd/index
--- a/.gitignore
+++ b/.gitignore
@@ -27,4 +27,3 @@ ui_*.h
 /src/QOwnNotes
 secrets/*
 
-.cache/clangd/index
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.2.2)
 
 project(QOwnNotes)
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 
 add_subdirectory(src)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.2.2)
 
 project(QOwnNotes)
-
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 
 add_subdirectory(src)
--- a/src/dialogs/tododialog.ui
+++ b/src/dialogs/tododialog.ui
@@ -105,16 +105,19 @@
             <property name="orientation">
              <enum>Qt::Horizontal</enum>
             </property>
+            <property name="tickPosition">
+             <enum>QSlider::TicksBelow</enum>
+            </property>
            </widget>
           </item>
-          <item row="1" column="0">
+          <item row="2" column="0">
            <widget class="QLabel" name="reminderLabel">
             <property name="text">
              <string>Reminder</string>
             </property>
            </widget>
           </item>
-          <item row="1" column="1">
+          <item row="2" column="1">
            <widget class="QFrame" name="reminderFrame">
             <layout class="QHBoxLayout" name="horizontalLayout_3">
              <property name="leftMargin">
@@ -144,6 +147,20 @@
               </widget>
              </item>
             </layout>
+           </widget>
+          </item>
+          <item row="1" column="0">
+           <widget class="QLabel" name="progressLabel">
+            <property name="text">
+             <string>Progress</string>
+            </property>
+           </widget>
+          </item>
+          <item row="1" column="1">
+           <widget class="QSlider" name="progressSlider">
+            <property name="orientation">
+             <enum>Qt::Horizontal</enum>
+            </property>
            </widget>
           </item>
          </layout>
--- a/src/dialogs/tododialog.ui
+++ b/src/dialogs/tododialog.ui
@@ -161,6 +161,15 @@
             <property name="orientation">
              <enum>Qt::Horizontal</enum>
             </property>
+            <property name="minimum">
+             <number>0</number>
+            </property>
+            <property name="maximum">
+             <number>100</number>
+            </property>
+            <property name="pageStep">
+             <number>1</number>
+            </property>
            </widget>
           </item>
          </layout>
--- a/src/entities/calendaritem.cpp
+++ b/src/entities/calendaritem.cpp
@@ -251,6 +251,7 @@ bool CalendarItem::fillFromQuery(const QSqlQuery &query) {
     this->description = query.value(QStringLiteral("description")).toString();
     this->hasDirtyData = query.value(QStringLiteral("has_dirty_data")).toInt() == 1;
     this->completed = query.value(QStringLiteral("completed")).toInt() == 1;
+    this->progress = query.value(QStringLiteral("progress")).toInt();
     this->priority = query.value(QStringLiteral("priority")).toInt();
     this->uid = query.value(QStringLiteral("uid")).toString();
     this->relatedUid = query.value(QStringLiteral("related_uid")).toString();
@@ -484,6 +485,7 @@ bool CalendarItem::store() {
             "description = :description, "
             "has_dirty_data = :has_dirty_data, "
             "completed = :completed, "
+            "progress = :progress, "
             "calendar = :calendar, uid = :uid, related_uid = :related_uid, "
             "ics_data = :ics_data, etag = :etag, "
             "last_modified_string = :last_modified_string, "
@@ -496,12 +498,12 @@ bool CalendarItem::store() {
         query.prepare(
             "INSERT INTO calendarItem "
             "(summary, url, description, calendar, uid, related_uid, ics_data, "
-            "etag, last_modified_string, has_dirty_data, completed, "
+            "etag, last_modified_string, has_dirty_data, completed, progress, "
             "alarm_date, priority, created, modified, "
             "completed_date, sort_priority) "
             "VALUES (:summary, :url, :description, :calendar, :uid,"
             " :related_uid, :ics_data, :etag, :last_modified_string, "
-            ":has_dirty_data, :completed, :alarm_date, :priority, "
+            ":has_dirty_data, :completed, progress, :alarm_date, :priority, "
             ":created, :modified, :completed_date, "
             ":sort_priority)");
     }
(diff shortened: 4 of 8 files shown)


Akselmo:
Added screenshot and more detail

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.2.2)
 
 project(QOwnNotes)
-
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
pbek, line 4: Why was this added?

.gitignore
```diff
@@ -27,3 +27,4 @@ ui_*.h
 /src/QOwnNotes
 secrets/*
 
+.cache/clangd/index
```
pbek, line 30: Is this change needed?

CMakeLists.txt
Akselmo: Oops I forgot I added that lol, I'll remove it

src/dialogs/tododialog.ui
```diff
@@ -105,16 +105,19 @@
             <property name="orientation">
              <enum>Qt::Horizontal</enum>
             </property>
+            <property name="tickPosition">
```
pbek, line 108: What are those new 3 lines for?

.gitignore
Akselmo: Nope, another mishap by me, will remove

CMakeLists.txt
Akselmo: Removed it now

.gitignore
Akselmo: Fixed

src/dialogs/tododialog.ui
Akselmo: This adds the tickmarks to the priority slider so it is easier to distinguish between priority and progress slider by glance. Check the screenshot in PR summary for what I mean :)

src/entities/calendaritem.cpp
```diff
@@ -486,12 +496,12 @@ bool CalendarItem::store() {
...
             "INSERT INTO calendarItem "
             "(summary, url, description, calendar, uid, related_uid, ics_data, "
-            "etag, last_modified_string, has_dirty_data, completed, "
+            "etag, last_modified_string, has_dirty_data, completed, progress, "
             "alarm_date, priority, created, modified, "
             "completed_date, sort_priority) "
             "VALUES (:summary, :url, :description, :calendar, :uid,"
             " :related_uid, :ics_data, :etag, :last_modified_string, "
-            ":has_dirty_data, :completed, :alarm_date, :priority, "
+            ":has_dirty_data, :completed, progress, :alarm_date, :priority, "
```
pbek, line 504: This looks like a `:` is missing. Are you sure this query ever worked? 😉

Akselmo: Oops.. I did not get any warnings though :joy: Added it there, will test again now.

pbek:
Did you test your changes with Nextcloud Tasks and mobile applications (like Tasks and jtx Board on Android)? Because this is where most likely regressions will happen.

src/entities/calendaritem.cpp
Akselmo: I don't see any noticeable difference but at least it's there

Akselmo:
> Did you test your changes with Nextcloud Tasks and mobile applications (like Tasks and jtx Board on Android)? Because this is where most likely regressions will happen.

Good points!

I tested with Nextcloud Tasks. It followed the task status closely to what I set to Qownnotes and vice versa.

On my phone I have Tasks.org and aCalendar+, I will also test it there!

src/services/databaseservice.cpp
```diff
@@ -876,6 +876,11 @@ bool DatabaseService::setupTables() {
         version = 41;
     }
 
+    if (version < 42) {
+        queryDisk.exec(QStringLiteral("ALTER TABLE calendarItem ADD progress INTEGER DEFAULT 0"));
```
pbek, line 880: What happens to already completed tasks with this change? You are setting the progress to "0" for those too!

pbek:
> > Did you test your changes with Nextcloud Tasks and mobile applications (like Tasks and jtx Board on Android)? Because this is where most likely regressions will happen.
> 
> Good points!
> 
> I tested with Nextcloud Tasks. It followed the task status closely to what I set to Qownnotes and vice versa.
> 
> On my phone I have Tasks.org and aCalendar+, I will also test it there!

Every app has their different ways to interpret or write the iCal "file". 🙈 I had to build in several fail-saves until it worked "everywhere".

src/services/databaseservice.cpp
Akselmo: For me at least, it did not change the progress to 0 for them. The progress data is already there (completed seemed to be just checking if progress is 0 or 100). 

I have old database without any of the changes and when I use it, it updates accordingly, the completed tasks stay completed and their progress slider is just turned to 100.

Akselmo: Tested it again, the priority slider is just set to max on anything that is set completed and it stays completed as it should.

Akselmo:
> Every app has their different ways to interpret or write the iCal "file". 🙈 I had to build in several fail-saves until it worked "everywhere".

Yeah seems Tasks.org does not have progress, nor aCalendar+ (which uses Tasks.org to handle the tasks anyway so we can concentrate on Tasks.org only.)

However, it does keep the old progress value around when I set the completion status off. Both nextcloud tasks and qownnotes show it as it should.

src/services/databaseservice.cpp
pbek: If the user doesn't open the todo dialog after a migration, the progress state of 0 could be synced to the CalDAV server. It would be more save when the migration had two more queries...

1) set the progress to 0 if completed is 0
2) set the progress to 100 if completed is 1

pbek:
In different apps...
How does setting progress on setting a parent task influences a child task?
And the other way around, how does setting progress on setting a child task influences the parent task?

I wonder if they are getting confused if the states of the parent doesn't match the child and vice versa... 🤔

Akselmo:
> In different apps... How does setting progress on setting a parent task influences a child task? And the other way around, how does setting progress on setting a child task influences the parent task?
> 
> I wonder if they are getting confused if the states of the parent doesn't match the child and vice versa... 🤔

Seems that child tasks don't show properly in either end for me. However I have had the same issue with the current version of the client, not the one with progress bar.. Seems something breaks with them?

For example creating todo with subtask on qownnotes doesnt show on nextcloud tasks, but if i do the same on nextcloud tasks, i see both tasks on qownnotes, but they're not connected to each other and shown as separate tasks.

src/services/databaseservice.cpp
Akselmo: Good idea. I'll add that... So something like this?

```cpp
    if (version < 42) {
        queryDisk.exec(QStringLiteral("ALTER TABLE calendarItem ADD progress INTEGER DEFAULT 0"));
        // loop over all items in calendarItem table
        // if completed is 0, set progress 0
        // if completed is 1, set progress 100
        version = 42;
    }

```

I am not very familiar with databases so how can I do this? :sweat_smile:

pbek:
> For example creating todo with subtask on qownnotes doesnt show on nextcloud tasks, but if i do the same on nextcloud tasks, i see both tasks on qownnotes, but they're not connected to each other and shown as separate tasks.

I just created a new task with sub-task in QOwnNotes (current release). It shows up for me in Nextcloud Tasks. 🤔

image

src/services/databaseservice.cpp
pbek: It's much easier: 😉

```sql
UPDATE calendarItem SET progress = 0 WHERE completed = 0;
UPDATE calendarItem SET progress = 100 WHERE completed = 1;
```

Akselmo:
> I just created a new task with sub-task in QOwnNotes (current release). It shows up for me in Nextcloud Tasks. 🤔

Okay I think something is broken with my caldav then, I have a task that references itself as subtask (i have no idea how that has happened???) so it just confuses all my clients. Fun.

pbek:
> Okay I think something is broken with my caldav then, I have a task that references itself as subtask (i have no idea how that has happened???) so it just confuses all my clients. Fun.

Phew, I wonder how that happened! 😬

src/services/databaseservice.cpp
Akselmo: Thanks, I added those there. :)

Akselmo:
> > Okay I think something is broken with my caldav then, I have a task that references itself as subtask (i have no idea how that has happened???) so it just confuses all my clients. Fun.
> 
> Phew, I wonder how that happened! 😬

Lol no idea.. Anyhow, I fixed it by downloading the calendar, removing the offending item and uploading it again... 

But yeah now I can create tasks just fine again. The progresses seem to be independent per task, although if I set the top level task as 100 in nextcloud, the childtasks get completed too. Seems something breaks there when trying to set progress to 100 in the task, maybe it's not marked completed properly. Will have to investigate that part.

pbek:
> Seems something breaks there when trying to set progress to 100 in the task, maybe it's not marked completed properly. Will have to investigate that part.

It's very important that it will also be set as completed!

src/entities/calendaritem.cpp
```diff
@@ -77,15 +79,21 @@ void CalendarItem::setAlarmDate(const QDateTime &dateTime) { this->alarmDate = d
 
 void CalendarItem::setPriority(int value) { this->priority = value; }
 
-void CalendarItem::setCompleted(bool value) { this->completed = value; }
+void CalendarItem::setProgress(int value) { this->progress = value; }
```
pbek, line 82: You need to use `setCompleted(value = 100)` too here! But don't use `setProgress` inside `setCompleted` then, but set the value directly. 😉

pbek: Hm, but you don't want to override progress too in that case inside setCompleted... you need to check that somewhere else...

Akselmo: Yeah, I wasn't sure which way to go here. :sweat_smile:

pbek: Either way, progress and completed need to be "linked" in some way.

pbek:
I tried the PR, two things came up:

1) I think it would be good to see and be able to edit the actual process as a number (like in Nextcloud tasks).
2) When I move the progress slider of a task to 100% and pressed *Save* I got an `QNetworkReply error 299` in the log, the task vanished from the list (I needed to *Clear cache and reload* to get it again) and the progress was not stored.
The Nextcloud log said: `Sabre\DAV\Exception\UnsupportedMediaType: Validation error in iCalendar: The supplied value () is not a correct DATE-TIME`

pbek:
ad 2) I wasn't even able to save any other change in any other task. Without this PR saving the same tasks work.

Akselmo:
Yeah I may give up on the progress slider part for now.. :/ I only really need the tag system personally, so I am concentrating on that now.

Sorry for the extra work I have caused with this. I may pick up on this PR later. (Or invite anyone else to do so!)

pbek:
Then I'll close this for now.

PR #3007, opened 2024-04-14
Tags support for todos

Akselmo:
As per discussion in https://github.com/pbek/QOwnNotes/issues/2998#issuecomment-2054165209

This adds a very simple input field to the todo items list, with comma delimited tags support.

322305792-3d989136-8d73-408d-85b7-807f1cb15b29

Tags can be any text and they're separated with `,`. If user leaves a `,` in the end of the string, it will be cleaned from the input when saving.

I am personally fine with just comma limited input field for this :)

### DIFF
--- a/src/dialogs/tododialog.cpp
+++ b/src/dialogs/tododialog.cpp
@@ -917,6 +917,10 @@ void TodoDialog::reloadCurrentTags()
     ui->tagCloudLayout->layout()->setSizeConstraint(QLayout::SetFixedSize);
     for (auto tag : _todoTagsList)
     {
+        if (tag.isEmpty())
+        {
+            continue;
+        }
         QPushButton *tagButton = new QPushButton(tag, ui->tagCloudLayout->layout()->widget());
         tagButton->setIcon(QIcon::fromTheme("tag-delete"));
         connect(tagButton, &QPushButton::released, this, [=](){
@@ -931,13 +935,19 @@ void TodoDialog::reloadCurrentTags()
 
 QString TodoDialog::getTagString()
 {
+    // Remove any possible empty items
+    _todoTagsList.removeAll(QString(""));
     return _todoTagsList.join(',');
 }
 
 void TodoDialog::on_tagsLineEdit_returnPressed()
 {
-    _todoTagsList.append(ui->tagsLineEdit->text());
+    auto newTag = ui->tagsLineEdit->text().simplified();
+    if (_todoTagsList.contains(newTag))
+    {
+        return;
+    }
+    _todoTagsList.append(newTag);
     ui->tagsLineEdit->clear();
-    updateCurrentCalendarItemWithFormData();
     reloadCurrentTags();
 }
--- a/src/dialogs/tododialog.cpp
+++ b/src/dialogs/tododialog.cpp
@@ -829,7 +829,13 @@ void TodoDialog::on_todoItemTreeWidget_currentItemChanged(QTreeWidgetItem *curre
 
         // Absolute monster of a regexp that checks for possible edge cases like \\,\\ and such
         QRegularExpression nonescapedCommas(R"(((?<!\\),(?!,))|((?<=\\\\),(?=\\\\))|((?<=\\\\),(?=\\)))");
-        _todoTagsList = currentCalendarItem.getTags().split(nonescapedCommas, Qt::SkipEmptyParts);
+
+#if (QT_VERSION < QT_VERSION_CHECK(5, 15, 0))
+        _todoTagsList = currentCalendarItem.getTags().split(QStringLiteral("."), QString::SkipEmptyParts);
+#else
+        _todoTagsList = currentCalendarItem.getTags().split(QStringLiteral("."), Qt::SkipEmptyParts);
+#endif
+
         reloadCurrentTags();
 
         QDateTime alarmDate = currentCalendarItem.getAlarmDate();
--- a/src/dialogs/tododialog.cpp
+++ b/src/dialogs/tododialog.cpp
@@ -901,7 +901,7 @@ void TodoDialog::on_showDueTodayItemsOnlyCheckBox_clicked() {
 
 void TodoDialog::cleanTagButtons(){
     QLayoutItem *child;
-    while ((child = ui->tagCloudLayout->layout()->takeAt( 0 )) != nullptr) {
+    while ((child = ui->tagCloudLayout->takeAt( 0 )) != nullptr) {
         delete child->widget();
         delete child;
     }
@@ -910,26 +910,35 @@ void TodoDialog::cleanTagButtons(){
 void TodoDialog::reloadCurrentTags()
 {
     cleanTagButtons();
+    ui->tagsFrame->setVisible(false);
     if (_todoTagsList.empty())
     {
         return;
     }
-    ui->tagCloudLayout->layout()->setSizeConstraint(QLayout::SetFixedSize);
+    ui->tagCloudLayout->setSizeConstraint(QLayout::SetFixedSize);
+    int tagRow = 0;
+    int tagCol = 0;
     for (auto tag : _todoTagsList)
     {
         if (tag.isEmpty())
         {
             continue;
         }
-        QPushButton *tagButton = new QPushButton(tag, ui->tagCloudLayout->layout()->widget());
+        QPushButton *tagButton = new QPushButton(tag, ui->tagCloudLayout->widget());
         tagButton->setIcon(QIcon::fromTheme("tag-delete"));
         connect(tagButton, &QPushButton::released, this, [=](){
             _todoTagsList.removeOne(tag);
             reloadCurrentTags();
         });
-        ui->tagCloudLayout->layout()->addWidget(tagButton);
+        if (tagCol > 3)
+        {
+            tagCol = 0;
+            tagRow++;
+        }
+        ui->tagCloudLayout->addWidget(tagButton, tagRow, tagCol);
+        tagCol++;
     }
-
+    ui->tagsFrame->setVisible(true);
 }
 
 
--- a/src/dialogs/tododialog.cpp
+++ b/src/dialogs/tododialog.cpp
@@ -463,8 +463,6 @@ void TodoDialog::updateCurrentCalendarItemWithFormData() {
     currentCalendarItem.setPriority(priority);
     currentCalendarItem.setSummary(ui->summaryEdit->text());
     currentCalendarItem.setDescription(ui->descriptionEdit->toPlainText());
-    // Clean the last comma
-    ui->tagsLineEdit->setText(ui->tagsLineEdit->text().remove(QRegularExpression(", *$")));
     currentCalendarItem.setTags(getTagString());
     currentCalendarItem.setModified(QDateTime::currentDateTime());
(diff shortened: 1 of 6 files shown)


pbek:
What happens if a tag akready contains a comma?

pbek:
In jtx board tags are just named categories, like in caldav itself. Are they named tags somewhere else?

Akselmo:
> What happens if a tag akready contains a comma?

Seems that they have `\,` in the actual ics file, but on Qownnotes it writes it as a normal comma. Needs to be fixed. I think this can be fixed UX wise by using the same kind of tagging method the notes are using: Spawn a button per tag. Or we keep the input field and render commas in tags as `\,`..

> In jtx board tags are just named categories, like in caldav itself. Are they named tags somewhere else?

They're named as tags in Nextcloud and Tasks.org apps, and KDE Merkuro Calendar.

pbek:
Ok, let's stay with tags.
But the tag assignment will need to be more "sophisticated" than just CS-lines... Maybe something like I did for note tagging.

Akselmo:
> But the tag assignment will need to be more "sophisticated" than just a CS-line. You can even assign tags with a comma in it in Nextcloud Tasks. Maybe something like I did for note tagging.

I agree and I will look into it this evening. I'll likely create a string list from all the items, then spawn a button per item in the string list. Then there is one input field for adding a tag, and clicking on tag will remove it. 

Best part is that I can likely re-use some of your code for that. It's also good idea to do it this way since the user might already know how tagging works in notes so they automatically know how to do it in todo items. :)

pbek:
Being able to see already used tags in a calendar surely would be nice too. 😊 I don't know how far you want to go with your feature. 😉

Akselmo:
> Being able to see already used tags in a calendar surely would be nice too. 😊 I don't know how far you want to go with your feature. 😉

Considering I already use my fork and the tag system with it, I want to have the editing there as well :D I'll see what I can do!

pbek:
That's the spirit. 😉

Akselmo:
Buttons!

image


I'll be using this in my own use for now and see if there's anything that needs improvement

Akselmo:
Only thing so far is that I need to make the buttons go on next row if they dont fit their area. Otherwise it works pretty well for my use! :)

src/dialogs/tododialog.cpp
```diff
@@ -448,6 +455,9 @@ void TodoDialog::updateCurrentCalendarItemWithFormData() {
     currentCalendarItem.setPriority(priority);
     currentCalendarItem.setSummary(ui->summaryEdit->text());
     currentCalendarItem.setDescription(ui->descriptionEdit->toPlainText());
+    // Clean the last comma
+    ui->tagsLineEdit->setText(ui->tagsLineEdit->text().remove(QRegularExpression(", *$")));
```
pbek, line 459: Is this still needed (or even wanted)?

src/dialogs/tododialog.ui
```diff
@@ -146,211 +344,40 @@
...
-         </widget>
-        </item>
-        <item row="4" column="0">
-         <widget class="QCheckBox" name="showDueTodayItemsOnlyCheckBox">
-          <property name="toolTip">
-           <string>This doesn't work for sub-items, because they may be hidden by the parent item!</string>
-          </property>
-          <property name="text">
-           <string>Show only items due today</string>
-          </property>
```
pbek, line 353: Hm, why did this large block move? 🤔

src/services/databaseservice.cpp
```diff
@@ -876,6 +876,11 @@ bool DatabaseService::setupTables() {
         version = 41;
     }
 
+    if (version < 42) {
+        queryDisk.exec(QStringLiteral("ALTER TABLE calendarItem ADD tags VARCHAR(255)"));
```
pbek, line 880: I hope no one has more tags than fitting in 255 char 😉

src/services/databaseservice.cpp
```diff
@@ -876,6 +876,11 @@ bool DatabaseService::setupTables() {
         version = 41;
     }
 
+    if (version < 42) {
```
pbek, line 879: This will clash with https://github.com/pbek/QOwnNotes/pull/3006/files#diff-dc0007d20d0985e45c46ab027eb6de6be1c5c8fbc38faf31229eaa428c4bed89R879, but is hard to fix before one of the 2 PRs are ready to be merged.

src/services/databaseservice.cpp
Akselmo: Heh, yeah, I think I'll set that higher value.

src/dialogs/tododialog.cpp
Akselmo: Not needed, removed it

src/services/databaseservice.cpp
Akselmo: Resolving this since we closed the other PR

src/services/databaseservice.cpp
Akselmo: I set it to 9999, but if there is better setting for this let me know! :)

pbek: SQLite doesn't seem to enforce that limit anyway... I'd just use 512 or so...

Akselmo: Ah, okay, I'll set it to 512 then :D

Akselmo:
> Looks nice so far! The only things I found so far is when you try to add a tag with a `,` in it.
> 
> Example:
> 
>     * In QOwnNotes add tag `new,1`, shows up correctly in the UI
> 
>     * Press save
> 
>     * Go to Nextcloud Tasks, see that there is a tag `new` and `1`
> 
> 
> When I try to `\` to escape it, the `\` gets doubled every time to `\\` (so that's a 2nd issue).
> 
> Adding tags with `,` does work in Nextcloud Tasks.

The commas should now work, it took some head-against-wall action but your testcase now works for me.

If you add two backslashes then it gets turned into one, because otherwise commas break. Honestly if someone is using `\\` in their categories then they should be ashamed. :joy: Single backslash works though.

pbek:
Thank you `,` and one `\` do work now!

> If you add two backslashes then it gets turned into one, because otherwise commas break.

Nextcloud Tasks does manage to use two `\`, so it is possible in iCal! 😉
I added `test\\test` in Nextcloud Tasks and after a reload it was still there. In QOwnNotes it was broken and it after storing the task there it was also broken in Nextcloud Tasks...

Akselmo:
> Thank you `,` and one `\` do work now!
> 
> > If you add two backslashes then it gets turned into one, because otherwise commas break.
> 
> Nextcloud Tasks does manage to use two `\`, so it is possible in iCal! 😉 I added `test\\test` in Nextcloud Tasks and after a reload it was still there. In QOwnNotes it was broken and it after storing the task there it was also broken in Nextcloud Tasks...

Well, I think they now work. At least when adding `test\\test` from nextcloud side.

I am very tempted to add a popup that says "please dont add backslashes, i beg you" if user tries to add them. :laughing:

pbek:
How do the tags with `\\` from Nextcloud Tasks look in the iCal file? And can't we just do the same here?

Akselmo:
In those cases it's just `\\\\` 

I've tried to save them as usual but I don't understand why they get suddenly replaced by `\\`

Akselmo:
Okay, i think i finally got it. Please give it a try, I can't repro the issues you had earlier. :tada:

pbek:
I added a tag `\test\`, `\\test\\` and `\\\test\\\`, this is what happened after a *Save*:

image


And another thing, are you aware that your "Priority" code also found its way into this PR? 😉

Akselmo:
> I added a tag \test\, \\test\\ and \\\test\\\, this is what happened after a Save:

Agh, i think i need to rework the whole system. Back to the drawing board...

> And another thing, are you aware that your "Priority" code also found its way into this PR? 😉

No, I wasn't :joy:  thanks for spotting that, I guess I made branch out of wrong branch... :facepalm:

Akselmo:
Wait, priority? I worked on progress. Priority was untouched here.

Akselmo:
> I added a tag \test\, \\test\\ and \\\test\\\, this is what happened after a Save

Funnily enough this seems to happen on Nextcloud too if you add those tags and refresh the page

pbek:
> Wait, priority? I worked on progress. Priority was untouched here.

Ah, sorry. My bad. Sorry for startling you!

pbek:
> Funnily enough this seems to happen on Nextcloud too if you add those tags and refresh the page

Yes, they have their own kind of bugs...

That's why I now tested in jtx Board, and they seem to do well.
I created some really nasty tags:

Screenshot_20240420_132431_jtx Board

And this is how it showed up in the database:

```ical
BEGIN:VCALENDAR
VERSION:2.0
PRODID:+//IDN bitfire.at//ical4android (at.techbee.jtx)
BEGIN:VTODO
DTSTAMP:20240420T112332Z
UID:8c9399bb-9a90-4ed8-8815-ae9d04ac85fa
SEQUENCE:20
CREATED:20230530T155448Z
LAST-MODIFIED:20240420T112256Z
SUMMARY:Card item
DESCRIPTION:test
STATUS:NEEDS-ACTION
CATEGORIES:#FirstSteps,\\test,\\test\\,\\test2\\,\\test\\\,,\,\\test\\
PERCENT-COMPLETE:0
PRIORITY:0
DUE:20230530T155451
END:VTODO
END:VCALENDAR
```

So now you have an example of how escaping might work.

Akselmo:
Awesome, thanks, this will help. Will use it for testing as well.

pbek:
GPT4 said:

In summary, to correctly escape characters for CATEGORIES: or any other textual property in an iCal file:

- Replace any backslash (`\`) with a double backslash (`\\`).
- Replace any semicolon (;) with `\;`.
- Replace any comma (,) with `\,`.
- To include a newline within a property value, use `\n` or `\N`.

Akselmo:
Okay, another try!

I think I figured out a regex that works with this.

image

image

src/dialogs/tododialog.ui
Akselmo: Ah I see what's going on. It reordered all the items since they were in wrong orders, like row=1 after row=2 in some items.

Visually it changes nothing, Qt Designer just reordered the file.

pbek:
I created a new card item, added a tag `,\crazy tag1`, which showed up correctly, but after storing this happened:

image

Akselmo:
Seems there is many possible edge cases here. I don't know if there is one true solution for this, except for listing all possible edge cases and parsing through those...

Akselmo:
Well, maybe this hellish beast of a regexp will help

image

pbek:
Have you tried using previous examples as tags in the dialog? (not just in a "regexp-helper") 😉

pbek:
Looks good to me! 🥳🎉

pbek:
But builds are failing for older Qt versions: https://github.com/pbek/QOwnNotes/actions/runs/8773496883/job/24143486722?pr=3007#step:15:255

Akselmo:
> But builds are failing for older Qt versions: https://github.com/pbek/QOwnNotes/actions/runs/8773496883/job/24143486722?pr=3007#step:15:255

Seems I had to wrap the `Qt::SkipEmptyParts` in a version check.

PR #3015, opened 2024-04-25
Added some swedish translation

AShadedBlobfish:
I'm not a native speaker and I'm an amateur translator, but I've done some basic translations that were needed


pbek:
Thank you very much, but I'm sorry. That file will automatically be generated by crowdin... You'd need to add the text there, see https://www.qownnotes.org/contributing/translation.html.

PR #3035, opened 2024-06-01
Create clickable sha256 file for QownNotes.zip

### LINKED ISSUE #2938: [FEATURE] Releases SHA256 file: Add file name of the ZIP
The SHA256 file included in the release is missing the name of the ZIP file for commands like `sha256sum --check QOwnNotes.zip.sha256` to work.

Its' content is currently:

```
fb95c0816e2efc32d631b0f2b9fb555359b7c73cdaf810d70dd84c8ecbc8c594

```

But it'd preferably be:

```
fb95c0816e2efc32d631b0f2b9fb555359b7c73cdaf810d70dd84c8ecbc8c594 *QOwnNotes.zip

```

o-l-a-v:
Implements feature request #2938 without breaking existing .sha256 file included in releases.

Also:

* Changed PowerShell aliases into the actual commands they refer to.
* Make release directory with PowerShell `New-Item` instead of mkdir.

### DIFF
--- a/.github/workflows/build-release-qt6.yml
+++ b/.github/workflows/build-release-qt6.yml
@@ -310,8 +310,8 @@ jobs:
       - name: '📤 Upload artifact: Windows clickable sha256sum'
         uses: actions/upload-artifact@v4
         with:
-          name: QOwnNotes.zip.clickable.sha256
-          path: QOwnNotes.zip.clickable.sha256
+          name: QOwnNotes.zip.sha256sum
+          path: QOwnNotes.zip.sha256sum
 
 #      - name: '📤 Release ZIP'
 #        uses: actions/upload-release-asset@v1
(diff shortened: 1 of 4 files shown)


pbek:
Thank you!
I'm not sure about the suffix `clickable`. What does it mean or what should the word symbolize? Is there a more common word for what you are trying to achieve?

o-l-a-v:
Naming is hard. 🙂 I have no preferences, I can set it to whatever you'd like.

It's clickable in the sense that at least on my system I can double click it and the checksum will be compared by some GUI program. I think this one: <https://github.com/idrassi/HashCheck>.

Ideally the name would be "nameof.file.sha256", but that's taken already.

pbek:
Hm, is there any default tool under Windows that is using those checksum files? Under Linux, it would be sha256sum.

o-l-a-v:
I don't think there is a native tool that can consume it as is, you'd have to do some PowerShell unless you installed something like the one I linked.

What breaks if changing the sha256 file to include the file name?

pbek:
Nothing from me, but I don't know about 3rd party workflows...
Why just don't you name it `QOwnNotes.zip.sha256sum`, like the others...

image

o-l-a-v:
Sure, lets do that. I'll try to update the PR tomorrow.

pbek:
Thank you!

PR #3069, opened 2024-08-03
Fix qt6 release

Waqar144:


pbek:
The regular Qt6 builds, like in https://github.com/pbek/QOwnNotes/actions/runs/10221164105 then would then also need fixing.

Waqar144:
> The regular Qt6 builds, like in https://github.com/pbek/QOwnNotes/actions/runs/10221164105 then would then also need fixing.

Those are irrelevant to this, but the changes here can be used to fix those windows builds as well after this is done.

With qmake, the build will only work if Qt > 6.5.4 or 6.6.1. I don't see any reason to support n different versions of Qt on windows though so should be simple if you drop the support for older versions.

Waqar144:
This is done from my side. The snap qt6 build doesn't work but I don't really understand snaps that well so will leave that. Mac/windows builds now work.

Windows qt6 ci is failing, but that is not the goal of this MR.

pbek:
Wow, that's amazing! 
I tried the AppImage and got a `Unable to establish a database connection with file '/home/omega/Nextcloud/Notes/notes.sqlite'. Are the folder and the file writeable?`. I wasn't able to read any notes.

Can anyone access the artifacts of https://github.com/pbek/QOwnNotes/actions/runs/10236720034?pr=3069?
@vinyatu, could you maybe please test the Windows and AppImage version?

@mystery-z and @ JOduMonT, could you maybe please test the macos version?

pbek:
> With qmake, the build will only work if Qt > 6.5.4 or 6.6.1. I don't see any reason to support n different versions of Qt on windows though so should be simple if you drop the support for older versions.

The only reason could be WindowsXP, but that's built on AppVeyor anyway and not on GitHub. And that's really ancient...

build-systems/github/snap.qt6/snapcraft.yaml
```diff
@@ -1,6 +1,5 @@
 name: qownnotes
 version: git
-version-script: cat src/version.h | sed "s/[^0-9,.]//g"
```
pbek, line 3: The version needs to be set somehow...

pbek:
For the Windows release... I have no idea what happens if the updater batch fine copies the Qt6 DLLs to the Qt5 DLLs. 😨

Waqar144:
Macos package will most likely work since it was mostly just fixing the compilation.

Windows, I am not sure about, the packaging script assumes mingw, it may need modifications.(or not) Only testing will tell. Also, it's something that will need to be fixed by someone who has access to a windows machine.

.github/workflows/build-release-qt6.yml
```diff
@@ -261,7 +261,7 @@ jobs:
     runs-on: windows-2019
     env:
       # Qt 6.5 windeployqt doesn't deploy DLLs!
-      QT_VERSION: 6.4.2
+      QT_VERSION: 6.6.1
```
pbek, line 264: I tested the Windows version, there are some DLLs missing. The App doesn't start. I guess windeployqt is still not working correctly in Qt 6.6.1 😠

build-systems/github/snap.qt6/snapcraft.yaml
Waqar144: Yes. But not sure how. I would suggest reusing the qt5 snap script, if that works then porting it to Qt6 should be trivial

The snap changes in this MR are mostly experiments, feel free to take over

.github/workflows/build-release-qt6.yml
Waqar144: Does it say which dlls?

pbek: Qt6PrintSupport, Qt6Widgets, Qt6Gui, Qt6Sql, then it stops.

pbek:
> Windows, I am not sure about, the packaging script assumes mingw, it may need modifications.(or not) Only testing will tell. Also, it's something that will need to be fixed by someone who has access to a windows machine.

The problem with the updater script is it is deployed with the update and will be used only with the next update, so if something breaks, it's game over. The application needs to be downloaded and unzipped manually by everyone then. 😅

.github/workflows/build-release-qt6.yml
Waqar144: I see all the needed dlls in the log

```
-a---          11/20/2023 12:18 PM        1765520 Qt6Core5Compatd.dll
-a---          11/17/2023  5:41 AM       16518800 Qt6Cored.dll
-a---          11/17/2023  5:41 AM       25592464 Qt6Guid.dll
-a---          11/17/2023  5:41 AM        4366480 Qt6Networkd.dll
-a---          11/17/2023  5:41 AM        4741264 Qt6OpenGLd.dll
-a---          11/17/2023  5:42 AM        1072272 Qt6PrintSupportd.dll
-a---          11/20/2023  6:14 AM       16275600 Qt6Qmld.dll
-a---          11/20/2023  6:13 AM        2558096 Qt6QmlModelsd.dll
-a---          11/20/2023  6:13 AM         248464 Qt6QmlWorkerScriptd.dll
-a---          11/20/2023  6:15 AM       18922128 Qt6Quickd.dll
-a---          11/17/2023  5:42 AM         816272 Qt6Sqld.dll
-a---          11/20/2023  5:35 AM        1125520 Qt6Svgd.dll
-a---          11/20/2023 12:05 PM         613520 Qt6WebSocketsd.dll
-a---          11/17/2023  5:42 AM       16437392 Qt6Widgetsd.dll
-a---          11/17/2023  5:42 AM         412304 Qt6Xmld.dll
```

They are debug dlls though. Release mode needs to be explicitly specified when running qmake I think.

Waqar144:
> The problem with the updater script is it is deployed with the update and will be used only with the next update, so if something breaks, it's game over. The application needs to be downloaded and unzipped manually by everyone then. 😅

Not sure I understand you fully. But I assume there is a way to stop that until qt6 package works. The Qt5 build is there and can be used.

.github/workflows/build-release-qt6.yml
pbek: Hm, I see they are ending with `d` 🤔

Waqar144: Its because of debug arg in `windeployqt --debug QOwnNotes.exe`. Probably should let windeployqt figure this out by itself. Also maybe `.` should be specified instead of QOwnNotes.exe so that it scans the whole build folder for dependencies

pbek:
> Not sure I understand you fully. But I assume there is a way to stop that until qt6 package works. The Qt5 build is there and can be used.

If the Windows build would / will be done with Qt6, the updater script will just copy all the files of the ZIP into the current QOwnNotes directory, it is not deleting files (I didn't dare to do that because of the portable mode and me not knowing what else there is in the directory). So if the old files will stir up something, I can't do anything about it...

.github/workflows/build-release-qt6.yml
pbek: 🤷🏻

Waqar144:
> > Not sure I understand you fully. But I assume there is a way to stop that until qt6 package works. The Qt5 build is there and can be used.
> 
> If the Windows build would / will be done with Qt6, the updater script will just copy all the files of the ZIP into the current QOwnNotes directory, it is not deleting files (I didn't dare to do that because of the portable mode and me not knowing what else there is in the directory). So if the old files will stir up something, I can't do anything about it...

Ah i see. Yes, that won't be a good idea. If anyone wants to use Qt6, it should be a clean install.

Also to be clear, Qt5 package is still needed because Qt6 doesn't support windows 7 afaik.

pbek:
> Also to be clear, Qt5 package is still needed because Qt6 doesn't support windows 7 afaik.

Ah, I almost forgot about Qt dropping support for old versions of Windows (and macOS)

As far as the updater script and the whole update mechanism in QOwnNotes is concerned, there can only one release for Windows (or macOS). And even if someone would use the Qt6 build, the updater script would overwrite it with that one release next time (Qt5 if that would stay the same).

Regardless if the built apps do work or not, I guess I can accept the PR. At least it's a good start. 👍️

PR #3070, opened 2024-08-04
Fix Qt6 cmake build

Waqar144:
qt5compat is no longer needed

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,6 +1,17 @@
 cmake_minimum_required(VERSION 3.2.2)
 project(QOwnNotes)
 
+# BEGIN options
+option(QON_QT6_BUILD "Build QOwnNotes with Qt6" OFF)
+# END options
+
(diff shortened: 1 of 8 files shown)


src/CMakeLists.txt
```diff
@@ -562,27 +581,27 @@ target_sources(QOwnNotes PRIVATE
...
     Google::DiffMatchPatch
     md4c-html
 )
-if(UNIX)
-    target_link_libraries(QOwnNotes PRIVATE
-        Qt5::X11Extras
-        ${X11_LIBRARIES}
-    )
-endif()
+# if(UNIX)
```
pbek, line 599: Aren't those needed for global shortcuts? 🤔

pbek:
Other than that... Thank you very much!

src/CMakeLists.txt
Waqar144: If it was needed it would fail to link, however its working fine here. Also, .pro file doesn't include this module as well.

PR #3071, opened 2024-08-05
Dont deploy debug dlls

Waqar144:
- Use release dlls for windows
- Raise min cmake version to 3.5 to silence warnings
- Change QOwnNotes.exe directory to `build` in cmake so that its the same as qmake


pbek:
Thank you!

PR #3077, opened 2024-08-06
Fix cmake install

Waqar144:
It now works and produces the same output as qmake. Also you don't need to run lrelease manually anymore with cmake.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -27,7 +27,8 @@ include(libraries/sonnet/src/plugins/hunspell/hunspell/CMakeLists.txt)
 
 # noneed for md2html
 set(BUILD_MD2HTML_EXECUTABLE FALSE)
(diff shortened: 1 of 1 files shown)


Waqar144:
One thing that is left a la cmake is that it still compiles the nssspellchecker on macos. That was removed long ago and we use hunspell everywhere IIRC so cmake should do the same.

Waqar144:
> One thing that is left a la cmake is that it still compiles the nssspellchecker on macos. That was removed long ago and we use hunspell everywhere IIRC so cmake should do the same.

Done

src/CMakeLists.txt
```diff
@@ -473,16 +474,6 @@ list(APPEND SOURCE_FILES
 )
 endif()
```
pbek, line 475: This line needs to be remove too then, right?

pbek:
Thank you!

Issue #3076, opened 2024-08-06
[FEATURE] Nix CMake build process


pbek:
As requested by @NickCao in https://github.com/NixOS/nixpkgs/pull/332556#issuecomment-2269875889, it would be great to build QOwnNotes with CMake for nixpkgs.

## Todo

- [x] Make build process work with nix
- [x] Get translations working
- [x] Scripting engine works
- [x] Add CI process with CMake for Qt5
- [x] Add CI process with CMake for Qt6
- [x] Update config for nixpkgs

pbek:
@Waqar144, I was trying to build QOwnNotes with Qt6 and cmake in https://github.com/pbek/QOwnNotes/blob/feature/nix-cmake/default.nix via https://github.com/pbek/QOwnNotes/blob/08c1657991e650ede70e950edf1b1ab3dabb5a07/Makefile#L16-L17

I got the error message:

```
CMake Error at libraries/botan/CMakeLists.txt:14 (find_package):
  By not providing "FindQt5.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Qt5", but
  CMake did not find one.

  Could not find a package configuration file provided by "Qt5" with any of
  the following names:

    Qt5Config.cmake
    qt5-config.cmake

  Add the installation prefix of "Qt5" to CMAKE_PREFIX_PATH or set "Qt5_DIR"
  to a directory containing one of the above files.  If "Qt5" provides a
  separate development package or SDK, be sure it has been installed.
```

Does this ring any bell for you?

pbek:
Note that I tried to use:

https://github.com/pbek/QOwnNotes/blob/08c1657991e650ede70e950edf1b1ab3dabb5a07/default.nix#L47-L49

Have you ever build QOwnNotes with Qt6 and cmake? I only built it in Qt5 with cmake, as far as I remember...

pbek:
`QON_QT6_BUILD=ON` got me one step further! 👍️

I'm now at:

```
CMake Error at libraries/botan/CMakeLists.txt:21 (find_package):
  By not providing "FindBotan2.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Botan2", but
  CMake did not find one.

  Could not find a package configuration file provided by "Botan2" with any
  of the following names:

    Botan2Config.cmake
    botan2-config.cmake

[...]

PR #3080, opened 2024-08-08
Compile md4c into QOwnNotes

Waqar144:
Ideally QOwnNotes should use the system md4c but its not possible atm.


pbek:
> Ideally QOwnNotes should use the system md4c but its not possible atm.

So what is the PR doing, solving or improving? Is it for #3076? Don't we need https://github.com/qownnotes/md4c?

pbek:
Ah, really for #3076, thank you very much!

Waqar144:
> > Ideally QOwnNotes should use the system md4c but its not possible atm.
> 
> So what is the PR doing, solving or improving? Is it for #3076? Don't we need https://github.com/qownnotes/md4c?

Just treating md4c as part of QOwnNotes source instead of linking it into QOwnNotes. This is what we do in the qmake file.

PR #3122, opened 2024-10-04
Initialize imageWidth

gbabin:
`imageWidth` must be initialized because `OwnCloudService::nextcloudPreviewImageTagToInlineImageTag` and `Utils::Misc::remotePreviewImageTagToInlineImageTag` can return without affecting `imageWidth`.


pbek:
What issue does this fix? Could you please open one and describe it?

pbek:
I see, thank you! 😊

PR #3148, opened 2024-11-02
refactor: replace non-empty QString constructors with QStringLiteral()

Integral-Tech:
- Qt provides a macro [QStringLiteral()](https://doc.qt.io/qt-6/qstring.html#QStringLiteral), which makes constructing QString objects from string literals more efficient.

### DIFF
--- a/src/entities/notefolder.cpp
+++ b/src/entities/notefolder.cpp
@@ -530,12 +530,12 @@ bool NoteFolder::isPathNoteFolder(const QString &path) {
 
 void NoteFolder::setSettingsValue(const QString &key, const QVariant &value) {
     SettingsService settings;
-    settings.setValue(QString("NoteFolder-%1/%2").arg(QString::number(id), key), value);
+    settings.setValue(QStringLiteral("NoteFolder-%1/%2").arg(id).arg(key), value);
 }
 
 QVariant NoteFolder::settingsValue(const QString &key, const QVariant &defaultValue) const {
     const SettingsService settings;
-    return settings.value(QString("NoteFolder-%1/%2").arg(QString::number(id), key), defaultValue);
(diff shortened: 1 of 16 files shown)


src/entities/notefolder.cpp
```diff
@@ -530,12 +530,12 @@ bool NoteFolder::isPathNoteFolder(const QString &path) {
 
 void NoteFolder::setSettingsValue(const QString &key, const QVariant &value) {
     SettingsService settings;
-    settings.setValue(QString("NoteFolder-%1/%2").arg(QString::number(id), key), value);
```
pbek, line 533: Did you test the code? 😅 This breaks the function.

src/entities/notefolder.cpp
```diff
@@ -530,12 +530,12 @@ bool NoteFolder::isPathNoteFolder(const QString &path) {
...
+    settings.setValue(QStringLiteral("NoteFolder-%1/%2").arg(id).arg(key), value);
 }
 
 QVariant NoteFolder::settingsValue(const QString &key, const QVariant &defaultValue) const {
     const SettingsService settings;
-    return settings.value(QString("NoteFolder-%1/%2").arg(QString::number(id), key), defaultValue);
```
pbek, line 538: Same here.

src/entities/notefolder.cpp
Integral-Tech: Oh, sorry :(

pbek: At least it started to throw warnings and I have a bad feeling, can you please stick to just replacing the QString. 😉

Integral-Tech: So... are failed tests due to this change?

pbek: There are no tests for testing if your change breaks the string. But it seems ok. Thank you.

PR #3219, opened 2025-01-28
Update openaiservice.cpp

Weej1:
Hi @pbek! Like you recommended I (attempted) a Pull Request to update the list of LLM models available in Groq. Let me know if this works.  Changes I made: 

- Replaced: `gemma-7b-it` (since I can't find it in the list on groq anymore) with `gemma2-9b-it`

- Replaced: `llama2-8b-8192` with `llama3-8b-8192` (for same reason above)

 - Did not add: the 3 whisper models because they  primarily process Audio files (unless QoN is also part Audacity which I'm not aware of!)


pbek:
Thank you very much!

PR #3230, opened 2025-02-10
Added ability to properly remove icons from menus (for a cleaner UX)

andrei-a-papou:
Hi, and thank you for creating QOwnNotes. I've been using it for a couple of weeks now (migrated from a legacy note-taking app) and it looks very impressive so far! Packed with features, very customizable and scriptable.

I'm using it on a mostly GTK-based Linux desktop where menus do not have icons and would like to have the same option in QOwnNotes for a cleaner user experience.

I'm very new to Qt and it took some research to properly remove the icons. 

First, I've tried to do it the way macOSX icons are removed in the QOwnNotes code-base (`Qt::AA_DontShowIconsInMenus`), but this method leaves empty spaces in place of the icons (at least on Linux), which doesn't look good at all, IMO. 

Another option is to simply iterate over all the actions and set icons to null. This removes the icons and the empty space, but kill icons in toolbars too.

After experimenting, I've found that the best way is to use QProxyStyle. Hence this patch.

I'm not interested in maintaining my own fork :), so would appreciate it if you could merge it on your end. I've noticed some other little bugs/quirks here and there. I'll try to report them soon as time permits.

Overall, QOwnNotes looks like a great app with a ton of useful features. So thank you again for creating it and let me know if the patch can be merged.


src/mainwindow.h
```diff
@@ -80,6 +84,24 @@ struct TagHeader;
 class TodoDialog;
 class SettingsDialog;
 
+/* See: https://stackoverflow.com/questions/44745459/remove-icon-space-from-qmenu */
+class NoMenuIconStyle : public QProxyStyle {
+public:
+    void drawControl(ControlElement element, const QStyleOption *option, QPainter *painter, const QWidget *widget) const override {
```
pbek, line 90: Can you please not do a class implementation in the header file? Rather, do it in a CPP file.

pbek:
There already is a rule for macOS 🤔:

https://github.com/pbek/QOwnNotes/blob/0d11bd1fbd899285cd17d4cae7ef861a3f11ebf6/src/mainwindow.cpp#L152-L155

We need to do something about that too...

andrei-a-papou:
I don't have a way to test things on a Mac, but would it be OK to limit the patch to Linux and Windows via #ifdef and such? The existing Mac logic can stay the way it is. Or it can be removed (but I can't test the result on a Mac.)

src/mainwindow.h
andrei-a-papou: Sure, I'll give it a go

pbek:
First one other question: Why was `QApplication::instance()->setAttribute(Qt::AA_DontShowIconsInMenus, true);` not enough for your use-case?

src/mainwindow.h
pbek: I guess we don't need `NoMenuIconStyle`, since `Qt::AA_DontShowIconsInMenus` also does the same.

src/dialogs/settingsdialog.ui
```diff
@@ -3209,6 +3209,16 @@
                        </property>
                       </widget>
                      </item>
+                     <item row="4" column="0">
+                      <widget class="QCheckBox" name="disableIconsInMenusCheckBox">
+                       <property name="toolTip">
+                        <string>Disable icons in menus as well as the "Find action" dialog. Requires restart to take effect</string>
+                       </property>
+                       <property name="text">
+                        <string>Disable menu icons</string>
```
pbek, line 3218: Can you please use `Hide` instead of `Disable`everywhere (text and code)?

andrei-a-papou:
> First one other question: Why was `QApplication::instance()->setAttribute(Qt::AA_DontShowIconsInMenus, true);` not enough for your use-case?

No, not enough, that's the whole point. The `Qt::AA_DontShowIconsInMenus` removes the icons, but leaves a lot of ugly empty space where the icons used to be. IMO, this makes the menus look worse, not better, especially if the menu contains checkboxes.

src/mainwindow.h
andrei-a-papou: > First, I've tried to do it the way macOSX icons are removed in the QOwnNotes code-base (Qt::AA_DontShowIconsInMenus), but this method leaves empty spaces in place of the icons (at least on Linux), which doesn't look good at all, IMO.

That's the thing: they don't do the same thing, unfortunately :)

pbek:
> just set the default for your setting to be true for macOS by default. Of course, the settings reading needs to go into a method then

ok, then still ☝🏻

src/dialogs/settingsdialog.ui
andrei-a-papou: Done

andrei-a-papou:
> > just set the default for your setting to be true for macOS by default. Of course, the settings reading needs to go into a method then
> 
> ok, then still ☝🏻 and the builds are still failing, because of the duplicate declaration (how did the build work for you? 

You're right. The build worked for me because the patch I provided depends on another patch (not included here for now).

Fixed it here. It now compiles on my end when I do a clean `qmake && make` from the github source tree.

src/mainwindow.h
andrei-a-papou: > Can you please not do a class implementation in the header file? Rather, do it in a CPP file.

Done, created `NoMenuIconStyle.cpp` and `NoMenuIconStyle.h`, updated the _pro_ file. Compiles and works on my end.

andrei-a-papou:
@pbek  Some builds failed due to _undefined reference to `vtable for NoMenuIconStyle'_, which is fixed now.

pbek:
Please use lower case characters for `NoMenuIconStyle.*` and move the files to the `src/helpers` folder.
Please also add the two files to `src/CMakeLists.txt`.

src/utils/misc.cpp
```diff
@@ -1300,6 +1300,15 @@ bool Utils::Misc::isNoteEditingAllowed() {
     return SettingsService().value(QStringLiteral("allowNoteEditing"), true).toBool();
 }
 
+/**
+ * Returns if "hideIconsInMenus" is turned on
+ *
+ * @return
+ */
+bool Utils::Misc::hideIconsInMenus() {
+    return SettingsService().value(QStringLiteral("hideIconsInMenus"), true).toBool();
```
pbek, line 1309: Default needs to be true for macOS and false for all other operating systems. 😉

src/utils/misc.cpp
```diff
@@ -1300,6 +1300,15 @@ bool Utils::Misc::isNoteEditingAllowed() {
     return SettingsService().value(QStringLiteral("allowNoteEditing"), true).toBool();
 }
 
+/**
+ * Returns if "hideIconsInMenus" is turned on
+ *
+ * @return
+ */
+bool Utils::Misc::hideIconsInMenus() {
```
pbek, line 1308: Please choose a method name that starts with, for example `is`, the method isn't actually hiding icons.

src/main.cpp
```diff
@@ -192,6 +194,17 @@ int mainStartupMisc(const QStringList &arguments) {
...
     }
 
+    // apply custom style to hide menu icons
+    const bool hideMenuIcons = Utils::Misc::hideIconsInMenus();
+    if (hideMenuIcons) {
+        QApplication::setStyle(new NoMenuIconStyle);
+    }
+
+#ifdef Q_OS_MAC
+    QApplication::setStyle(new NoMenuIconStyle);
```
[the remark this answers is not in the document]
andrei-a-papou: OK, no forcing, but can we just set a default like I did here? https://github.com/pbek/QOwnNotes/pull/3230/commits/259c49fe26e98722a70bd0534e1c2693f1ae54cd

I've tested it under Linux with #ifdef Q_OS_LINUX instead of #ifdef Q_OS_MAC and it seems to work fine. 

When no setting has been defined, the app hides icons on macOS. If a user ticks the checkbox, the app reads that setting instead.

src/utils/misc.cpp
andrei-a-papou: I've made it start with the verb `are` :)

andrei-a-papou:
> Please use lower case characters for `NoMenuIconStyle.*` and move the files to the `src/helpers` folder. Please also add the two files to `src/CMakeLists.txt`. And please run `./scripts/clang-format-project.sh` on your changes...

Renamed and moved the files, but the `clang-format-project.sh` doesn't work on my machine. This is the output of `clang-format --version`:

```
clang-format --version
Debian clang-format version 11.0.1-2
```

Running the script produces a bunch of "not found" errors...

pbek:
> Debian clang-format version 11.0.1-2

omg, I'm, at clang-format version 19.1.5

src/utils/misc.cpp
pbek: You could instead of setting `true` just use a variable that is set to false and set it to true with a compiler directive for macOS.

andrei-a-papou: I'm sorry, I don't quite follow... Do you mean we shouldn't use `if (!settings.contains(QStringLiteral("hideIconsInMenus"))) { ... }`? if so, why?

I'm new to Qt and the code-base, so if you think there's a better way to do the macOS check, please feel free to give it a quick edit :)

src/main.cpp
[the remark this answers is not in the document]
pbek: I was talking about a variable for the default value, like here:

https://github.com/pbek/QOwnNotes/blob/c6c35898fd5aa9a1a78f5d436c3593796e2a0bb1/src/dialogs/settingsdialog.cpp#L1058-L1066

pbek:
> Can you please run the script then?

I will.

src/utils/misc.cpp
pbek: See https://github.com/pbek/QOwnNotes/pull/3230#discussion_r1950261532

pbek:
>  Please also add the two files to src/CMakeLists.txt.

Still missing.

andrei-a-papou:
> src/CMakeLists.txt

Apologies, fixed: https://github.com/pbek/QOwnNotes/pull/3230/commits/61c622dbab2ce96e72a3f25ffe6de76d522545a0

src/main.cpp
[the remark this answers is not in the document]
andrei-a-papou: OK, how about this? https://github.com/pbek/QOwnNotes/pull/3230/commits/7f7cc8c3ea5d1e52dba067e83f6ef76acb4f86c8

src/utils/misc.cpp
```diff
@@ -1300,6 +1300,29 @@ bool Utils::Misc::isNoteEditingAllowed() {
...
+#ifdef Q_OS_MAC
+    // hide icons in menus by default on macOS
+    bool hideIconsInMenusDefault = true;
+#else
+    bool hideIconsInMenusDefault = false;
+#endif
+
+    SettingsService settings;
+
+    if (!settings.contains(QStringLiteral("hideIconsInMenus"))) {
```
pbek, line 1318: You don't need that whole if-clause, because there is no old setting to migrate from.

src/utils/misc.cpp
```diff
@@ -1300,6 +1300,29 @@ bool Utils::Misc::isNoteEditingAllowed() {
...
+        settings.value(QStringLiteral("hideIconsInMenus"), hideIconsInMenusDefault).toBool();
+        return hideIconsInMenusDefault;
+    }
+
+    return settings.value(QStringLiteral("hideIconsInMenus"), true).toBool();
```
pbek, line 1323: Just use `hideIconsInMenusDefault` instead of `true`.

pbek: And you can use `return SettingsService().value(...` instead of declaring `SettingsService settings`, you only need to read a setting once.

andrei-a-papou: OK, thanks for the feedback, see https://github.com/pbek/QOwnNotes/pull/3230/commits/748cdc38e2f3b33fe840cd993b1d118e9b784c5d

pbek:
Oh, there is something important that slipped my review:

https://github.com/pbek/QOwnNotes/blob/da690c2e88dafb01c39a934a00ceb8e8f3dbc448/src/main.cpp#L192-L201

You are breaking the interface style setting with your QApplication::setStyle!

pbek:
I'll disable your `QApplication::setStyle(new NoMenuIconStyle);` in favor for `QApplication::setAttribute(Qt::AA_DontShowIconsInMenus, true);` for now. I still can't see a difference between the two in the menu. Could you maybe provide screenshots where you see a difference on your system?

andrei-a-papou:
> I'll disable your `QApplication::setStyle(new NoMenuIconStyle);` in favor for `QApplication::setAttribute(Qt::AA_DontShowIconsInMenus, true);` for now until you come up with a different idea.

This is very frustrating, I'm sorry to say :) Because now it looks broken...

> But I still can't see a difference between the two in the menu. 🤔 Could you maybe provide screenshots where you see a difference on your system?

Sure. The first screenshot uses `Qt::AA_DontShowIconsInMenus`.

image

The second screenshot uses the PR that actually deals with the problem:

image

pbek:
No difference for under KDE Plasma...

## QApplication::setAttribute(Qt::AA_DontShowIconsInMenus, true);

image

## QApplication::setStyle(new NoMenuIconStyle);

image

pbek:
Maybe you can find a solution, that will still keep the interfaceStyle setting.

andrei-a-papou:
It's interesting that `Qt::AA_DontShowIconsInMenus` works under KDE Plasma. Maybe it's a Qt5 vs Qt6 thing, who knows, but I know for sure I'm not the only one having this problem: https://stackoverflow.com/questions/44745459/remove-icon-space-from-qmenu

Would you accept a patch  that relies on `NoMenuIconStyle`, but enables the option to hide icons only if "Interface style" is set to _Automatic_? (macOS can use `Qt::AA_DontShowIconsInMenus` like before.)

pbek:
> Would you accept a patch that relies on NoMenuIconStyle, but enables the option to hide icons only if "Interface style" is set to Automatic?

That would be an option, but you need to keep in mind that the interfaceStyle setting is applied dynamically while the setting dialog is opened and when the selectbox in the dialog changes!

The better solution would be if your style could somehow inherit the interfaceStyle, maybe like in the example from https://doc.qt.io/qt-6/qapplication.html#setStyle.

> (macOS can use Qt::AA_DontShowIconsInMenus like before.)

I guess macOS can use the same handling as above...

andrei-a-papou:
> The better solution would be if your style could somehow inherit the interfaceStyle

Thanks for the pointer. Right now time does not permit me to work on any more patches, but maybe in the future. Good luck!

pbek:
Thank you, and thank you for your contribution!

pbek:
## 25.2.4
- There now is a setting to hide the icons in the main menu and command bar in the *Interface settings*
  (for [#3230](https://github.com/pbek/QOwnNotes/pull/3230), thank you, @andrei-a-papou)

pbek:
Can you please also test the latest main branch with https://github.com/pbek/QOwnNotes/commit/854a3ab396492f5f9c40fb2817428f899fae885a, where I tried to make interfaceStyle work with NoMenuIconStyle?

Can you please test different interfaceStyle settings with hideIconsInMenus set to true and false?

andrei-a-papou:
> Can you please also test the latest main branch with [854a3ab](https://github.com/pbek/QOwnNotes/commit/854a3ab396492f5f9c40fb2817428f899fae885a), where I tried to make interfaceStyle work with NoMenuIconStyle?
> 
> Can you please test different interfaceStyle settings with hideIconsInMenus set to true and false?

The https://github.com/pbek/QOwnNotes/commit/854a3ab396492f5f9c40fb2817428f899fae885a commit works fine, I tested with "Automatic" (which Kvantum on my box), as well as "motif", "windows" and "plastique" styles. Awesome!

I've tried with "hideIconsInMenus" on and off. 

There's one hiccup:  with "hideIconsInMenus" on, if a user changes the interface style, the menu icons reappear until the app is restarted. Seems like prompting for a restart after changing the interface style would make everything work perfectly.

Will be happy to test.

pbek:
> There's one hiccup: with "hideIconsInMenus" on, if a user changes the interface style, the style is applied, but the menu icons reappear until the app is restarted. Seems like prompting for a restart after changing the interface style would make everything work perfectly.

Hah, didn't catch my eyes. Thank you for mentioning. Forgot a stray `QApplication::setStyle()`. I'll fix that.

> P. S. Incidentally, I now remember playing with interface styles when I first installed QOwnNotes, and I think the styles did not always perfectly applied unless I restarted the app. Seems like it's a good idea to prompt for a restart when changing a style anyway.

`QApplication::setStyle()` is supposed to work instantly.
But now they are applying instantly for you too?

andrei-a-papou:
> `QApplication::setStyle()` is supposed to work instantly.
> But now they are applying instantly for you too?

Yes. Everything works as expected, except the hiccup I mentioned above.

pbek:
> Yes. Everything works as expected, except the hiccup I mentioned above.

Ok, great. I'll fix the hiccup in the evening.

pbek:
## 25.2.5
- The main menu and command bar now has a cleaner user experience when icons are hidden
  on GTK-based Linux desktops (for [#3230](https://github.com/pbek/QOwnNotes/pull/3230))

andrei-a-papou:
@pbek Works nicely now, thank you!

pbek:
Great, thank you for testing!

PR #3343, opened 2025-08-18
Modified gentoo ebuild template to build against qt6

nagelp:
- Modernize ebuild and switch from qt5 to qt6
- In build-for-gentoo.sh add functionality to keep only one ebuild per major version (except for the latest major version, where it keeps three) and also maintain the Manifest accordingly.

### DIFF
--- a/build-systems/gentoo/build-for-gentoo.sh
+++ b/build-systems/gentoo/build-for-gentoo.sh
@@ -64,8 +64,14 @@ ARCHIVE_FILE=qownnotes-${QOWNNOTES_VERSION}.tar.xz
 cd ../overlay/app-office/qownnotes/ || exit 1
 cp ../../../QOwnNotes/build-systems/gentoo/qownnotes.ebuild .
 
-# replace the version in the ebuild file
-sed -i "s/VERSION-STRING/$QOWNNOTES_VERSION/g" qownnotes.ebuild
+# create metadata file
+echo '<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "https://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+        <upstream>
+                <remote-id type="github">pbek/QOwnNotes</remote-id>
+        </upstream>
+</pkgmetadata>' >metadata.xml
 
 # update the Manifest file
 echo "DIST ${ARCHIVE_FILE} ${QOWNNOTES_ARCHIVE_SIZE} SHA512 ${QOWNNOTES_ARCHIVE_SHA512}" >>Manifest
@@ -75,7 +81,7 @@ mv qownnotes.ebuild ${eBuildFile}
 
 echo "Committing changes..."
 git add ${eBuildFile}
-git commit -m "releasing version $QOWNNOTES_VERSION" ${eBuildFile} Manifest
+git commit -m "releasing version $QOWNNOTES_VERSION" ${eBuildFile} Manifest metadata.xml
 git push
 
 # remove everything after we are done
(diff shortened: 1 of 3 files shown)


nagelp:
I've tested it locally and it's working on my machine™

build-systems/gentoo/qownnotes.ebuild
```diff
@@ -1,38 +1,25 @@
-# Copyright 2014-2021 Patrizio Bekerle
+# Copyright 2014-2025 Patrizio Bekerle, Patrick Nagel
```
pbek, line 1: Is this line about the package config or the application?
If it's the former, can you please another line. If it's the latter, please remove yourself.

nagelp: It's for the ebuild file itself that I've amended. I'm fine with you being the lone copyright holder of the file, just did what I'm used to doing.

build-systems/gentoo/qownnotes.ebuild
```diff
@@ -41,15 +28,15 @@ src_prepare() {
 }
 
 src_compile() {
-	eqmake5 QOwnNotes.pro -r
+	eqmake6 QOwnNotes.pro -r
```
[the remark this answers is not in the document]
nagelp: I can switch to CMake, but I'd propose to do that in another PR. Let's get QOwnNotes to build with a recent Qt first on Gentoo, so that it becomes usable again. As it is, I don't think anyone is able to even build it, because the many Qt5 dependencies are conflicting with a recent system using Qt6. At least that's what I remember from trying to get to work QOwnNotes on my Gentoo system a couple of weeks back.

I also have an idea for the ebuild generator shell script (in yet another PR), because as it is, it leaves a lot (all?) versions lying around, which results in a long list as seen below. I'd implement a way to thin out the old versions, e.g. only keep the last of every major version? Let me know what you think.

image

build-systems/gentoo/qownnotes.ebuild
pbek: If it's about the ebuild file, then just add another line with 2025. 👍️

pbek: I don't even remember why I added the header myself. ️🤷🏻

build-systems/gentoo/build-for-gentoo.sh
```diff
@@ -64,8 +64,14 @@ ARCHIVE_FILE=qownnotes-${QOWNNOTES_VERSION}.tar.xz
 cd ../overlay/app-office/qownnotes/ || exit 1
 cp ../../../QOwnNotes/build-systems/gentoo/qownnotes.ebuild .
 
-# replace the version in the ebuild file
-sed -i "s/VERSION-STRING/$QOWNNOTES_VERSION/g" qownnotes.ebuild
+# create metadata file
```
pbek, line 94: This is just a static file, isn't it?
Can you please just create it in the `build-systems/gentoo` folder and in the script copy it to where it should go.

build-systems/gentoo/qownnotes.ebuild
nagelp: Done.

build-systems/gentoo/qownnotes.ebuild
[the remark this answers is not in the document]
nagelp: I've implemented it. Looks pretty ugly in bash, but should work...

build-systems/gentoo/build-for-gentoo.sh
nagelp: Done

pbek:
Looks good to me! Thanks a lot!

pbek:
## 25.8.5
- The Gentoo ebuilds were updated to now use Qt6 instead of Qt5
  (for [#3343](https://github.com/pbek/QOwnNotes/pull/3343), thank you, @nagelp)

https://github.com/qownnotes/gentoo-overlay/tree/master/app-office/qownnotes seems to have been updated quite nicely!

nagelp:
Awesome, for me it works (I had of course tested it in a local overlay). But sadly another Gentoo user has an issue: https://github.com/pbek/QOwnNotes/issues/3345 - I've just opened a PR to fix that (https://github.com/pbek/QOwnNotes/pull/3346)

pbek:
It's for scripting... And the web-app.

PR #3418, opened 2025-12-31
metadata: Add bugtracker and vcs-browser URLs

yakushabb:
Add the bugtracker and vcs-browser URLs to point to the related pages.


pbek:
Thank you!

PR #3435, opened 2026-01-14
create 'languages' subdirectory for QM files

alexanderino:
1. a new `languages` subdirectory is created under the `release` directory (line 11)
2. the translation files (*.QM) are copied into `languages` instead of the root directory (line 24)

PR #3436, opened 2026-01-14
delete language files located in root directory of install during update

alexanderino:
When `update.bat` is run during an update, this code removes all language files (with a QM file extension) from the root directory.

Depends on #3435.

### DIFF
--- a/appveyor/update.bat
+++ b/appveyor/update.bat
@@ -38,6 +38,9 @@ echo Deleting temporary files
 del %source%
 del %unzipExecutable%
 echo ======================================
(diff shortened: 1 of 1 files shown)


pbek:
> Depends on https://github.com/pbek/QOwnNotes/pull/3435.

It would be better to put it into the other PR then by committing to your `patch-1` branch.

alexanderino:
> > Depends on #3435.
> 
> It would be better to put it into the other PR then by committing to your `patch-1` branch.

I've merged all three into one PR. Hope that does the trick.

PR #3441, opened 2026-01-15
Update Fedora installation instructions (for F41 and later)

ema-pe:
Hello! I created this pull request to fix the installation instructions for QOwnNotes on Fedora 41 and later. Starting with Fedora 41, dnf5 is used as the default package manager. The former config-manager plugin is now a built-in subcommand, with slightly different command-line flags.

### DIFF
--- a/webpage/src/installation/fedora.md
+++ b/webpage/src/installation/fedora.md
@@ -10,6 +10,18 @@ For most users you can just use `dnf install qownnotes` in a terminal window to 
 If you want the **most up-to-date version**, please continue reading.
 :::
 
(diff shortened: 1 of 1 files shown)


pbek:
Thank you very much!

ema-pe:
@pbek also note that QOwnNotes in the [official Fedora repository](https://packages.fedoraproject.org/pkgs/qownnotes/qownnotes/) is stuck at 25.5.10, and I don't know why. I tried contacting the package maintainers but never got a reply, maybe if you try they'll respond to you since you are the upstream maintainer.

PR #3551, opened 2026-04-11
Fix building against Qt6 in presence of Qt5

fsLeg:
On systems with both Qt6 and Qt5 present (in my case, Slackware 15.0) building QOwnNotes against Qt6 fails with this error:

```
Using Qt version 6
-- Found Cups: /usr/lib64/libcups.so (found version "2.4.16")
-- ccache found
-- Configuring done (4.2s)
CMake Error: The INTERFACE_QT_MAJOR_VERSION property of "Qt5::Core" does
not agree with the value of QT_MAJOR_VERSION already determined
for "QOwnNotes".
```

I managed to find the culprit: qlitehtml. Its CMakeLists.txt looks specifically for Qt5 first despite explicitly requiring Qt6. Removing the first instance of find_package() should be fine as QT_VERSION_MAJOR is defined by the parent CMakeLists.txt.

After this change both Qt6 and Qt5 versions build with no issues on a system with both Qt6 and Qt5 present.

### DIFF
--- a/src/libraries/qlitehtml/CMakeLists.txt
+++ b/src/libraries/qlitehtml/CMakeLists.txt
@@ -27,10 +27,6 @@ else()
 endif()
 
 find_package(
-  QT 5.11 NAMES Qt6 Qt5
-  COMPONENTS Widgets
-  REQUIRED)
-find_package(
   Qt${QT_VERSION_MAJOR}
   COMPONENTS Widgets
   REQUIRED)


pbek:
Won't that just only work for QT 5.11? And with none of the other Qt 5 versions?

fsLeg:
Slackware 15.0 has Qt 5.15.3, and it still works.

pbek:
Ah, you removed the code. Sorry. Thank you!

---

Now I have the full picture. Here's the analysis:

**The change is correct and works for all supported Qt 5 versions.**

Here's why pbek's concern about "Qt 5.11" is a misreading of the CMake syntax:

The removed code was:
```cmake
find_package(
  QT 5.11 NAMES Qt6 Qt5
  COMPONENTS Widgets
  REQUIRED)
```

`5.11` here is **not** a minimum Qt version requirement. In CMake's `find_package`, when you use the `NAMES` keyword, the version number (`5.11`) applies to the **versionless `QT` package name**, which is just a CMake compatibility shim — not a real package with a version. This line is essentially CMake's way of finding whichever Qt version is available (Qt6 preferred, then Qt5), and the `5.11` is effectively meaningless/ignored there.

The **actual** problem is that this `find_package(QT ...)` call sets `QT_VERSION_MAJOR` by searching for Qt, and on a system with both Qt6 and Qt5, it may find Qt5 first (since `Qt5` is listed second but CMake might still resolve it differently depending on the system). This conflicts with the `QT_VERSION_MAJOR` already set by the parent `CMakeLists.txt` (which correctly found Qt6), causing the error:

```
The INTERFACE_QT_MAJOR_VERSION property of "Qt5::Core" does not agree
with the value of QT_MAJOR_VERSION already determined for "QOwnNotes".
```

**The fix is valid**: removing the first `find_package` is safe because:
1. `QT_VERSION_MAJOR` is already defined by the parent `CMakeLists.txt`
2. The second `find_package(Qt${QT_VERSION_MAJOR} ...)` correctly uses that inherited variable
3. qlitehtml is a subdirectory — it should rely on the parent's Qt version decision, not rediscover it independently

The `5.11` in the removed line was never a meaningful version guard for Qt 5 compatibility.

PR #3554, opened 2026-04-11
Allow `just src-run` to work on macOS

louwers:
On macOS, a `QOwnNotes.app` is built, so the current command does launch it.


pbek:
Thank you 😉

PR #3555, opened 2026-04-11
Do not delete existing settings and database when closing welcome dialog

louwers:
On macOS, it is possible that there is an issue reading the settings file. This will cause the welcome dialog to be shown, which deletes the existing settings and database when closed.

I haven't been able to find the root cause of this issue (#3548). For some reason, it doesn't seem to happen with debug builds. 😢 

This PR makes sure that the settings are at least preserved...

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -351,9 +351,6 @@ int mainStartupMisc(const QStringList &arguments) {
         WelcomeDialog welcomeDialog;
         // exit QOwnNotes if the welcome dialog was canceled
         if (welcomeDialog.exec() != QDialog::Accepted) {
-            settings.clear();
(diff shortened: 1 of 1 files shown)


pbek:
> This PR makes sure that the settings are at least preserved...

But it leaves settings traces if a new user decides to not use QOwnNotes while in the welcome dialog. That's what the deleting of settings was for. It's better to find the reason why the dialog was triggered.

louwers:
Note: I think I found a workaround.

image

If you add QOwnNotes to "Full Disk Access" in "Privacy & Security" in the system settings, the problem seems to go away.

pbek:
Awesome!

louwers:
Oh no, it's back... It's so random. 

Really puzzling.

PR #3567, opened 2026-04-15
Add menu option to rename current note

louwers:
Add 'Rename current note' menu option under 'Edit'. When selected, it will allow editing the name in the note list (if visible) or in a dialog otherwise. If 'Allow note file name to be different from headline' is disabled, the menu item will be hidden.

Here is a video showing the feature:

https://github.com/user-attachments/assets/856377ce-1cc6-40ee-9500-3c80caa1eeae

I read in the changelog that multiple directories can have a different 'Allow note file name to be different from headline' setting. That is one complication I have not considered.

Written with Copilot assistance.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2400,8 +2400,6 @@ void MainWindow::readSettings() {
         on_actionAllow_note_editing_triggered(isAllowNoteEditing);
     }
 
-    ui->action_Rename_note->setVisible(Note::allowDifferentFileName());
-
     // we want to trigger the event afterward so the settings of the note edits
     // are updated
     const bool centerCursor = settings.value(QStringLiteral("Editor/centerCursor")).toBool();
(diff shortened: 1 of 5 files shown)


pbek:
Why do we need that main menu action? You can already do that in the note list context menu, or by pressing F2 in the note list.

pbek:
This PR also changes existing code, and it's unclear to me why.

louwers:
> Why do we need that main menu action? You can already do that in the note list context menu, or by pressing F2 in the note list.

Because it is not possible to set up a keybind for it right now.

> This PR also changes existing code, and it's unclear to me why.

All it does is factoring out the code for the rename into a method, so it can be called outside of the existing flow. And it makes sure the renaming also is possible when the note list is not visible.

pbek:
The current shortcut is F5, F2. The discussion to pull every context menu action into the main menu should not be done in a pull request.

louwers:
OK thank you for clarifying contribution etiquette. I will create an issue.

PR #3618, opened 2026-05-18
misc: fix litehtml mingw-w64 build

lazka:
The noexcept compat shim in os_types.h uses `#if _MSC_VER < 1900`. When building with clang on Windows, _MSC_VER is undefined and evaluates to 0, which is < 1900, so `noexcept` gets blanked out globally, breaking all of libc++. Guard it with `defined(_MSC_VER)` so it only fires for actual old MSVC.

---

Here is an example build log that failed: https://github.com/msys2/MINGW-packages/actions/runs/24926398391/job/72996953935?pr=29152#step:9:541


pbek:
Thank you!

PR #3672, opened 2026-08-01
Fix Joplin RAW import silently skipping some attachments

MySkeletonHurts:
## Description

While migrating a personal note collection from Joplin to QOwnNotes using
the "Import notes from Joplin" RAW-export importer, I noticed a meaningful
share of attachments weren't showing up in the imported notes. There's no
warning or log entry when this happens — the only way to notice is to
compare attachment counts before and after import. Given the volume of
missing files, I had Claude Code run an analysis. It dug into
`src/dialogs/joplinimportdialog.cpp` to figure out why.

It found four separate places where a resource reference fails to resolve
and gets silently skipped instead of imported, and wrote this PR to fix all
four. I built and ran the patched importer against my own real export
(several thousand notes and resources, migrated over multiple Joplin
versions) as well as a small synthetic repro export (attached below), and
confirmed each fix resolves the case it targets.

### 1. Empty `file_extension` metadata isn't handled

Joplin resource metadata notes (`type_: 4`) normally carry the resource's
extension in a `file_extension:` field, with the note's own title as a
fallback source for deriving one. For resources created by older Joplin
versions, I found cases where **both** are empty, even though `mime:` is
set correctly and the underlying file exists on disk with the right
extension. The importer's fallback logic then derives an empty extension,
builds a path with a trailing dot and no suffix, and never finds the file.

**Fix:** factored resource-file resolution into a shared
`findResourceFile()` helper. When both `file_extension:` and the
derived-filename fallback come up empty, it globs the `resources/`
directory for anything starting with the resource id, instead of guessing
a single exact path.

### 2. `` tags pointing at non-image-mime resources aren't found

Resources are bucketed into an "image" map or an "attachment" map purely by
mime prefix. Joplin's web clipper sometimes embeds a saved page snapshot
(`mime: text/html`) via an `` tag, e.g.:

```html
<img width="32" height="32" src=":/347198f47fd0459081a9c14d573105b6"/>
```

The `` regex matches this fine, but the lookup only checks the
"image" map — since this id was bucketed as an attachment, the lookup comes
back empty.

**Fix:** `importImage()` (and symmetrically `handleAttachments()`) now
falls back to the other map before giving up on an id.

### 3. Back-to-back attachment links can lose every other link

`handleAttachments()`'s regex used a *consuming* `[^!]` character class to
check that a link isn't actually an image reference, instead of a
zero-width lookbehind. When Joplin exports attachment links with no
separator between them (e.g. a note that's just a list of scanned
documents), the first link's match consumes the character the next link's
`[^!]` check needs, so the regex can't find a valid starting position for
it — silently dropping most links in dense chains.

**Fix:** changed the consuming `[^!]` prefix to a zero-width `(?<!!)`
negative lookbehind, so adjacent links no longer compete for the same
delimiter character.

### 4. An escaped bracket in link/alt text breaks the match entirely

The id-capturing regexes use `[^\]]*` to capture label/alt text up to the
closing `]`. Text containing an escaped bracket — not uncommon in
scraped/OCR'd web content — stops the match at the first literal `]` it
sees (the one in `\]`), which isn't followed by `(`, so the whole match
attempt fails with no fallback.

**Fix:** changed the label/alt capture in all three regexes from
`[^\]]*` to `(?:\\.|[^\]])*`, so an escaped `\]` is treated as a literal
character instead of prematurely closing the group.

## Validation

Built from source (Qt6/CMake) and ran the patched importer's GUI against my
own real Joplin export multiple times, applying the fixes incrementally to
confirm each one addressed a distinct set of previously-missing resources.
Before this patch, roughly a third of resources were silently missing after
import; with all four fixes applied, essentially all of them import
successfully.

The only resource that didn't come through was a genuinely 0-byte file that
was already corrupted in my Joplin export before I ever touched QOwnNotes —
`Note::getInsertMediaMarkdown()` correctly declines to import a 0-byte
file, which is expected behavior, not something this patch changes.

## Repro data

I can't share my actual export (personal data), so I put together a small,
fully synthetic Joplin RAW export — no real data, 5 notes + 1 notebook, 7
resources, using genuinely valid/openable PDFs/PNGs/JPEG/HTML rather than
stub files — that reproduces all four issues independently. Attached as 
[repro-export-v2.zip](https://github.com/user-attachments/files/30622360/repro-export-v2.zip). Import it via "Import notes from Joplin" as usual;
on unpatched `main` some of the resources fail to import, and with this
patch all of them do. This is really the most important part of the PR
in case you want to solve these in different ways.

### DIFF
--- a/src/dialogs/joplinimportdialog.cpp
+++ b/src/dialogs/joplinimportdialog.cpp
@@ -5,7 +5,10 @@
 #include <entities/tag.h>
 
 #include <QDebug>
+#include <QDir>
+#include <QFile>
 #include <QFileDialog>
+#include <QFileInfo>
 #include <QRegularExpression>
 #include <QRegularExpressionMatch>
 
@@ -402,8 +405,13 @@ void JoplinImportDialog::handleImages(Note& note, const QString& dirPath) {
     QString noteText = note.getNoteText();
 
     // Handle format: [![](:/imageId "hover text")]
+    //
+    // The label uses `(?:\\.|[^\]])*` instead of `[^\]]*` so that an escaped
+    // bracket (`\]`) inside the alt text -- common in web-clipped or OCR'd
+    // content -- doesn't prematurely end the capture and make the whole
+    // pattern fail to match.
     {
-        auto i = QRegularExpression(R"regex(!\[([^\]]*)\]\(:\/([\w\d]+)\s+"([^"]*)"\))regex")
+        auto i = QRegularExpression(R"regex(!\[((?:\\.|[^\]])*)\]\(:\/([\w\d]+)\s+"([^"]*)"\))regex")
                      .globalMatch(noteText);
 
         while (i.hasNext()) {
@@ -421,7 +429,7 @@ void JoplinImportDialog::handleImages(Note& note, const QString& dirPath) {
 
     // Handle format: ![alt text](:/imageId)
     {
-        auto i = QRegularExpression(R"(!\[([^\]]*)\]\(:\/([\w\d]+)\))").globalMatch(noteText);
+        auto i = QRegularExpression(R"(!\[((?:\\.|[^\]])*)\]\(:\/([\w\d]+)\))").globalMatch(noteText);
 
         while (i.hasNext()) {
             QRegularExpressionMatch match = i.next();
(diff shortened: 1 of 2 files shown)


pbek:
Thank you very much. I'm waiting for the CI results.

pbek:
## 26.8.0

- Fixed Joplin RAW import silently skipping some attachments
  (for [#3672](https://github.com/pbek/QOwnNotes/pull/3672), thank you, @MySkeletonHurts)

MySkeletonHurts:
Awesome, thanks for merging it so fast!  I just built 26.8.0 and can confirm a 100% success rate on attachment importing with my dataset (all accounted for and hashed identically).

pbek:
Thank you for the fix!

PR #3700, opened 2026-08-17
media: use the override suffix when finding an available file name

VXNCXNX:
## Summary

`Utils::Misc::findAvailableFileName()` declares an `overrideSuffix` parameter (`misc.h:151`) but never reads it, so the returned name always keeps the suffix of the input path.

Both call sites in `src/entities/note.cpp` pass a suffix they detected from the file's mime type, precisely because the input is a temporary file whose name has no suffix or the wrong one. So pasted or imported media can be saved as `qownnotes-media-AbCdEf` with no extension, or keep a `.jpeg` the mime type says should be `.jpg`.

## Repro

```
before:
findAvailableFileName("qownnotes-media-AbCdEf", dir, "png") -> "qownnotes-media-AbCdEf"
findAvailableFileName("screenshot.jpeg", dir, "jpg")        -> "screenshot.jpeg"

after:
                                                            -> "qownnotes-media-AbCdEf.png"
                                                            -> "screenshot.jpg"
```

## The fix

Use `overrideSuffix` when it is not empty, otherwise the suffix of the path. This matches `makeFileNameRandom()`, which has the same parameter and already honours it.

## Verification

`testFindAvailableFileNameOverrideSuffix` covers the two cases above, the unchanged behaviour when no override is given, and the collision counter keeping the override (`image` plus an existing `image.png` gives `image-1.png`). Reverting `newSuffix` to `fileInfo.suffix()` fails it on the first comparison.

Full unit test target passes: 45, 3, 3, 5, 9, 2, 8 and 36 across the eight test classes, `[Tests PASS]`.

## Checklist

- [x] I read the code contribution guidelines
- [x] This is a small, self-contained fix
- [x] Created against `main`, from a separate branch
- [x] Followed the repository's commit message style
- [x] No translation files edited

*Written with AI assistance (Claude Code). I built the test target against Qt 5.15.2 and ran the suite and the mutation check myself.*

### DIFF
--- a/src/utils/misc.cpp
+++ b/src/utils/misc.cpp
@@ -2739,7 +2739,9 @@ QString Utils::Misc::findAvailableFileName(const QString &filePath, const QStrin
     const QFileInfo fileInfo(filePath);
     QString baseName = fileInfo.baseName();
     baseName.truncate(200);
-    const QString newSuffix = fileInfo.suffix();
+    // The path can be a temporary file without a correct suffix, so the caller may pass the
+    // suffix it detected from the mime type
+    const QString newSuffix = overrideSuffix.isEmpty() ? fileInfo.suffix() : overrideSuffix;
     QString newBaseName = baseName;
     QString newFileName = newBaseName;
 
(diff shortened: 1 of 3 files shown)


pbek:
Is there an open issue (which I can reproduce) this PR is attempting to fix? If not, can you please open one?

VXNCXNX:
No open issue, so I opened #3701 with the repro. It boils down to `findAvailableFileName()` taking an `overrideSuffix` it never reads, while both callers in `note.cpp` pass a mime-detected one.

pbek:
Thank you!

PR #3736, opened 2026-09-12
Fix Joplin import resource dedup for MIME-mismatched extensions (follow-up to #3726)

MySkeletonHurts:
## Description

Follow-up to #3726. While validating it I found #3726's resource-dedup fix
only covers plain attachment links (`handleAttachments()`'s resource-ID
cache) — the image-tag path (`handleImages()` → `importImage()` →
`Note::getInsertMediaMarkdown()`) has an independent existence-check issue
that still duplicates a resource whenever its real file extension differs
from Qt's MIME-derived canonical suffix for it (`.jpeg` → `.jpg`, `.htm` →
`.html`, etc.) — common enough in Joplin/browser-clipper exports.

## Root cause

`Note::getInsertMediaMarkdown()`'s "does this file already exist" check
(`src/entities/note.cpp:6276`) computes a MIME-normalized suffix for the
*destination* filename, but the existence check itself still re-derives the
filename from the original source extension and never gets told about the
normalization:

```cpp
const QFileInfo fileInfo(file->fileName());     // file->fileName() = ".../resources/9584....jpeg"
QString suffix = fileInfo.suffix();              // suffix = "jpeg"
QMimeDatabase db;
const QMimeType type = db.mimeTypeForFile(file->fileName());
if (type.isValid() && !type.suffixes().isEmpty()) {
    suffix = type.suffixes().at(0);              // suffix = "jpg"  <- now MIME-normalized
}

bool useExistingFile = false;
if (Utils::Misc::fileNameExists(file->fileName(), mediaDir.path())) {   // <- passes file->fileName() again, not suffix
    ...                                            // checks for "9584....jpeg", never "9584....jpg"
}

const QString newFileName = useExistingFile
    ? ...
    : Utils::Misc::findAvailableFileName(file->fileName(), mediaDir.path(), suffix);  // <- this one does use suffix
```

Walking through two references to the same `.jpeg` resource:

1. **First reference**: existence check
   (`Utils::Misc::fileNameExists(file->fileName(), mediaDir.path())`,
   `src/utils/misc.cpp:2792`) looks for `9584....jpeg` in `media/` — not
   there, so `useExistingFile = false`.
   `Utils::Misc::findAvailableFileName()` (`src/utils/misc.cpp:2737`) is
   handed the correct `suffix` ("jpg"), sees `9584....jpg` doesn't exist
   yet, writes it as `9584....jpg`.
2. **Second reference**: existence check *again* looks for `9584....jpeg`
   — still not there (the file on disk is named `.jpg`, not `.jpeg`) — so
   `useExistingFile = false` *again*, even though the resource was just
   imported. `findAvailableFileName()` now finds `9584....jpg` already
   taken, so it mints `9584-1.jpg`.
3. Repeat for every further reference: `-2`, `-3`, ... — the existence
   check never has a chance to succeed, because it's permanently asking
   about a filename that will never exist once the MIME-normalization has
   kicked in even once.

This never manifests when source and MIME-derived extensions already
match (the common case) — `file->fileName()`'s extension and `suffix` are
the same string there, so the existence check happens to ask the right
question anyway. Confirmed resources referenced 100+ times each with
matching source/MIME extensions dedup correctly to exactly one file, zero
`-N` copies — the extension mismatch is a hard requirement for this to
trigger, not just a common trigger, even though the override
(`MessageBoxOverride/insert-media-use-existing-image`, forced to "Yes" for
the whole import in
`JoplinImportDialog::on_importButton_clicked()`,
`src/dialogs/joplinimportdialog.cpp:136-140`) is meant to guarantee reuse
in every case.

There's a second, compounding effect for resources referenced more than once
*within the same note*: `handleImages()`'s regex loop matches all occurrences
against a snapshot of the note text, but `noteText.replace(imageTag,
mediaMarkdown)` replaces every occurrence of that literal tag text at once on
the first match. So only the first occurrence's generated link actually
survives in the note; every later match in the same loop still runs
`importImage()` (writing yet another duplicate file, given the issue above)
but its resulting link is never inserted anywhere — a fully orphaned file.
This has no visible effect on the note itself either way — repeated
occurrences of a tag always render identically, pointing at the same first
file, both before and after this fix. The only observable difference is the
extra unlinked file sitting in `media/`.

## Impact

No data loss — this failure mode can only add redundant files, since it
only ever fires on the "should I skip the copy" branch, never the "should I
skip the write" branch. It's wasted disk space and, for orphaned copies,
dead weight not linked from anywhere.

## Fix

- `getInsertMediaMarkdown()`'s existence check now uses the same
  MIME-normalized `suffix` it computes for the destination filename,
  instead of `file->fileName()`'s original extension — fixes the
  cross-note/cross-reference duplication.
- `handleImages()`/`importImage()` reworked to search-and-replace by
  position instead of the old snapshot-`globalMatch()` + blanket
  `noteText.replace(tagText, ...)`, so an identical tag appearing more than
  once in one note no longer has its later occurrence's imported file
  silently orphaned.
- Three new regression tests in `tests/unit_tests/testcases/app/test_notes.cpp`:
  `testGetInsertMediaMarkdownReusesFileDespiteMimeExtensionMismatch` and
  `testHandleImagesDoesNotOrphanRepeatedIdenticalImageTagInSameNote` cover the
  two fixes directly; `testHandleImagesTerminatesOnZeroByteResource` was added
  after an automated review raised (and this then disproved, empirically) a
  suspected infinite loop on zero-byte resources.

## Testing

- All 3 new regression tests pass; full `TestNotes` suite passes with no
  regressions. Verified the tests actually catch the issue: reverted just
  the fix, confirmed both dedup tests fail against the old code, restored
  the fix, confirmed they pass.
- Synthetic repro (`repro-mime-mismatch.zip`, attached) covering a control
  case plus both the cross-note dedup issue and the same-note orphan
  compounding issue — imported through the real Joplin import dialog
  side-by-side on an unpatched build and this branch.
- Re-imported a large real-world Joplin export against this branch and
  hash-verified every resource against the source export: all notes and
  resources present and byte-identical, with the duplicate/orphan files seen
  on the unpatched build no longer appearing at all.

## Repro

Attached: `repro-mime-mismatch.zip`. A synthetic Joplin raw-export directory
with:

- a control case (`.png`, matching source/MIME extension, referenced from
  two different notes) — should dedup to one file both before and after the
  fix, confirming this isn't "any repeated resource duplicates"
- a `.jpeg` resource (Qt's canonical suffix for image/jpeg is `jpg`)
  referenced from two different notes — isolates the cross-note dedup issue
  on its own
- a `.htm` resource (canonical suffix `html`) referenced twice with the
  identical `` tag in the *same* note — exercises both the dedup issue
  and the compounding same-note orphan issue in `handleImages()` together.
  Note: both occurrences will look identical and correctly show the image in
  the app either way — the file-count difference in `media/` is what to
  check, not the rendered note.

Import it via QOwnNotes' Joplin import dialog into an empty note folder.
On an unpatched build, the repeated-reference cases each mint a `-1`
duplicate (the same-note one ending up orphaned, unlinked from either
occurrence). On this branch: one file each, correctly reused/linked.

[repro-mime-mismatch.zip](https://github.com/user-attachments/files/32145439/repro-mime-mismatch.zip)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 26.9.5
 
+- Fixed duplicate resource files being created during Joplin imports when a
+  resource's file extension differs from Qt's MIME-derived canonical suffix
+  (for [#3736](https://github.com/pbek/QOwnNotes/pull/3736))
 - Internal note links like `[QOwnNotes Android Done](QOwnNotes%20Android%20Done.md)` are now
   styled like external links, where the note file name is grayed out while the link text keeps
   the internal link color (for [#3735](https://github.com/pbek/QOwnNotes/issues/3735))
--- a/src/dialogs/joplinimportdialog.cpp
+++ b/src/dialogs/joplinimportdialog.cpp
@@ -485,59 +485,79 @@ void JoplinImportDialog::handleImages(Note& note, const QString& dirPath) {
     // bracket (`\]`) inside the alt text -- common in web-clipped or OCR'd
     // content -- doesn't prematurely end the capture and make the whole
     // pattern fail to match.
+    //
+    // Each block below re-searches noteText from a moving offset and
+    // replaces only the specific occurrence just matched (by position), then
+    // resumes searching from right after the replacement. Matching against a
+    // one-time globalMatch() snapshot while mutating noteText as we go (the
+    // previous approach) breaks whenever the exact same image tag text
+    // appears more than once in a note: QString::replace(needle, replacement)
+    // replaces every occurrence of the needle at once, so the first match's
+    // replacement also silently consumes every later match's tag text,
+    // leaving their imported files orphaned (imported, but never linked).
     {
-        auto i =
-            QRegularExpression(R"regex(!\[((?:\\.|[^\]])*)\]\(:\/([\w\d]+)\s+"([^"]*)"\))regex")
-                .globalMatch(noteText);
+        const QRegularExpression re(
+            R"regex(!\[((?:\\.|[^\]])*)\]\(:\/([\w\d]+)\s+"([^"]*)"\))regex");
+        int searchOffset = 0;
+        QRegularExpressionMatch match = re.match(noteText, searchOffset);
 
-        while (i.hasNext()) {
-            QRegularExpressionMatch match = i.next();
-            QString imageTag = match.captured(0);
+        while (match.hasMatch()) {
             QString imageName = match.captured(1);
             QString imageId = match.captured(2);
             QString hoverText = match.captured(3);
 
             // Use hover text as image name if no alt text is provided
             QString finalImageName = imageName.isEmpty() ? hoverText : imageName;
-            importImage(note, dirPath, noteText, imageTag, imageId, finalImageName);
+            searchOffset = importImage(note, dirPath, noteText, match.capturedStart(0),
+                                       match.capturedLength(0), imageId, finalImageName);
+            match = re.match(noteText, searchOffset);
         }
     }
 
     // Handle format: ![alt text](:/imageId)
     {
-        auto i =
-            QRegularExpression(R"(!\[((?:\\.|[^\]])*)\]\(:\/([\w\d]+)\))").globalMatch(noteText);
+        const QRegularExpression re(R"(!\[((?:\\.|[^\]])*)\]\(:\/([\w\d]+)\))");
+        int searchOffset = 0;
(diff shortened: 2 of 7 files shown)


MySkeletonHurts:
Bah, forgot to run clang-format before opening this.  Sorry about that.  You already addressed it while i was running it locally, thanks!

pbek:
No worries, thank you for the fix!

PR #3742, opened 2026-09-19
Fix links into the media/ folder being misread as missing notes

MySkeletonHurts:
## Description

Ctrl+clicking a link into `media/` whose target's extension matches a
configured **Note file extension** (General settings → "Note file
extensions" — `md` and `txt` by default for any existing install)
incorrectly triggers "Note was not found, create new note **\<name\>** at
path **media/**?" instead of opening the file. This is extension-based, not
about renderability — an actual image link (`.png`/`.jpg`/etc.) never hits
this path at all.

Example from a real note:

```markdown
![da34af51d2564c7cb99f1f21967da329](../../media/da34af51d2564c7cb99f1f21967da329.txt)
```

The file exists on disk at that exact path. Ctrl+clicking it still prompts
to create a new note.

## Root cause

`UrlHandler::handleNoteUrl()` only runs for links whose extension matches a
configured Note file extension — other links are opened directly, which is
why images are unaffected. Inside it, an existing special case opens
`attachments/` links externally instead of misreading them as note links,
but only checks for the literal substring `"attachments"`:

```cpp
// Open attachments with extensions that are used for notes externally
if (relativeFilePath.contains(QStringLiteral("attachments"))) {
    if (QDesktopServices::openUrl(url)) {
        return;
    }
}
```

`media/` holds the same kind of resource file
(`NoteFolder::currentMediaPath()` / `currentAttachmentsPath()`) but was
never added to the check, so a link into it falls through to the note-lookup
path and offers to create a phantom note instead.

## Fix

Extend the check to also match `"media"`.

## Testing

- `devenv shell just -- format` (the full lint/format suite) and
  `devenv shell -- just src-test` (the full test suite, all 10 test classes)
  both pass clean, 0 failures.
- Manual repro (attached: `repro-media-link.zip`) — a note folder with one
  note linking to `media/photo.png` and `media/photo-notes.txt`. Before the
  fix: the image link opens fine, but ctrl+click on the `.txt` link pops
  "Note was not found, create new note **photo-notes** at path **media/**?".
  After the fix, both open normally. Open the folder as a note folder in
  QOwnNotes, open the note, and ctrl+click each link.
- No automated regression test — `handleNoteUrl()` is tightly coupled to
  `MainWindow::instance()` and isn't currently unit-testable in isolation;
  felt out of scope for a one-line fix.

[repro-media-link.zip](https://github.com/user-attachments/files/32419387/repro-media-link.zip)


pbek:
**Review finding (medium): substring matching can misclassify note folders**

`relativeFilePath.contains(QStringLiteral("media"))` also matches legitimate note subfolders such as `multimedia`, `media-notes`, or `immediate`.

For example, with subfolders enabled, clicking a missing note link such as `[New note](multimedia/new-note.md)` reaches `handleNoteUrl()`. After lookup fails, the new condition may dispatch the nonexistent file through `QDesktopServices::openUrl()` and return, suppressing the expected create-note prompt.

Please match the actual `media` path component (or the resolved `NoteFolder::currentMediaPath()`) rather than an arbitrary substring. The existing `attachments` check has the same weakness, so this would be a good opportunity to make both checks boundary-aware.

MySkeletonHurts:
Good call.  I changed this to an exact regex match for attachments or media folders starting at position 0.

This fixes my original case of ```media/photo-notes.txt``` but keeps your existing behavior for something like ```multimedia/new-note.md``` or ```vacation/media/new-note.md```.

pbek:
Thank you!

